PR #2, opened 2018-10-22
Added template for cluster role bindings

matthew-fuller:


evrardjp:
@cinerama70 problem would be the requirement of the k8s openshift python client on the remote openstack-helm deploy node.
If we are okay with the added complexity of adding the venv creation and ansible python version wiring in this code we could go ahead for it. (keep in mind for stability purposes we would have to pin said version and keep the pin somewhere).

For me it sounded like a lot of effort with idempotency as only gain, more over that kubectl apply is supposed to handle that gracefully.

PR #4, opened 2018-11-05
Update node labels with compute, node-class, and openvswitch 

matthew-fuller:
There may be a better way to handle the single-node labels, but the idea was to pick the first non-master node to ensure the node-class label gets applied to a node where pods can be scheduled even if the master node taints are left in place (in a cluster with at least 3 worker nodes).


evrardjp:
Superseded by https://github.com/SUSE-Cloud/socok8s/pull/15

PR #8, opened 2018-11-07
Improve run sh

evrardjp:


script_library/cleanup-k8s.sh
```diff
@@ -0,0 +1,23 @@
...
+rm -rf /var/lib/openstack-helm/*
+rm -rf /var/lib/nova/*
+rm -rf /var/lib/libvirt/*
+rm -rf /etc/libvirt/qemu/*
+findmnt --raw | awk '/^\/var\/lib\/kubelet\/pods/ { print $1 }' | xargs -r -L1 -P16 sudo umount -f -l
+
+for pvc in $(kubectl get pvc -n openstack | awk ' NR > 1 { print $1 ; }'); do
+    kubectl delete pvc $pvc -n openstack;
+done
+
```
matthew-fuller: I would add an additional loop to delete pv instances once the pvcs have been deleted. They will show a state of "failed", but they may actually still be consuming space in the rbd pool if they're not manually deleted.

evrardjp: would
`kubectl delete pv -n openstack || true`

be enough?

matthew-fuller: Yes, that should do it. I'm not sure if there are any states that would cause the delete command to ignore a pv, so it might be a good idea to add the '--all' flag as well.

script_library/cleanup-k8s.sh
```diff
@@ -0,0 +1,27 @@
...
+rm -rf /var/lib/openstack-helm/*
+rm -rf /var/lib/nova/*
+rm -rf /var/lib/libvirt/*
+rm -rf /etc/libvirt/qemu/*
+findmnt --raw | awk '/^\/var\/lib\/kubelet\/pods/ { print $1 }' | xargs -r -L1 -P16 sudo umount -f -l
+
+for pvc in $(kubectl get pvc -n openstack | awk ' NR > 1 { print $1 ; }'); do
+    kubectl delete pvc $pvc -n openstack --all;
+done
+
```
cinerama70: This is great - pretty much aligns with what's suggested in the OSH documentation. It might be nice to clean out OSH's files in /tmp. Most are yml. (If doing a non-CAASP deploy you might run into the ceph UUID in a txt file but we don't have to worry about that here. :) This could easily be done in a follow-up change.

evrardjp: Great idea! Will do this in a follow up change.

PR #9, opened 2018-11-07
Freeze OSH repos

evrardjp:
Currently the OSH are tracking master, which is a fast moving
target.

This is a problem, as master has introduced regressions in the
past.

This fixes it by setting a SHA for each of the consumed repos,
so that we have a more stable interface.


evrardjp:
Yes it can be easily passed in user space.

PR #19, opened 2018-11-13
Including horizon install as per SCRD-5852

kantarun:
Added submodules initialization and clone if git installed and are
in project workspace

Modifying wait for pods logic in case job is stuck and did not report
status at all.


7_deploy_osh/roles/suse-osh-deploy/templates/suse-horizon.yaml.j2
```diff
@@ -0,0 +1,2 @@
+---
+duMMyKey: duMMyValue
```
jgu17: I am not sure I am following this.

7_deploy_osh/roles/suse-osh-deploy/files/085-horizon.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
```
cinerama70: Can we use component_install.yml rather than bringing this in when it's basically the same as what we have upstreaM?

7_deploy_osh/roles/suse-osh-deploy/files/wait-for-pods.sh
```diff
@@ -0,0 +1,43 @@
...
+while true; do
+    kubectl get pods --namespace=$1 -o json | jq -r \
+        '.items[].status.phase' | grep Pending > /dev/null && \
+        PENDING=True || PENDING=False
+    query='.items[]|select(.status.phase=="Running")'
+    query="$query|.status.containerStatuses[].ready"
+    kubectl get pods --namespace=$1 -o json | jq -r "$query" | \
+        grep false > /dev/null && READY="False" || READY="True"
+    kubectl get jobs -o json --namespace=$1 | jq -r \
+        '.items[] | .spec.completions == .status.succeeded or status == {}' | \
```
cinerama70: Is the main difference here? Is this something we could propose upstream?

I would really appreciate if we could tag all borrowed files with their origins, along with comments as to what has been modified, and link(s) to relevant upstream reviews. If we need to update these later from what is upstream it'll be very helpful information to have.

7_deploy_osh/roles/suse-osh-deploy/tasks/main.yml
```diff
@@ -174,6 +184,27 @@
     - keystone
     - run
 
+- name: Use a different horizon script as missing a multinode version without node port requirement
```
cinerama70: Aha! Could you please put that in a comment in the script itself, so we know that socok8s has added it and what the origin of the file is in case we need to update it later?

7_deploy_osh/roles/suse-osh-deploy/templates/suse-horizon.yaml.j2
cinerama70: This is part of the values override for installing the chart but I'm not sure where it gets consumed either. Comment here would be super helpful.

kantarun: There seems to be bug in ansible config_template. When src template file is empty, I was seeing error as mentioned in rocket chat discussion ( https://chat.suse.de/channel/cloud-ccp?msg=ar8HekzS2TwHR9zdj ).
I am kind of surprised this is happening as its an upstream openstack ansible module . Ideally it needs to be addressed in config_template ansible module.

To get around this issue, I am just adding a dummy entry so that it does not match and override any chart value. After adding this entry, I don't see that error.

7_deploy_osh/roles/suse-osh-deploy/files/wait-for-pods.sh
kantarun: Will add comment about change. Correctness of this change needs to be validated as I cannot find in what case status : {} comes. I saw this in my local setup and added logic.
Upstream has this https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#jobstatus-v1-batch though not still no clue when empty value comes.

7_deploy_osh/roles/suse-osh-deploy/tasks/main.yml
kantarun: Will do that

PR #22, opened 2018-11-14
New image build role; Makefiles and Dockerfiles for SUSE images

cinerama70:
Begin to define a new role related to building SUSE-based images. Add
SUSE-specific Dockerfiles and Makefiles to build them.

Building the SUSE images can be done by running 'make images-suse'.


cinerama70:
Probably, but we'll need to figure out the best way to tie them in and it'll take a while for them to land. Arun and I thought it'd be good to park them here for now. This seems like a good plan as we can do some collaborative fine-tuning.

7_deploy_osh/roles/osh-suse-image-build/files/openstack-helm/tools/images/openvswitch/Makefile
```diff
@@ -0,0 +1,53 @@
...
+# Build openvswitch Docker image for this project
+.PHONY: images
+images: build_$(IMAGE_NAME)
+
+.PHONY: images-ubuntu
+images-ubuntu: images
+
+.PHONY: images-suse
+images-suse: DOCKERFILE=Dockerfile.suse
+images-suse: images
```
kantarun: Do we need IMAGE_TAG override here other wise it seems generated docker IMAGE tag will be same as ubuntu one.

evrardjp:
You should have a look here for the upstreaming of Dockerfiles:
https://review.openstack.org/#/q/topic:osh_suse+(status:open+OR+status:merged)

With that we can leverage the upstream build process (see documentation in said patches) to build our images locally.

evrardjp:
I think the effort here should be moved to #43 and upstream now.

Itxaka:
@cinerama70 can this be closed? Seems like the way moving forward is to upstream as much as possible

evrardjp:
@Itxaka @cinerama70 isn't working for the squad nor for SUSE anymore, and we can drop this, as we have moved to different approaches since then.

PR #23, opened 2018-11-14
Modifying cleanup logic and fixing cli usage of --all

kantarun:
delete commands with component name and --all complains silently and
did not delete the requested resource. As we are providing component
name and related namespace so prioviding --all option is erroneous

### DIFF
--- a/script_library/cleanup-k8s.sh
+++ b/script_library/cleanup-k8s.sh
@@ -6,7 +6,17 @@ for line in $(helm ls -a | awk 'NR > 1 {print $1 }'); do
     helm delete $line --purge;
 done
 
-for NS in openstack ceph nfs libvirt; do
+reserved_namespaces=" kube-system kube-public default "
+deploy_namespaces=()
+for namespace_used in $(kubectl get namespaces | awk ' NR > 1 { print $1 ; }'); do
+    if [[ ! $reserved_namespaces =~ $namespace_used ]]; then
+        deploy_namespaces+=($namespace_used)
+    fi
+done
+
+echo "deploy namespaces are: ${deploy_namespaces[@]}"
+
+for NS in "${deploy_namespaces[@]}"; do
    helm ls --namespace $NS --short | xargs -r -L1 -P2 helm delete --purge
 done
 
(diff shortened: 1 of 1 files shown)


script_library/cleanup-k8s.sh
```diff
@@ -6,7 +6,17 @@ for line in $(helm ls -a | awk 'NR > 1 {print $1 }'); do
     helm delete $line --purge;
 done
 
-for NS in openstack ceph nfs libvirt; do
+reserved_namespaces=" kube-system kube-public default "
```
jgu17: thinking ahead if we deploy on shared k8s infra, I'd prefer that we explicitly limit the namespace that ccp creates/deploys to, openstack, ceph, docker-registry. reserved list is one step in the right direction, but we could still be cleaning up resources in the namespace that ccp doesn't own.

cinerama70: ..which doesn't necessarily mean the explicit namespaces are safe either. This brings up a couple of questions:
- is the current intent of this tooling developer-focused, or intended to be a part of the product?
- is it likely that it will rely on a shared CAASP infrastructure?
I think you have an entirely valid point, but I'm perceiving your comment as being bikesheddy given the current state of what we're doing. If we have good reason to assume that either of the above will be true for the tech preview though, then I'm overreacting.

jgu17: we can always get the complete list of namespaces that ccp uses, can't we? Did I miss anything when you say "doesn't necessarily mean the explicit namespaces are safe either"?

My point is that once the tool is out, we don't have any control how others are going to use the tool, deleting everything except few known namespaces doesn't seem to be safe to me,

cinerama70: I mean that it's theoretically possible for a potential user with an existing CAASP installation to have created namespaces by that name prior to using our tool.

cinerama70: Here's a suggestion: why not print what's going to happen and get the user to acknowledge it somehow prior to actually deleting stuff? See ceph's "--yes-i-really-really-mean-it".

jgu17: the ideal solution would be each chart to remove all the resource it created when the chart release is purged. with that being said, I still think it's a lot more risky we would remove more than ccp created/owns if we try to delete everything in all the names except for few known system namespaces than going thru an explicitly defined namespaces that ccp knows/owns. The proposal to ask the user confirmation is a good improvement but I don't want to impose more work to optimize at this OSH stage, I'll leave that to Arun to decide :-).

evrardjp: @cinerama70 this script runs unattended, after having a global confirmation script. We could split the script in two: One to see what will be removed, one to remove,but that sound overly complex.

I would prefer if we didn't 'whitelist' what to keep, but instead remove what we know, and maybe extend the cleanup in the future. That's more work, but that should be restricted overall.

script_library/cleanup-k8s.sh
```diff
@@ -16,17 +26,35 @@ rm -rf /var/lib/nova/*
...
-
-for pv in $(kubectl get pv -n openstack | awk ' NR > 1 { print $1 ; }'); do
-    kubectl delete pv $pv -n openstack --all || true;
+echo "deleted persistent volume claims : ${deleted_items[@]}"
+
+deleted_items=()
+echo "deleting persistent volumes in namespaces: ${deploy_namespaces[@]}"
+for NS in "${deploy_namespaces[@]}"; do
+    for pv in $(kubectl get pv -n $NS | awk ' NR > 1 { print $1 ; }'); do
+        kubectl delete pv $pv -n $NS || true;
```
evrardjp: I was not in favour of the --all in the first place, so I am fine with removing it.

kantarun:
As added another PR (https://github.com/SUSE-Cloud/socok8s/pull/40) for fixing --all usage only where resources are cleaned for given namespace only. So abandoning this one as its not needed.

PR #24, opened 2018-11-14
Add optional verbose mode for ansible playbook output

cinerama70:
If the -v flag is passed to run.sh or VERBOSITY is set to "-vvv"
in your environment, you can get verbose output from
ansible-playbook.


cinerama70:
Arun reminded me that you can set an environment variable to increase verbosity, so I'll post a version of this later which simply enables that when "-v" is provided.

evrardjp:
@cinerama70 should we close this?

PR #25, opened 2018-11-15
Added temp_ses_config.yml example file

matthew-fuller:
This ses_config file structure should be used with 7_deploy_osh until the values provided in the example file taken from SES documentation can be tested.


cinerama70:
The ses-ansible stuff needs to be updated to reflect this so our engcloud installation works and stage 2 can be used on a local SES install - have a look at ses-ansible/roles/setup_ses_configs/templates/ses-config.yml.j2

PR #26, opened 2018-11-16
Added missing conf to yaml structure

matthew-fuller:
This fix is necessary for nova and libvirt overrides to be applied during deployment of these charts


matthew-fuller:
Changes included in PR 27

PR #28, opened 2018-11-19
improvements on documentation

evrardjp:
This improves readability of the markdown content.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,29 +1,59 @@
-# fictional-octo-doodle
-fictional-octo-doodle
+# Intro
 
-This project automates the deployment of OpenStack Helm on CAASP and
+This project automates the deployment of OpenStack-Helm on CaaSP and
 SES via a series of shell scripts and Ansible playbooks.
+
+# General requirements
+
+## Ansible
+
+You need to install Ansible 2.7 or later to run these playbooks.
+
+## Cloning this repository
+
+To get started, you need to clone this repository. This repository
+uses submodules, so you need to get all the code to make sure
+the playbooks work.
+
+
+```
+git clone --recursive https://github.com/SUSE-Cloud/socok8s.git
+```
+
+Alternatively, one can fetch/update the tree of the submodules by
+running:
+
+```
+git submodule update --init --recursive
+```
 
 # Deployment on SUSE engcloud
 
+## Installing base software on your local machine
+
 On the system you plan to deploy OSH from, you'll need to install some
-software and create local configuration. Install Ansible 2.7 or
-later. Install the OpenStack and Heat CLI. Install the jq utility.
+software and create local configuration.
 
-In your clone of this repository, run "git submodule update --init
---recursive" to fetch the git submodules required for the CAASP
-deployment.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -34,43 +64,70 @@ clouds:
...
 You'll also need to pre-create some configuration in engcloud. It's
 convention here to use your username as part of the name of objects you create.
 
-Using the engcloud web interface or OpenStack CLI (i.e. "openstack
-keypair create..."), create a keypair
+Create a keypair on engcloud (using either the engcloud web interface or
+OpenStack CLI's `openstack keypair create`)
 for accessing the instances created. Remember the name of this
-keypair.
+keypair (named further `foctodoodle-key`)
```
cinerama70: I'm guessing "named further" is roughly equivalent to "which appears as 'foctodoodle-key' in the example below". Which is, of course, more verbose, but clearer.

README.md
```diff
@@ -34,43 +64,70 @@ clouds:
...
 cleanup prior to retrying scripts or playbooks should they fail. In
 some steps a delete.sh script is provided to clean up any created
 resources. Reconfirming that you've done all the previous steps to set
 up now will save you some time later.
 
+## Deploying
+
 To begin a deployment from scratch, go to the root of your socok8s
-clone and run "./run.sh". This will run each of the seven top-level
+clone and run `./run.sh` or `./run.sh full-deploy`. This will run each of the seven top-level
```
cinerama70: We might want to explain here that the default action is 'full-deploy' and what the other options are and how they are used. For the scope of what you're doing here, the first part (mentioning that 'full-deploy' is the default) is something I'd like to see addressed in this review, and the second part (explaining the other options) is something that could be done in a follow-up change. (I'm trying not to be one of those reviewers who would like you to fix every problem in the file in the current change along with giving everyone a free kitten and world peace :D)

README.md
```diff
@@ -1,29 +1,59 @@
...
 
-This project automates the deployment of OpenStack Helm on CAASP and
+This project automates the deployment of OpenStack-Helm on CaaSP and
 SES via a series of shell scripts and Ansible playbooks.
 
+# General requirements
+
+## Ansible
+
+You need to install Ansible 2.7 or later to run these playbooks.
```
kantarun: I think installing ansible 2.7 is error prone on opensuse system. Just now tried via pip install and cannot install 2.7.2 version.
May be we do this automatically (via pre-flight checks) if version is less than 2.7.x ( we need this for ansible reboot module) ? What do you think?

cinerama70: I'd be curious to see what problem you were having with it. I've successfully installed and used 2.7.1 and just pip installed 2.7.2.

README.md
evrardjp: Thanks! Will fix!

README.md
evrardjp: While I thought I was doing another PR for it, it was so small I did another commit to fix it, on top of this one. It still is the same context (improvements on the documentation)

evrardjp:
These documentation changes should be okay to review.

PR #29, opened 2018-11-19
Do not assume deployer is localhost

evrardjp:
Without this patch, deployment would fail on a non SLE (for
example a macbook) localhost.

This is a problem, as we confused two roles in this tooling, the
developer/ci node, and the OSH deployer, focused on that purpose.
They can be the same node, or be different.

This fixes it, and clarifies the usage in the documentation.

### DIFF
--- a/1_ses_node_on_openstack/create.sh
+++ b/1_ses_node_on_openstack/create.sh
@@ -27,7 +27,7 @@ openstack server add floating ip ${SERVER_NAME} $IP_CREATE
 
 pushd ${MAIN_FOLDER}
     echo ${IP_CREATE} > .ses_ip
-    grep ${IP_CREATE} inventory-ses.ini 2>&1 > /dev/null || (echo "${SERVER_NAME} ansible_ssh_host=${IP_CREATE} ansible_host=${IP_CREATE} ansible_user=root ansible_ssh_user=root" >> inventory-ses.ini)
+    grep ${IP_CREATE} inventory-ses.ini 2>&1 > /dev/null || (echo "[ses_nodes]" > inventory-ses.ini && echo "${SERVER_NAME} ansible_ssh_host=${IP_CREATE} ansible_host=${IP_CREATE} ansible_user=root ansible_ssh_user=root" >> inventory-ses.ini)
     echo "Waiting for the node to come up before scanning ssh key" && sleep 120 # 60 seconds are not enough
     ssh-keyscan -H ${IP_CREATE} >> ~/.ssh/known_hosts
 popd
--- a/1_ses_node_on_openstack/playbooks/play.yml
+++ b/1_ses_node_on_openstack/playbooks/play.yml
@@ -1,5 +1,5 @@
 ---
-- hosts: all:!localhost
+- hosts: ses_nodes
   gather_facts: no
   vars:
     ses_packages:
--- a/2_deploy_ses_aio/prepare-ses-ansible.yml
+++ b/2_deploy_ses_aio/prepare-ses-ansible.yml
@@ -6,6 +6,8 @@
     ses_ansible_folder: "{{ playbook_dir }}/../ses-ansible/"
     ses_ip: "{{ lookup('file', playbook_dir + '/../.ses_ip') }}"
   tasks:
+    # The role and play expect the hostname to be 'ses' in the
+    # 'all' group.
     - name: Prepare inventory
       copy:
         content: |
--- a/README.md
+++ b/README.md
@@ -71,6 +71,61 @@ To begin a deployment from scratch, go to the root of your socok8s
 clone and run "./run.sh". This will run each of the seven top-level
 sections of the script in order.
 
+## Reference archiecture and inventories
+
+By default, it is expected these playbooks and scripts would run
+on a CI/developer machine.
+
+In order to not pollute the developer/CI machine (called further
+'localhost'), all the data relevant for a deployment (like any
+eventual override) will be stored in user-space, unpriviledged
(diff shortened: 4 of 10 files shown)


evrardjp:
Not tested yet, but should be better.

run.sh
```diff
@@ -66,7 +66,7 @@ function delete_on_kvm(){
 function clean_k8s(){
     echo "DANGER ZONE"
     read -p "Press Enter or Ctrl-C. Enter will remove all the deployed components on your kubernetes cluster"
-    ansible -m script -a "script_library/cleanup-k8s.sh" localhost
+    ansible -m script -a "script_library/cleanup-k8s.sh" osh-deployer -i inventory-osh.ini
```
kantarun: Is it correct syntax to limit execution to osh-deployer ? I recall using --limit or -l as option to limit against specific group/host.

examples/config/inventory-osh.ini
```diff
@@ -30,6 +30,3 @@ osh-deploy-admin
 osh-deploy-masters
 osh-deploy-workers
 osh-deployer
-
-[osh-deploy-all-nodes:vars]
```
kantarun: Is this does not work ? or any specific reason/benefit of passing ansible_user in each host definition line?

1_ses_node_on_openstack/playbooks/play.yml
```diff
@@ -1,5 +1,5 @@
 ---
-- hosts: all:!localhost
+- hosts: ses_nodes
```
kantarun: Where is this host group defined ?

1_ses_node_on_openstack/create.sh
```diff
@@ -27,7 +27,7 @@ openstack server add floating ip ${SERVER_NAME} $IP_CREATE
 
 pushd ${MAIN_FOLDER}
     echo ${IP_CREATE} > .ses_ip
-    grep ${IP_CREATE} inventory-ses.ini 2>&1 > /dev/null || (echo "${SERVER_NAME} ansible_ssh_host=${IP_CREATE} ansible_host=${IP_CREATE} ansible_user=root ansible_ssh_user=root" >> inventory-ses.ini)
+    grep ${IP_CREATE} inventory-ses.ini 2>&1 > /dev/null || (echo "[ses_nodes]" > inventory-ses.ini && echo "${SERVER_NAME} ansible_ssh_host=${IP_CREATE} ansible_host=${IP_CREATE} ansible_user=root ansible_ssh_user=root" >> inventory-ses.ini)
```
kantarun: Looks like ansible_ssh_* variables are deprecated ansible 2.x forward. May be we don't need to use it.

Also is assumption that there will be always one node SES deployment in automated deployments?

https://docs.ansible.com/ansible/2.4/intro_inventory.html#hosts-and-groups

run.sh
evrardjp: it is. You could limit, but you'd still need to pass a group.

For example
```
ansible -m script -a "script_library/cleanup-k8s.sh" all -i inventory-osh.ini --limit osh-deployer
```

It makes more sense in terms of ansible workers/strategy to target the right nodes directly.

examples/config/inventory-osh.ini
evrardjp: Yes, the variables are evaluated in a group level, not at the host level. Sensu stricto, this var should be at a host level. But I am fine with keeping it. Just bad memories of overrides that had to go back and forth for certain connection plugins :)

1_ses_node_on_openstack/playbooks/play.yml
evrardjp: This group was indeed created into step 1, a little below.

1_ses_node_on_openstack/create.sh
evrardjp: You're right, we should probably remove the `ansible_ssh_*` vars.

The step 1 deploys a ses_node_on_openstack, so it creates only one node (as it uses the ses-ansible AIO automation from flavio). Therefore it creates an inventory with only one node.

I guess we could re-create another "step 1" to create ses with multi node, and (should it be required) adapt step 2.

evrardjp:
I'd rather do the ansible_ssh_* changes in a different commit. Are we okay with this?

2_deploy_ses_aio/prepare-ses-ansible.yml
```diff
@@ -6,6 +6,8 @@
     ses_ansible_folder: "{{ playbook_dir }}/../ses-ansible/"
     ses_ip: "{{ lookup('file', playbook_dir + '/../.ses_ip') }}"
   tasks:
+    # The role and play expect the hostname to be 'ses' in the
```
cinerama70: Thanks for adding this - I will also add it to my notes for using the playbooks on manually-installed SES/CAASP.

README.md
```diff
@@ -71,6 +71,61 @@ To begin a deployment from scratch, go to the root of your socok8s
 clone and run "./run.sh". This will run each of the seven top-level
 sections of the script in order.
 
+## Reference archiecture and inventories
```
cinerama70: Should be 'architecture'. If you make further changes or cleanups to this patch, it'd be great to address typo fixes, but if not we can always tidy it in a follow-up.

jgu17:
>           Not tested yet, but should be better.

has it been tested now?

evrardjp:
Will test and resolve conflicts today.

evrardjp:
All the work here can be abandonned, as part of its implementtion has been done in another PR to add CI, which had the same requirements.

PR #30, opened 2018-11-19
Simplify worker setup

evrardjp:
This involved a lot of tasks that can be summarized to a few
elements.

It should be easier to read now.


7_deploy_osh/roles/suse-osh-setup-workers/tasks/subvolumes-setup.yml
```diff
@@ -0,0 +1,11 @@
+---
+- name: Add subvolume setup script
+  script: caasp_cp_node_set_subvolumes.sh
+  register: _subvolume_setup
+  args:
+    creates: /var/lib/nova
+
+- name: Reboot machine
+  reboot:
```
kantarun: Reboot module is added in 2.7.x version which is a very recent ansible version. We may want to add logic (possibly in pre-flight validations) to warn user if ansible version is less than required version. Most OS will not have that version available by default in their installation.

7_deploy_osh/roles/suse-osh-setup-workers/tasks/main.yml
```diff
@@ -1,15 +1,7 @@
...
   include: subvolumes-setup.yml
   when: ('osh-deploy-workers' in group_names)
   tags:
     - preinstall
-
-#- name: Generate and read kubeconfig from one caasp node
-#  include: get-kubeconfig.yml
-#  run_once: true
-#  tags:
-#    - preinstall
```
kantarun: Logic is to generate kubeconfig using caasp-cli on worker node against master. Generated file requires some tweaking, especially idp related certs path and server name. Its a work-in-progress as did not have time to look further in this area. 
It will be useful to have kube config generated as manually copied config via velum UI expires after some time and need to be manually generated abd copied again.

evrardjp: I agree on this, but the code was checked out once, so I am sure it's fine if we remove comments :) 

I am fine with adding this automation, but in this case, we should refactor this: https://github.com/SUSE-Cloud/socok8s/blob/master/7_deploy_osh/roles/suse-osh-deploy/tasks/kubectl-install.yml#L38-L64
The code I linked above is relying on automation tooling to get the credentials. Sadly it relies on environment.json.

I am fine with removing this whole block of code in `kubectl-install.yml` together with the addition of local fetching of the worker node.

However, it won't work in my environment, as worker nodes don't have floating IPs, so I can't contact them directly.
But I would say that this assumption is maybe wrong, as I'd need to contact the worker anyway for different cases (subvolumes).

Should we consider this case and put it in the requirements/reference architecture?

evrardjp:
@kantarun opinion there?

evrardjp:
I will bring auto bootstrapping of a correct ansible version into run.sh in a different commit.

evrardjp:
What is the status of this?

7_deploy_osh/play.yml
```diff
@@ -29,11 +29,11 @@
         - developer_mode
         - build_osh_images | default('False') | bool
 
-- hosts: osh-deploy-cp
+- hosts: osh-deploy-workers
   gather_facts: yes
   any_errors_fatal: true
   roles:
-    - role: suse-osh-setup
+    - name: suse-osh-setup-workers
```
jsuchome: Are you sure the `name` is correct?

rhafer: Good question. Anyway, everytthing else in this playbook uses `role` so I think it's cleaner to stay with `role`

evrardjp:
Done.

7_deploy_osh/play.yml
```diff
@@ -29,11 +29,11 @@
         - developer_mode
         - build_osh_images | default('False') | bool
 
-- hosts: osh-deploy-cp
+- hosts: caasp-workers
```
evrardjp: This will be needed to match the YAML inventory

7_deploy_osh/play.yml
```diff
@@ -29,11 +29,11 @@
...
   gather_facts: yes
   any_errors_fatal: true
   roles:
-    - role: suse-osh-setup
+    - role: setup-caasp-workers
```
evrardjp: I renamed the role to explain clearly what it's supposed to do, to which kind of target nodes. We can rename again later if we want to reuse the role based on different groups (we probably shouldn't though).

PR #31, opened 2018-11-19
Allow simpler user level overrides

evrardjp:
Without this path, the default overrides provided contains
the common image overrides.

This is a problem, as it means the deployers who want to override
a single image will have to provide the entire list of overrides.

This fixes the problem by ensuring all the templates have all
the content at all time, so users can override the way they want.

To avoid not repeating ourselves, this patch uses a single
common image overrides files, but it becomes very complex to
follow.

A more surgical approach of the images overrides should be taken
in a later patchset.


evrardjp:
@kantarun I believe the templates should represent the "SUSE way". (that's why I prefixed them all with `suse-`). I therefore believe we should provide templates with our opinionated deployment things (ses, caasp, images,...). I would prefer if those overrides per template are more surgical, as it would give us far more flexibility in terms of overriding by default without userspace overrides.

Keep in mind the overrides file IS what we are doing. The difference, is that we are passing a simple file vs a global file. The reason for that is ease of understanding.
Should we have different levels of "suse default overrides" (project and global), the deployer at the end won't really know what to override, and what the precedence is. ("Do I need to template the project file or the global file here?" "Do my overrides apply if I do this..."). Simplifying to a single level (project) would still allow the deployer to do its own thing, by having a simple mechanism for overrides. 

For upstream, I was told the global overrides are used for too many things, and could therefore be phased out in some places. Yet, I believe images is a great use case for a 'global', common, override file.
It's indeed easier for consumers of OSH to have all these images in a file.

In upstream, I wanted to get rid of the multiple level override files, and just put the files into `overrides`, named it appropriately, like `leap15-loci-queens.yml`. But that's outside this review :)

PR #34, opened 2018-11-28
Draft documentation for semi-automated install

cinerama70:
Here's a draft of how to use Stages 2 & 7 on a manually installed
SES and CAASP. This is a little incomplete in the "how to install
CAASP/SES" area right now - I'm of the opinion that we can either
build on it or obsolete it via KVM with kubic-automation.

Note that this install process is not currently guaranteed to produce
a full working environment due to disagreements re the inventory-ses.ini
format between Stage 2 and Stage 7 the last time I worked with it
(which presumably also break run.sh deployment on engcloud :() But it
should give everyone an idea of what needs to be done and can be refined
as we go.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,3 +75,84 @@ sections of the script in order.
 # Deployment using KVM
 
 This is not currently supported, but is a planned future addition.
+
+# Deployment using manually-installed SES and CAASP
+
+While a fully automated baremetal or VM based installation is not
+currently supported, selected parts of the automated install process
+can be used to configure existing SES and CAASP nodes. Here is an
+overview of how to do this. Note that these instructions are a work in
+progress and do not currently provide a comprehensive overview of
+installing SES and CAASP themselves.
+
+## Manually installing nodes
+
+Install SLES12SP3 with the SES5 extension from the SLES12SP3
+ISO. Please ensure that your SES system has a second disk device
+available.
+
+Install a CAASP admin, master, and two worker nodes manually from the
+CAASP ISO, setting the hostname on each node at install time. Visit
+the admin web interface and configure it, being sure to install
+Tiller.
+
+You will also want a system that acts as a deployer, which should have
+SSH access via keys to the CAASP and SES nodes. Note that the
+playbooks currently expect to use the root user on the remote systems
+and you may have better luck running the playbooks as the root user on
+your deployer as well.
+
+## Manually running Stage 2 to configure SES
+
+On your deployer, check out this repository. Be sure to import the git
+submodules as above. In the top level of the repository, create a file
+named .ses_ip containing the IP address of your SES node. Also create
+a file named inventory-ses.ini with the following contents:
+
+ses ansible_ssh_host=<SES IP> ansible_user=root ansible_ssh_user=root
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -75,3 +75,84 @@ sections of the script in order.
...
+  examples/config/inventory-osh.ini and edit it to reflect your
+  environment.
+
+- Enable and start sshd on your deployer and the location you are
+  running the playbooks. [NOTE: this step can be removed or adapted
+  subject to merge of PR #29]
+
+- Add suse_osh_deploy_vip_with_cidr to
+  ~/suse-osh-deploy/user_variables.yml. This should be an IP available
+  on the network you're using which can be used as a VIP.
```
evrardjp: I think giving an example here is good, as ppl might believe it was only giving an ip (and IIRC, we need an IP/MASK)

README.md
```diff
@@ -75,3 +75,84 @@ sections of the script in order.
...
+
+- Download the kubeconfig from Velum on the CAASP admin node and copy
+  it to ~/suse-osh-deploy/kubeconfig and ~/.kube/config
+
+- pip install --upgrade pip to avoid segfaults when installing
```
evrardjp: Shouldn't this be done in a venv?

README.md
cinerama70: Good catch - the full value is used in suse-ingress-kube-system.yaml.j2 (TODO: we should validate that the user supplied it...)

README.md
cinerama70: I didn't add the context of where this was required to my notes, but thinking back I think it is related to getting the playbooks to run all the way through. We aren't currently making use of venvs here AFAIK.

README.md
cinerama70: I forgot to mention that this isn't currently how it looks in the examples/ dir either...can you and @kantarun clear that up?

README.md
evrardjp: my ansible and openstack client runs on a non-suse host in a venv, works fine, but yeah segfaults happen to old pip versions, so that's why I suggested the venv: Not interfering with host provided packages. But that's detail.

TODO(evrardjp): Add a venv/ansible test and bootstrap in run.sh

sgrasley: On my self deployed system... running run.sh deploy_osh... I still needed to create rdb pool in ceph for this to function.  `sudo ceph -k /etc/ceph/ceph.client.admin.keyring osd pool create rbd 128 128`

Do we need this step or did I mess up something?

evrardjp: I don't know the current state of the ses integration, sorry :(

matthew-fuller: Good catch. That pool isn't described by the ses_config file but is required for the pvcs for mariadb and rabbitmq. We'll need to add a step to get that created.

cinerama70: I might have answered this in chat but the rbd pool gets created in Stage 2 currently.

README.md
cinerama70: https://jira.suse.de/browse/SCRD-6164 created

README.md
cinerama70: @evrardjp created https://jira.suse.de/browse/SCRD-6165 for you.

PR #35, opened 2018-11-29
Updated example config file and templates to match openstack.integrate output

matthew-fuller:
This should be the final change for the ses_config processing. It makes our example config file consistent with the output of the DeepSea openstack.integrate runner, and it has a workaround for the fact that "cinder-backup" has a dash instead of an underscore in the variable name.


cinerama70:
LGTM but can you throw the link:

https://github.com/SUSE/DeepSea/blob/master/srv/modules/runners/openstack.py

Into the commit message, just to give some context?

matthew-fuller:
Commit message has been amended with the link.

PR #38, opened 2018-11-30
Fix Stage 2 output of user vars

cinerama70:
/tmp/ceph-storage-classes.yaml should now be correctly generated in
Stage 7 with this change to generation of user-variables.yml in
Stage 2.


sgrasley:
LGTM

PR #40, opened 2018-12-04
Adding heat charts deployment needed for openstack cli usage

kantarun:
Fixing cli usage of --all during resources cleanup. Lookup is done
for the namespaces added by helm charts.


cinerama70:
Can you mention something about your changes to the cleanup script in the commit message? Just want to make sure I understand the intent there.

PR #41, opened 2018-12-05
Update readme

toabctl:


cinerama70:
Thanks for taking an interest in what we're working on!

You are correct that the reiteration of the Ansible requirement is redundant, though we may want to emphasize why that is (certain features used only work in the later versions).

I like that you've outlined the openstack CLI commands required here as well, though the previous documentation isn't incorrect - it simply refers to using the web interface. So I don't think it's necessarily appropriate to replace it so much as suggest it as an alternative.

While I genuinely appreciate the help you've provided and I do want to incorporate these changes, I'm also not sure how much time and energy we want to devote to reviewing and approving this type of change in the near future - this documentation is still very much in flux. We are presently more in the state of getting something basic working than applying polishing, so I tend to have an initial negative reaction to seeing stuff like this come up.

toabctl:
> You are correct that the reiteration of the Ansible requirement is redundant, though we may want to emphasize why that is (certain features used only work in the later versions).

I don't see the `why` in the current README so I thing the redundancy just makes the README larger for no good reason.

> I like that you've outlined the openstack CLI commands required here as well, though the previous documentation isn't incorrect - it simply refers to using the web interface. So I don't think it's necessarily appropriate to replace it so much as suggest it as an alternative.

Maybe I'm missunderstanding the reason of this README. For me, this is a quickstart guide how I can start using the current work. And clicking through the web interface isn't something that I want to do (and it is wasted time).

> While I genuinely appreciate the help you've provided and I do want to incorporate these changes, I'm also not sure how much time and energy we want to devote to reviewing and approving this type of change in the near future - this documentation is still very much in flux.

Seriously? This change will save everybody who starts using this repository ~3-5 minutes bacause he/she doesn't have to click in the web interface but can just copy&paste the `openstack` commands. And the review doesn't look very complicated, or does it?

> We are presently more in the state of getting something basic working than applying polishing, so I tend to have an initial negative reaction to seeing stuff like this come up.

Wow. What a welcome for new contributors!

garloff:
Collaboration is hard -- so let's support each other!

PR #43, opened 2018-12-06
Building suse images locally via openstack-helm-images patches

kantarun:
This change is using openstack-helm-images and loci upstream patches
to build openstack services images. So added support for pulling-in
those patches. If a given patch is merged in master, then that change
is not checked out as its already available via clone of that repo with
master branch.

Currently its building stable/queens images.

### DIFF
--- a/7_deploy_osh/play.yml
+++ b/7_deploy_osh/play.yml
@@ -1,4 +1,18 @@
 ---
+
+- hosts: osh-deployer
+  gather_facts: yes
+  any_errors_fatal: true
+  roles:
+    - suse-dev-patcher
+
+- hosts: osh-deployer
+  gather_facts: yes
+  any_errors_fatal: true
+  roles:
+    - role: suse-build-images
+      when: build_osh_images | default('False') | bool
+
 - hosts: osh-deploy-cp
   gather_facts: yes
   any_errors_fatal: true
--- a/7_deploy_osh/roles/suse-build-images/tasks/main.yml
+++ b/7_deploy_osh/roles/suse-build-images/tasks/main.yml
@@ -37,9 +37,10 @@
       {{ ansible_default_ipv4.address | ipaddr('address') }} {{ local_registry_name }}
 
 - name: Make sure docker daemon file is present on host
-  file:
-    path: "/etc/docker/daemon.json"
-    state: touch
+  copy:
+    dest: "/etc/docker/daemon.json"
+    content: "{}"
+    force: no
     mode: 0644
 
 - name: Read existing docker daemon conf file
--- a/7_deploy_osh/roles/suse-build-images/tasks/main.yml
+++ b/7_deploy_osh/roles/suse-build-images/tasks/main.yml
@@ -1,0 +1,79 @@
+---
+
+- name: Install system packages for build images
+  package:
+    name: "{{ suse_build_images_packages }}"
+    state: present
+  register: install_packages
+  until: install_packages is success
+  retries: 5
+  delay: 2
+  tags:
+    - install
+
+- name: Ensure docker is running on host
+  service:
+    name: docker
+    state: started
+
+- name: Check if local docker registry server is running
+  shell: |
+    docker ps | grep {{ local_registry_name }}
+  register: registry_running_cmd_status
+  failed_when: false
+  tags:
+    - run
+
+- name: Start local docker registry if not running
+  shell: |
+    docker run -d -p {{ local_registry_port }}:{{ local_registry_port }} --name {{ local_registry_name }} --restart always registry:2
+  when:
+    - registry_running_cmd_status.rc != 0
+
+- name: Ensure local-registry DNS resolution is set in /etc/hosts
+  blockinfile:
+    path: /etc/hosts
+    block: |
+      {{ ansible_default_ipv4.address | ipaddr('address') }} {{ local_registry_name }}
+
+- name: Make sure docker daemon file is present on host
+  file:
+    path: "/etc/docker/daemon.json"
+    state: touch
+    mode: 0644
+
+- name: Read existing docker daemon conf file
+  set_fact:
+    _docker_daemon_conf: "{{ lookup('file', '/etc/docker/daemon.json') | default({}) | from_json }}"
+
+- name: Set registry entry in docker daemon conf if not defined
+  set_fact:
+    _upd_docker_daemon_conf: "{{ _docker_daemon_conf | combine({ item.key: item.value }) }}"
+  with_items:
+    - { key: "insecure-registries", value: "[\"{{local_registry_name}}:{{local_registry_port}}\"]" }
+  when:
+    - _docker_daemon_conf | json_query('"insecure-registries"') | default("", true) == ""
+
+- name: Write back updated docker daemon file
+  copy:
+    content: "{{ _upd_docker_daemon_conf | to_nice_json }}"
+    dest: "/etc/docker/daemon.json"
+  when:
+    - _upd_docker_daemon_conf is defined
+
+- name: Restart docker if needed
+  service:
+    name: docker
+    state: restarted
+  when: _upd_docker_daemon_conf is defined
+
+- name: Build Openstack Services images via loci
+  shell: |
+    export LOCI_SRC_DIR={{ devenv_root }}/loci
+    export REGISTRY_URI={{ local_registry_name }}:{{ local_registry_port }}/openstackhelm/
+    export BASE_IMAGE=leap15
+    export OPENSTACK_VERSION="{{ openstack_svcs_build_image_version }}"
+    export PUSH_TO_REGISTRY=yes
+    ./openstack/loci/build.sh
+  args:
+    chdir: "{{ devenv_root }}/openstack-helm-images"
--- a/7_deploy_osh/roles/suse-dev-patcher/defaults/main.yml
+++ b/7_deploy_osh/roles/suse-dev-patcher/defaults/main.yml
@@ -1,0 +1,46 @@
+---
+# defaults file for suse-dev-patcher
+
+devenv_root: /opt
+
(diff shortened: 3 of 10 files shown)


jgu17:
the dev patcher part looks good to me. the docker registry seems to need cert mgmt logic from Scott's notes on etherpad.

7_deploy_osh/play.yml
```diff
@@ -1,4 +1,17 @@
...
+  gather_facts: yes
+  any_errors_fatal: true
+  roles:
+    - suse-dev-patcher
+
+- hosts: osh-deployer
+  gather_facts: yes
+  any_errors_fatal: true
+  roles:
+    - { role: suse-build-images, when: build_osh_images | default('no') == 'yes' }
```
evrardjp: I'd totally prefer if that was given in non one-liner format. Also, moving towards (build_osh_images | default(False)) | bool is less error prone, should build_osh_images be provided.

7_deploy_osh/roles/suse-build-images/tasks/main.yml
```diff
@@ -0,0 +1,88 @@
...
+  shell: |
+    docker ps | grep {{ local_registry_name }}
+  register: registry_running_cmd_status
+  failed_when: false
+  tags:
+    - run
+
+- name: Start local docker registry if not running
+  shell: |
+    docker run -d -p {{ local_registry_port }}:{{ local_registry_port }} --name {{ local_registry_name }} --restart always registry:2
```
evrardjp: This doesn't ensure that the registry named {{ local_registry_name }} runs on the expected port.
I am fine with that, as its for dev purposes only, but I'd prefer seeing a comment about it before the task.

7_deploy_osh/roles/suse-build-images/tasks/main.yml
```diff
@@ -0,0 +1,88 @@
...
+    path: "/etc/docker/daemon.json"
+  register: _docker_daemon_conf_check
+
+- name: Make sure docker daemon file is present on host
+  file:
+    path: "/etc/docker/daemon.json"
+    state: touch
+    mode: 0644
+
+- name: Check if local-registry is already added in docker daemon conf file
```
evrardjp: Alternatively, you could command cat/slurp the file here, and use a  `| from_json` to build a dict object on which you can check the presence of `insecure-registries` using a simple _docker_daemon_conf['insecure-registries']. It also will be easier to review/read ansible code when you'll set the value, and writing said value to file (`to_json`). The ordering doesn't matter if you write the file anyway, because you'll only write the file if there is a change to be made.

evrardjp: Another alternative is to use quote filter to simplify reading.

7_deploy_osh/roles/suse-dev-patcher/defaults/main.yml
```diff
@@ -0,0 +1,46 @@
...
+osh_repos:
+  openstack-helm-infra:
+    src: https://git.openstack.org/openstack/openstack-helm-infra
+    version: 5da44ee3095906d97808fe73b8d0f6617cd623c7  # commit ref 4th December 2018
+  openstack-helm:
+    src: https://git.openstack.org/openstack/openstack-helm
+    version: 030868aee3e14122eac1d39a1bb2d154ef9c86c9  # frozen ref 4th December 2018
+  openstack-helm-images:
+    src: https://github.com/openstack/openstack-helm-images
+    version: master
```
evrardjp: could you freeze shas there?

7_deploy_osh/roles/suse-build-images/tasks/main.yml
```diff
@@ -0,0 +1,88 @@
...
+- name: Build Openstack Services images via loci
+  shell: |
+    export LOCI_SRC_DIR={{ devenv_root }}/loci
+    export REGISTRY_URI={{ local_registry_name }}:{{ local_registry_port }}/openstackhelm/
+    export BASE_IMAGE=leap15
+    export OPENSTACK_VERSION="{{ openstack_svcs_build_image_version }}"
+    export PUSH_TO_REGISTRY=yes
+    ./openstack/loci/build.sh
+  args:
+      chdir: "{{ devenv_root }}/openstack-helm-images"
```
evrardjp: there are extra spaces here.

7_deploy_osh/roles/suse-dev-patcher/tasks/clone-repos.yml
```diff
@@ -0,0 +1,11 @@
+---
+- name: Clone OpenStack-Helm related code repositories
+  git:
+    repo: "{{ item.value.src }}"
+    dest: "/opt/{{ item.key }}"
+    update: yes
+    force: yes
+    version: "{{ item.value.version }}"
+  with_dict: "{{ osh_repos }}"
```
evrardjp: nit: we should probably get away of with_dict, but it's so easy to read that way! :)

7_deploy_osh/roles/suse-dev-patcher/tasks/main.yml
```diff
@@ -0,0 +1,24 @@
...
+    state: present
+  register: install_packages
+  until: install_packages is success
+  retries: 5
+  delay: 2
+  tags:
+    - install
+
+- name: Clone OSH related repos
+  include: clone-repos.yml
```
evrardjp: `include` should probably replaced with a `import_tasks` here.

7_deploy_osh/roles/suse-dev-patcher/tasks/main.yml
```diff
@@ -0,0 +1,24 @@
...
+  tags:
+    - run
+
+- name: Apply upstream reviews in flight to OSH related repos
+  include: patch-review-repos.yml
```
evrardjp: `include` should probably be replaced with `include_tasks` here.

7_deploy_osh/roles/suse-dev-patcher/tasks/patch-review-repos.yml
```diff
@@ -0,0 +1,15 @@
+---
+
+- name: Apply changesets in flight for review site {{ review_site }}
+  shell: |
```
evrardjp: I'd use a `set` to fail on any error (including pipe), else it can be silently passing through until the last line.

This can be re-implemented in ansible with one task with the uri module to fetch data from the API, and another task to fetch/apply the cherry-pick. As this was also an include with a single task, you could simplify the readability of the tasks by looping direcly on the include directive. Simple and very readable.

But ofc, I'd prefer if that was a simple ansible module (or action plugin, for laziness) instead :)

7_deploy_osh/play.yml
kantarun: What format you are looking for non one-liner format.

Is it something like this? If not, please clarify.
- { role: suse-build-images, 
    when: build_osh_images | default (False) | boolean }

evrardjp: you can remove the `{` `,` and `}`.

7_deploy_osh/roles/suse-build-images/tasks/main.yml
kantarun: We are specifying local_registry_port whose default value is 5000. Currently same port is mapped on host and container side.
I am not sure what exactly you are expecting which is missing here.

7_deploy_osh/roles/suse-dev-patcher/defaults/main.yml
kantarun: I did not add specific shas here as we are merging patches in and may want to leverage latest changes there instead of keep updating sha with each patch merge. I see only our team patches in upstream so we may not worry about destabilization of master branch.

7_deploy_osh/roles/suse-dev-patcher/tasks/main.yml
kantarun: Okay..will change it here.

7_deploy_osh/roles/suse-dev-patcher/tasks/patch-review-repos.yml
kantarun: I will add set -e in beginning of shell commands.

For loop aspect, here we are looping over all patches within a review site ( e.g. review.openstack.org). That review site may have multiple patches per repo. So this is looping at repo and then at patches within that repo.

Outer loop (at include task level) is per  review site.
Given above info, this current logic works now. I am not too sure if I am understand the above suggestion. Can it be clarified little more or may be we can improve in subsequent patch if needed. As dev-patcher functonality is needed for other dependent tasks so better merge this patch if there are no significant issue.

7_deploy_osh/roles/suse-build-images/tasks/main.yml
kantarun: Will fix in next update

7_deploy_osh/roles/suse-build-images/tasks/main.yml
kantarun: Good idea. Agree treating content as json and updating is defintely better. Will make change like that in next update.

7_deploy_osh/roles/suse-dev-patcher/tasks/main.yml
evrardjp: I said `include_tasks` not `import_tasks` :)

7_deploy_osh/roles/suse-dev-patcher/tasks/patch-review-repos.yml
evrardjp: Indeed, it can be improved later.

evrardjp:
If deployer is different than localhost, the pip install trick won't work, but this is a temporary fix anyway.

PR #44, opened 2018-12-13
Fix missing repo

sgrasley:
Need a quick replacement for upstream repo being removed. SLE12_sp3:devel_tools is not available any more.


evrardjp:
Shouldn't we add this to ses-ansible role instead?

PR #45, opened 2018-12-13
Update README.md with missing package requirement.

sgrasley:
When running the full install on engcloud discovered a missing package.


jgu17:
@evrardjp Scott has added the additional pkgs. looks good to your original request?

PR #46, opened 2018-12-14
Adding suse libvirt image build step

kantarun:


kantarun:
> You probably don't need to change ses-ansible's inventory :)
> Could you explain in the commit message WHY we need the gnu parallel?

Yep, reverted inventory change. Updated commit message.

PR #48, opened 2019-01-11
Add CI

evrardjp:


script_library/requirements.txt
```diff
@@ -0,0 +1,4 @@
+ansible==2.7.5
```
dirkmueller: why do we depend on a specific set of packages?

script_library/requirements.txt
```diff
@@ -0,0 +1,4 @@
...
+netaddr==0.7.19
+jmespath==0.9.3
+requests==2.21.0
```
dirkmueller: same here - wouldn't any requests version work?

1_ses_node_on_openstack/playbooks/play.yml
```diff
@@ -18,16 +18,21 @@
...
-        zypper ar -f -G http://provo-clouddata.cloud.suse.de/repos/x86_64/SLE12-SP3-SDK-Pool/ SDK-Pool;
-        zypper ar -f -G http://provo-clouddata.cloud.suse.de/repos/x86_64/SLE12-SP3-SDK-Updates/ SDK-Updates;
-        zypper ar -f -G http://download.suse.de/ibs/SUSE:/CA/SLE_12_SP3/ SUSE-CA;
-        zypper ar -f -G http://provo-clouddata.cloud.suse.de/repos/x86_64/SUSE-OpenStack-Cloud-8-Pool/ SUSE-OpenStack-Cloud-8-Pool;
-        zypper ar -f -G https://provo-mirror.opensuse.org/repositories/utilities/SLE_12/ SLE12-utilities;
-        zypper ar -f -G https://provo-mirror.opensuse.org/repositories/Virtualization:/containers/SLE_12_SP3/ Virtualization_containers;
-        zypper ar -f -G http://provo-clouddata.cloud.suse.de/repos/x86_64/SUSE-Enterprise-Storage-5-Pool/ SUSE-Enterprise-Storage-5-Pool;
-        zypper ar -f -G http://provo-clouddata.cloud.suse.de/repos/x86_64/SUSE-Enterprise-Storage-5-Updates/ SUSE-Enterprise-Storage-5-Updates;
+        zypper ar -f -G http://kirk.provo.novell.com/repositories/SUSE/Products/SLE-SERVER/12-SP3/x86_64/product/ sle12sp3-product
```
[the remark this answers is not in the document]
evrardjp: I did that during an outage of clouddata, where only MF mirror worked. It seems fixed now, so I can probably revert.

1_ses_node_on_openstack/playbooks/play.yml
```diff
@@ -18,16 +18,21 @@
...
+        zypper ar -f -G http://kirk.provo.novell.com/repositories/SUSE/Updates/SLE-SDK/12-SP3/x86_64/update/ sle12sp3-sdk-update
+        zypper ar -f -G http://download.suse.de/ibs/SUSE:/CA/SLE_12_SP3/ SUSE-CA
+        zypper ar -f -G http://kirk.provo.novell.com/repositories/SUSE/Products/OpenStack-Cloud/8/x86_64/product/ openstackcloud-8-product
+        #TODO(evrardjp): Check if utilities/devel tools are really needed
+        #zypper ar -f -G https://provo-mirror.opensuse.org/repositories/utilities/SLE_12/ SLE12-utilities;
+        #zypper ar -f -G https://provo-mirror.opensuse.org/repositories/Virtualization:/containers/SLE_12_SP3/ Virtualization_containers;
+        #TODO(evrardjp): Check if cloud8updates are really needed
+        #zypper ar -f -G  http://kirk.provo.novell.com/repositories/SUSE/Products/OpenStack-Cloud/8/x86_64/product/ openstackcloud-8-update
+        #zypper ar -f -G http://provo-clouddata.cloud.suse.de/repos/x86_64/SUSE-OpenStack-Cloud-8-Pool/ SUSE-OpenStack-Cloud-8-Pool;
+        zypper ar -f -G http://kirk.provo.novell.com/repositories/SUSE/Updates/SLE-Module-Containers/12/x86_64/update/ slemodulecontainers-12-update
```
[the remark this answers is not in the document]
evrardjp: Totally agreed.

script_library/requirements.txt
evrardjp: We require a minimum version.

There is a series of tests that run to check minimal requirements in the pre-flight checks.
Should those requirements not be met, then software needs installation.

Currently the jenkins environment to deploy is non privileged, and doesn't allow software installation as root.
Additionally, the tooling is prepared in a way that's completely abstracted of the local environment (you could run this code on mac os if you want), so there is no assumption of any rpm/suse platform for the localhost.

Therefore, if the requirements are not meant, a python virtualenv is booted in user space with those requirements.

When we'll package software as rpms, our documentation will point to use said packages. (We'll probably go even further and package this code in a rpm which will have ansible as dependency).

Anyway, long story short, currently we need a requirements file to install software in a venv.

The version is bundled as static requirements, as I will introduce a bot to bump those on a regular basis.
I'd rather keep the requirements frozen, and bump when we now, instead of having a minimum requirement, as those might change and break at any time. I would rather have it updated when we want, with an auto PR (I might even setup a github bot for that).

script_library/requirements.txt
evrardjp: Most likely.

See also previous comment on the reason of the venv, and their freezing.

evrardjp:
Removing kirk sound a good idea, I will do it.

1_ses_node_on_openstack/playbooks/play.yml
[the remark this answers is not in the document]
evrardjp: Reverted now.

evrardjp:
Ok now that this is done, I think we should go forward.
Mirroring extra repos are done in a separate activity, for which we can then point to the mirror in a later PR.

PR #49, opened 2019-01-21
Use SES-ansible from Automation

evrardjp:
Don't carry our own fork, use automation instead.

PR #50, opened 2019-01-23
Only use dev-patcher and build-images when developer mode


evrardjp:
Only developers have a need to temporarily patch and build images
for the environment.

The deployment can now toggles theses features by reading
the OSH_DEVELOPER_MODE environment variable.
Should that variable be a truthy boolean-like (yes, true, True),
then developer mode will be enabled, and patching/image building
will be done.


evrardjp:
To be reviewed after #49

evrardjp:
This is causing the CI to fail and should be reviewed whenever possible.

PR #51, opened 2019-01-23
Simplify dev patcher


evrardjp:
The suse-dev-patcher role was filled with more complex data structures,
including repetitions of current structures. It had also unnecessary includes.

This is a problem, as its harder to maintain over time.

Instead, this moves as many tasks as possible directly in the main.yml, and
keeps the include for clarity when deploying (using a task title that can be
looped over for example). It also ensures ONE data structure for the
upstream repositories is set in place for later re-use.


evrardjp:
To be reviewed after #50

vars/manifest.yml
```diff
@@ -0,0 +1,15 @@
+---
+upstream_repos_clone_folder: ~/git/opt
```
evrardjp: that needs changing.

7_deploy_osh/roles/suse-dev-patcher/defaults/main.yml
```diff
@@ -1,46 +0,0 @@
...
-
-suse_dev_patcher_packages:
-  - autoconf
-  - git
-  - jq
-  - curl
-
-osh_repos:
-  openstack-helm-infra:
-    src: https://git.openstack.org/openstack/openstack-helm-infra
```
kantarun: why we are deleting frozen logic. I thought we will need that even after development cycle as we don't have concept of release branches in these projects. So we will need to freeze code to some point for product delivery

evrardjp: The frozen logic is not deleted. It's merely moved to a common variable file, so that everything that requires freezing is in one place (vs defaults of different roles that all need updating). The manifest of what's frozen is moved to vars/manifest.yml.

PR #54, opened 2019-01-29
Fetch code when not developer mode


evrardjp:
A customer is not expected to run dev-patcher, and for that reason
a developer_mode was introduced in the past [1].
However, the introduction of dev-patcher incorrectly removed the fetching of
the code for customers, like if dev-patcher was necessary [2].

This is a problem, as the cp/git/make would fail, because the code wouldn't
be there.

This fixes it by ensuring, if developer_mode is not enabled, than the
code will be fetched during the process.

[1]: https://github.com/SUSE-Cloud/socok8s/commit/c552066372149e6c67f00db1937a00badf6faeaa
[2]: https://github.com/SUSE-Cloud/socok8s/commit/6c5e33c6dd168c18f675149b23242375adbfe2be#diff-045e1e347bbceb0d7ccb3892819713fe


evrardjp:
https://ci.suse.de/job/cloud-ccp-full-periodic/153/console ran and went over the issue previously reached. Achievement unlocked!

PR #55, opened 2019-01-30
Partially fix step 7

evrardjp:
SES has evolved since last manual testing, and we need to adapt the
step 7 code to the latest standards.


7_deploy_osh/roles/suse-osh-deploy/tasks/apply_on_k8s.yml
```diff
@@ -0,0 +1,11 @@
+---
+- name: {{ item.action }} - file generation
```
Itxaka: This needs to be quoted

7_deploy_osh/roles/suse-osh-deploy/tasks/apply_on_k8s.yml
```diff
@@ -0,0 +1,11 @@
...
+  template:
+    src: "{{ template_src_file }}"
+    dest: "{{ templated_location }}"
+    mode: "0600"
+    user: root
+    group: root
+
+- name: {{ item.action }} - apply generated file in k8s
```
Itxaka: Same

7_deploy_osh/roles/suse-osh-deploy/tasks/apply_on_k8s.yml
evrardjp: Yes, and user doesn't exist... It was done late yesterday :P

evrardjp:
Got a few extra time, so I did a little gardening of the code (Uniform template names and cleanup some calls). I will remove those from the list if we prefer, as they are not related to this commit per se.

evrardjp:
That becomes quite ugly, I will remove those optional patches, and make them in separate PRs.

evrardjp:
Cleaned up.

evrardjp:
That deploys until cinder. I will investigate why it fails on cinder.

7_deploy_osh/roles/suse-osh-deploy/tasks/main.yml
```diff
@@ -9,9 +9,8 @@
   tags:
     - always
 
-- name: Create ses_cluster_configuration
-  set_fact:
-    ses_cluster_configuration: "{{ lookup('file', suse_osh_deploy_ses_config_path) | from_yaml }}"
+- name: Get ses_cluster_configuration
+  include_vars: "{{ suse_osh_deploy_ses_config_path }}"
```
jsuchome: This is wrong, there's no `ses_cluster_configuration` key in the default version of  `ses_config.yml` file. We might adapt this file's format, or we need to change this `include_vars` part

7_deploy_osh/roles/suse-osh-deploy/templates/cinder.yaml.j2
```diff
@@ -1,19 +1,19 @@
...
   ceph:
     enabled: true
-    admin_keyring: {{ ceph_admin_keyring_b64key | b64decode }}
+    admin_keyring: {{ ceph_admin_keyring_key }}
   cinder:
     DEFAULT:
       debug: true
       backup_driver: cinder.backup.drivers.ceph
-      backup_ceph_user: {{ ses_cluster_configuration['cinder-backup']['rbd_store_user'] }}
+      backup_ceph_user: {{ ses_cluster_configuration['cinder-backup']['rbd_store_pool_user'] }}
```
jsuchome: This change should be reflected in the example `ses_config.yml ` (not just cinder)

7_deploy_osh/roles/suse-osh-deploy/tasks/main.yml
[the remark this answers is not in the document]
jsuchome: Or, if `ses_cluster_configuration` should be part of that file (see https://github.com/SUSE-Cloud/socok8s/pull/67) that we only need to adapt the example file

7_deploy_osh/roles/suse-osh-deploy/templates/k8s-privileged-cluster-role-binding.yml.j2
```diff
@@ -41,6 +41,12 @@ subjects:
 - kind: ServiceAccount
   name: ingress-kube-system-ingress
   namespace: kube-system
+- kind: ServiceAccount
+  name: ingress-openstack-ingress
+  namespace: openstack
+- kind: ServiceAccount
+  name: ingress-ceph-ingress
+  namespace: ceph
```
rhafer: with latest openstack-helm master we need "cinder-volume" here as well.

evrardjp:
Rebased.

vars/manifest.yml
```diff
@@ -2,13 +2,13 @@
...
-    version: 030868aee3e14122eac1d39a1bb2d154ef9c86c9  # frozen ref 4th December 2018
+    version: 7edbfc8a7eadd32e6f13685b7fbc6f45b0ea660d
   openstack-helm-images:
     src: https://github.com/openstack/openstack-helm-images
-    version: master
+    version: 96048a62fa6bbe09468d316eaffc3ee9b24f74eb
   loci:
     src: https://github.com/openstack/loci
-    version: master
+    version: 96048a62fa6bbe09468d316eaffc3ee9b24f74eb
```
Itxaka: Wrong copypaste? seems to be the same sha as the one above and it makes the deploy fail :)

7_deploy_osh/roles/suse-osh-deploy/templates/libvirt.yaml.j2
```diff
@@ -1,9 +1,9 @@
 conf:
   ceph:
     enabled: true
-    admin_keyring: {{ ceph_admin_keyring_b64key | b64decode }}
+    admin_keyring: {{ ceph_admin_keyring_key }}
     cinder:
-      user: {{ ses_cluster_configuration.cinder.rbd_store_user }}
-      keyring: {{ ses_cluster_configuration.cinder.key }}
+      user: {{ ses_cluster_configuration['cinder']['rbd_store_pool_user'] }}
+      keyring: {{ suse_ses_cinder_pool_key | default(ceph_admin_keyring_key) }}"
```
jsuchome: There's an extra quote at the end here

vars/manifest.yml
evrardjp: I am a pro at wrong copy pasting. Will check.

7_deploy_osh/roles/suse-osh-deploy/templates/libvirt.yaml.j2
evrardjp: oh yeah. Will adapt with @rhafer input.

7_deploy_osh/roles/suse-osh-deploy/files/common-images-overrides.yml
```diff
@@ -27,8 +27,8 @@ images:
     heat_db_sync: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_osh_image_version }}"
     heat_engine: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_osh_image_version }}"
     heat_engine_cleaner: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_osh_image_version }}"
-    horizon: "{{ suse_osh_registry_location }}/openstackhelm/horizon:{{ suse_osh_image_version }}-clients"
-    horizon_db_sync: "{{ suse_osh_registry_location }}/openstackhelm/horizon:{{ suse_osh_image_version }}-clients"
+    horizon: "{{ suse_osh_registry_location }}/openstackhelm/horizon:queens-clients"
+    horizon_db_sync: "{{ suse_osh_registry_location }}/openstackhelm/horizon:queens-clients"
```
jsuchome: Why is `{{ suse_osh_image_version }}` dropped here? (It would be correct to drop the `-clients` parts)

evrardjp: Good catch: https://github.com/openstack/openstack-helm/blob/3cd4d0898ab517634d8c485d9e216150800f7828/horizon/values.yaml#L25

PR #58, opened 2019-01-31
travis: add some syntax checks

Itxaka:
There is no testing currently so this adds some syntax errors checks
to avoid the most basic mistakes when making changes to the repo so
we can catch those fast

As shell checks are currently failing they are adding in allowed_failure
as to not block any current work from passing the tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+language: python
+cache: pip
+dist: xenial
+
+matrix:
+  include:
+    - name: "shell syntax"
+      script: find . -path "./submodules" -prune -o -type f -name "*.sh" -print | xargs -n1 shellcheck -x --format=gcc
+    - name: "playbook syntax"
+      script: ansible-lint */*.yml
+  allow_failures:
+    - name: "shell syntax"
+    - name: "playbook syntax"
+
+install:
+  - pip install ansible-lint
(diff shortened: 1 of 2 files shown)


Itxaka:
Would need a repo owner to enable travis for the repo so this works :)

aojea:
@jsuchome can you help here? I guess you have superpowers on this repo

.travis.yml
```diff
@@ -0,0 +1,16 @@
+language: python
+cache: pip
+dist: xenial
+
+matrix:
+  include:
+    - name: "shell syntax"
+      script: find . -path "./submodules" -prune -o -type f -name "*.sh" -print | xargs -n1 shellcheck -x --format=gcc
+    - name: "playbook syntax"
+      script: ansible-lint */*.yml
```
evrardjp: I think this should be done on a case by case basis instead of massively, but I suppose that's why you added allow_failure on playbook syntax.

Sounds a good first start.

evrardjp:
This patch can be merged 2019-02-01.

.travis.yml
Itxaka: Indeed, we should strive to fix this (and shell) in small batches, thats why its disabled, so we can enable at a later date when everything is fixed so we dont introduce new broken stuff

PR #59, opened 2019-01-31
ansible: rename checks.yml to play.yml

Itxaka:
Makes it easier to syntax check playbooks based on name only and
syncs the name with the other playbooks


evrardjp:
This patch can be merged 2019-02-01.

PR #61, opened 2019-02-01
All merged changes (Do not merge and review. Now only for reference)

kantarun:


Itxaka:
Can you squash the commits with "Merge Branch" or at least change them so they reflect what was done?

kantarun:
This is no longer needed as there are other reviews added instead.

PR #62, opened 2019-02-01
Improve deletion script


evrardjp:
The deletion script let a few kubernetes components hanging.
This should be remove all the objects we didn't remove.

When all the SUSE generated files will be moved to /tmp/socok8s-* this
will also simplify the cleanup script, by deleting them in one go.

### DIFF
--- a/script_library/cleanup-k8s.sh
+++ b/script_library/cleanup-k8s.sh
@@ -20,6 +20,10 @@ for NS in openstack ceph nfs; do
     kubectl get job -n ${NS} -o name | xargs kubectl delete -n ${NS}  --ignore-not-found=true
 done
 
+# Remove extra data
+kubectl delete clusterrolebinding PrivilegedRoleBinding
+kubectl delete clusterrolebinding NonResourceUrlRoleBinding
+kubectl delete clusterrolebinding ingress-kube-system-ingress
 
 rm -rf /var/lib/openstack-helm/*
 rm -rf /var/lib/nova/*
--- a/script_library/cleanup-k8s.sh
+++ b/script_library/cleanup-k8s.sh
@@ -2,13 +2,24 @@
 
 set -x
 
-for line in $(helm ls -a | awk 'NR > 1 {print $1 }'); do
-    helm delete $line --purge;
+# TODO(evrardjp): Instead of removing ALL the charts, we should
+# list all the charts to remove and delete them in parallel
+# Delete helm charts releases
+helm ls -a | awk 'NR > 1 {print $1 }' | xargs helm delete $line --purge
+
+# Delete dangling deployments
+for NS in openstack ceph; do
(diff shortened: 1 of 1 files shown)


script_library/cleanup-k8s.sh
```diff
@@ -2,14 +2,29 @@
...
-for line in $(helm ls -a | awk 'NR > 1 {print $1 }'); do
-    helm delete $line --purge;
+# TODO(evrardjp): Instead of removing ALL the charts, we should
+# list all the charts to remove and delete them in parallel
+# Delete helm charts releases
+helm ls -a | awk 'NR > 1 {print $1 }' | xargs helm delete $line --purge
+
+# Delete dangling deployments
+for NS in openstack ceph; do
+    kubectl get deployments -n ${NS} -o jsonpath='{.items[*].metadata.name}' | xargs kubectl delete deployment -n ${NS} --ignore-not-found=true
```
kantarun: This may work as well

kubectl get deployments -n ${NS} -o name | xargs kubectl delete -n ${NS} --ignore-not-found=true

Based on this test ...

```
kubectl get deployments -n openstack -o name
...
....
deployment.extensions/nova-consoleauth
deployment.extensions/nova-novncproxy

kubectl delete deployment -n openstack deployment.extensions/nova-novncproxy
error: there is no need to specify a resource type as a separate argument when passing arguments in resource/name form (e.g. 'kubectl get resource/<resource_name>' instead of 'kubectl get resource resource/<resource_name>'
linux-bybo:/opt/socok8s # kubectl delete -n openstack deployment.extensions/nova-novncproxy
deployment.extensions "nova-novncproxy" deleted

File: script_library/cleanup-k8s.sh [hunk 2]
```diff
@@ -2,14 +2,29 @@
...
 done
 
+# If deployments didn't kill everything, let's go deeper.
 for NS in openstack ceph nfs; do
-   helm ls --namespace $NS --short | xargs -r -L1 -P2 helm delete --purge
+    kubectl get pods -n ${NS} -o name | xargs kubectl delete -n ${NS} --now --ignore-not-found=true
+    kubectl get pdb -n ${NS} -o name | xargs kubectl delete -n ${NS}  --ignore-not-found=true
+    kubectl get replicaset -n ${NS} -o name | xargs kubectl delete -n ${NS}  --ignore-not-found=true
+    kubectl get job -n ${NS} -o name | xargs kubectl delete -n ${NS}  --ignore-not-found=true
```
kantarun: Better approach with '-o name' than other patch I have...looks good

script_library/cleanup-k8s.sh
evrardjp: Interesting I got an issue when running this, that's why I didn't use -o name here. Let me retry when I am back from conference.

script_library/cleanup-k8s.sh
evrardjp: Thanks!

PR #63, opened 2019-02-01
Do not test environment is ready everytime the run.sh runs


evrardjp:
This makes the deploy_osh more bearable when developping.

### DIFF
--- a/script_library/deployment-actions-openstack.sh
+++ b/script_library/deployment-actions-openstack.sh
@@ -8,10 +8,10 @@ set -o errexit
 
 echo "Deploying on OpenStack"
 
-source ${scripts_absolute_dir}/pre-flight-checks.sh check_openstack_environment_is_ready_for_deploy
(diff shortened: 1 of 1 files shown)


jsuchome:
Could we move the check to some extra function, so that it's more DRY?

evrardjp:
I would say that this isn't a "user interface", but is a check instead every time the deployer uses an interface. The content of this is already a function, it's just about triggering it at the right time.

I don't believe we should run a "pre-flight check" function as user interface, as it would be:
- Easy to bypass
- Easy to forget

Both of those could lead to incorrectly issued bugs.

jsuchome:
I really only mean that it would make sense to have _pre_flight_checks_ function that would be called from all those places instead of copying the whole command. 
But it's not really important

evrardjp:
> I really only mean that it would make sense to have _pre_flight_checks_ function that would be called from all those places instead of copying the whole command.
> But it's not really important

If you're okay, you can do a little approval, so we can merge it without waiting for a day :)

PR #64, opened 2019-02-01
Add ARA suport


evrardjp:
This makes instrumentation easier

### DIFF
--- a/script_library/run-ansible.sh
+++ b/script_library/run-ansible.sh
@@ -35,7 +35,8 @@ function run_ansible(){
         echo "Inventory directory (${inventorydir}) exists, adding it to the ansible-playbook call."
         inventory="-i ${inventorydir}"
     fi
-    if [[ -f ${HOME}/.socok8svenv/ara.rc ]]; then
+    if [[ ${USE_ARA:-False} == "True" ]]; then
+        echo "Loading ARA"
         source ${HOME}/.socok8svenv/ara.rc
     fi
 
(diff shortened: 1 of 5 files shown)


evrardjp:
Do you think we should add a test in install_ansible venv to force installing ara when not present?

nicolasbock:
You mean as in upgrading an existing environment? That would be helpful, yes.

Itxaka:
> Do you think we should add a test in install_ansible venv to force installing ara when not present?

we should probably decide if we just want to run everything of an virtualenv or let the package installation to the end user (In my case I use pipenv for dependency management so I rather have control over it myself) with proper instructions in the readme

Also ara looks really nice! Does it introduce any delays in the playbook execution? In that case, should we make ARA optional with a env/config switch?

evrardjp:
@Itxaka it's very lightweight.

I think we can add more conditionals in the wrapper script, but it would make the wrapper slower.

For the BYO ansible, I am fine with that.

script_library/run-ansible.sh
```diff
@@ -35,6 +35,9 @@ function run_ansible(){
         echo "Inventory directory (${inventorydir}) exists, adding it to the ansible-playbook call."
         inventory="-i ${inventorydir}"
     fi
+    if [[ -f ${HOME}/.socok8svenv/ara.rc ]]; then
+        source ${HOME}/.socok8svenv/ara.rc
```
Itxaka: can we add an `ECHO` statement here so we are explicit that we are loading it?

evrardjp: Done.

PR #67, opened 2019-02-05
Bring back code for importing SES configuration

rhafer:
This is a partial revert and rework of commit 2afb55084. It brings back
the code to create the SES configuration (ses-config.yml). This file is
later consumed by get-ses-data.yml. Without it new deployments from
scratch are failing.

### DIFF
--- a/2_deploy_ses_aio/play.yml
+++ b/2_deploy_ses_aio/play.yml
@@ -23,52 +23,9 @@
   vars:
     ses_openstack_config: True
 
-- hosts: ses_nodes[0]
-  vars:
-    ses_config_file_path: "~/suse-osh-deploy/ses_config.yml"
-  tasks:
-    - name: Ensure the OSH config folder exists
-      file:
-        state: directory
-        name: "{{ ses_config_file_path | dirname }}"
-      delegate_to: localhost
-
-    # The AIO ses creates /tmp/ses_config/ on localhost,
-    # but that can disappear anytime. So we'll store
-    # The ses config in our userspace.
-    - name: Copy SES config in userspace
-      copy:
-        src: "/tmp/ses_config/ses_config.yml"
-        dest: "{{ ses_config_file_path | dirname }}"
-        remote_src: yes
-      delegate_to: localhost
-
-    # TODO(evrardjp): Add return code to latest call to make sure idempotency is known.
-    # This step is needed until we decide to use ses_config created pools (and have properly tested them)
-    - name: Create pools
-      shell: "ceph -k /etc/ceph/ceph.client.admin.keyring osd pool ls | grep {{ item }} || ceph -k /etc/ceph/ceph.client.admin.keyring osd pool create {{ item }} 128 128"
-      with_items:
-        - rbd
-        - volumes
-        - cinder_backup
-        - images
-        - vms
-- hosts: ses_nodes[0]
-  gather_facts: yes
-  vars:
-    suse_osh_deploy_user_variables: "~/suse-osh-deploy/env/extravars"
-  tasks:
-    - name: Find the admin keyring b64 value
-      shell: ceph auth get-key client.admin | base64
-      changed_when: false
-      register: _clientadminkey
-
-    - name: Create user_variables
-      delegate_to: localhost
-      blockinfile:
-        path: "{{ suse_osh_deploy_user_variables }}"
-        create: yes
-        block: |
-          ceph_admin_keyring_b64key: {{ _clientadminkey.stdout_lines[0] }}
-          ceph_user_keyring_b64key: {{ _clientadminkey.stdout_lines[0]  }}
-          suse_osh_deploy_ceph_mons: [{% for ip in ansible_all_ipv4_addresses %}'{{ ip }}:6789'{% if not loop.last %},{% endif %}{% endfor %}] #list containing all ips of the host (which should be an AIO containing the mons).
+- name: Import SES configs
+  hosts: ses_nodes[0]
+  remote_user: root
+  gather_facts: false
+  roles:
+    - ses-osh
--- a/2_deploy_ses_aio/roles/ses-osh/templates/ses-config.yml.j2
+++ b/2_deploy_ses_aio/roles/ses-osh/templates/ses-config.yml.j2
@@ -1,0 +1,28 @@
+ses_cluster_configuration:
+  ses_cluster_name: ceph
+{% if radosgw_keystone %}
+  ses_radosgw_url: "http://{{ hostvars[groups['ses_nodes'][0]].ansible_default_ipv4.address }}:8080/swift/v1"
+{% endif %}
+
+  conf_options:
+      ses_fsid: {{ hostvars[groups['ses_nodes'][0]].conf_options.stdout_lines.0 }}
+      ses_mon_initial_members: {{ hostvars[groups['ses_nodes'][0]].conf_options.stdout_lines.1 }}
+      ses_mon_host: {{ hostvars[groups['ses_nodes'][0]].conf_options.stdout_lines.2 }}
+      ses_public_network: {{ hostvars[groups['ses_nodes'][0]].conf_options.stdout_lines.3 }}
+      ses_cluster_network: {{ hostvars[groups['ses_nodes'][0]].conf_options.stdout_lines.4 }}
+  cinder:
+      rbd_store_pool: cinder
+      rbd_store_pool_user: cinder
+      keyring_file_name: ceph.client.cinder.keyring
+  cinder-backup:
+      rbd_store_pool: backups
+      rbd_store_pool_user: cinder_backup
+      keyring_file_name: ceph.client.cinder_backup.keyring
+ # Nova uses the cinder user to access the nova pool, cinder pool
+ # So all we need here is the nova pool name.
+  nova:
+      rbd_store_pool: nova
+  glance:
+      rbd_store_pool: glance
+      rbd_store_pool_user: glance
+      keyring_file_name: ceph.client.glance.keyring
--- a/2_deploy_ses_aio/roles/ses-osh/defaults/main.yml
+++ b/2_deploy_ses_aio/roles/ses-osh/defaults/main.yml
@@ -1,0 +1,5 @@
+---
+
+ses_config_path: "~/suse-osh-deploy/"
+suse_osh_deploy_user_variables: "{{ ses_config_path }}/env/extravars"
+radosgw_keystone: False
--- a/2_deploy_ses_aio/roles/ses-osh/tasks/main.yml
+++ b/2_deploy_ses_aio/roles/ses-osh/tasks/main.yml
@@ -1,0 +1,45 @@
+---
+- name: Ensure ses config directory exist
+  delegate_to: localhost
+  file:
+    path: "{{ ses_config_path }}"
+    state: directory
+
+- name: Drop ceph conf and keys
+  delegate_to: localhost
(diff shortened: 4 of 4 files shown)


2_deploy_ses_aio/roles/setup_ses_configs/templates/ses-config.yml.j2
```diff
@@ -0,0 +1,28 @@
+ses_cluster_configuration:
```
Itxaka: I would let this var be part of the playbooks instead as it seems more nicer there to point out what is loading as reference rather than hidden in this file which would make it difficutl to know where the data is coming from.

jsuchome: I guess it depends how the "official" ses_config looks like

rhafer: Not sure if there is anything really "official". The stuff coming out of the salt "openstack.integrate" function looks somewhat different (ideally we would be using that).
This PR currently is basically using the same output  as: https://github.com/SUSE-Cloud/automation/blob/master/scripts/jenkins/ardana/ansible/roles/ses_ardana_integration/tasks/main.yml

rhafer: Not sure if I am a using the right prefix here, but I guess you get the idea:

```
rhafer-ses:~ # salt-run --out=yaml openstack.integrate prefix=cloud
ceph_conf:
  cluster_network: 192.168.100.0/24
  fsid: 949e7d17-16da-37a9-b84a-ce6b52faee1e
  mon_host: 192.168.100.7
  mon_initial_members: rhafer-ses
  public_network: 192.168.100.0/24
cinder:
  key: AQBWm1pcAAAAABAAqyoU6usgH2YzmASRFZOdCg==
  rbd_store_pool: !!python/unicode cloud-cloud-volumes
  rbd_store_user: !!python/unicode cloud-cinder
cinder-backup:
  key: AQBWm1pcAAAAABAAWUYFZbTyLc4a3ElHSiuY+w==
  rbd_store_pool: !!python/unicode cloud-cloud-backups
  rbd_store_user: !!python/unicode cloud-cinder-backup
glance:
  key: AQBWm1pcAAAAABAAJOCuucpxBacq5JvfB1XKkA==
  rbd_store_pool: !!python/unicode cloud-cloud-images
  rbd_store_user: !!python/unicode cloud-glance
nova:
  rbd_store_pool: !!python/unicode cloud-cloud-vms
radosgw_urls:
- http://rhafer-ses:8080/swift/v1
```

rhafer: The ses role itself currently doesn't generate any ses_config yml. It just registers the `conf_options` and  `ceph_files` variables.

There is a another variant for import ses_config (based on "openstack.integrate") in the works for ardana currently, but that is currently blocked by a bug in deepseaa it seem. See: https://github.com/SUSE-Cloud/automation/pull/3121

evrardjp: So what we can do is indeed (just an idea of the code):
```
- name: Run openstack.integrate
  command: salt-run --out=yaml openstack.integrate prefix=cloud
  register: _saltrunopenstackintegrate
  when:  ### Ensure here it runs only on the first node of the cluster.
### Ensure the yaml is loaded in memory on that node
- name: Manipulate ses_config in userspace 
  set_fact:
    ses_config_file: "{{ _saltrunopenstackintegrate.stdout | from_yaml }}" ## Also add a top level item, and reoutput to nice yaml. Alternatively manipulate the string as json for making it easier.

- name: Drop ses_config in temp usersave
  copy:
    content: "{{ ses_config_file }}"
    dest: ### what we've decided as temp location, but in the socok8s folder, so that there is automatic deletion, and we don't have this kind of errors anymore.
  delegate_to: localhost
```

Itxaka:
Tested this and it fixed my ses-config.yml

2_deploy_ses_aio/roles/setup_ses_configs/templates/ses-config.yml.j2
rhafer: or just borrow from the ardana integration: https://github.com/SUSE-Cloud/automation/blob/5a111ad735b324752190005f285d6e1275469f4f/scripts/jenkins/ardana/ansible/roles/ses_ardana_integration/tasks/ses_configure_v2.yml

As said the main issue seems to be that "openstack.integrate" is currently broken.

2_deploy_ses_aio/roles/setup_ses_configs/defaults/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+
+ses_config_path: "/tmp/ses_config"
```
jsuchome: Can we omit this part and generate the file directly under `~/suse-osh-deploy` ?

2_deploy_ses_aio/roles/setup_ses_configs/templates/ses-config.yml.j2
evrardjp: Oh sorry, I missed that openstack.integrate is broken.

Itxaka:
I broke this with https://github.com/SUSE-Cloud/socok8s/pull/69 , sorry @rhafer :(

I would also change the role name into something else and move everything from the play.yml file into a role, but that can come later (something like `ses-osh` as the role and play.yml only includes both `ses` and `ses-osh` roles)

evrardjp:
I think we now need to figure out what's needed.... Can we completely remove SES part here?

evrardjp:
I will create a card to list how we can reduce tech debt here. I think we are masters of our own destiny, and don't need to have complex things to get it done.

2_deploy_ses_aio/roles/setup_ses_configs/defaults/main.yml
evrardjp: We should do this.

2_deploy_ses_aio/roles/setup_ses_configs/tasks/main.yml
```diff
@@ -0,0 +1,24 @@
+---
+- name: Ensure ses config directory exist
+  file:
+    path: "{{ ses_config_path }}"
+    state: directory
+  when:
```
evrardjp: we can remove the when condition.

2_deploy_ses_aio/ses-install.yml
```diff
@@ -22,3 +22,14 @@
   gather_facts: true
   roles:
     - ses
+
+- name: Get SES configs
+  hosts: localhost
+  remote_user: root
+  gather_facts: false
+  vars_files:
+    - roles/ses/defaults/main.yml
```
evrardjp: is that necessary?

2_deploy_ses_aio/roles/setup_ses_configs/tasks/main.yml
```diff
@@ -0,0 +1,24 @@
...
+    - ses_openstack_config
+
+- name: Drop ceph conf and keys
+  copy:
+    dest: "{{ ses_config_path }}/{{ item.item }}"
+    content: "{{ item.content | b64decode }}"
+    mode: 0644
+  with_items:
```
evrardjp: Where does .conf_options or .ceph_files come from?

rhafer: It's set by 2_deploy_ses_aio/roles/ses/tasks/openstack_config.yml

2_deploy_ses_aio/roles/setup_ses_configs/defaults/main.yml
rhafer: I agree. To get something working  quick I basically just checkout the old playbook and copied the parts that were missing. I'll rework it to get it to a saner state.

2_deploy_ses_aio/ses-install.yml
rhafer: Nope. Not any longer.

2_deploy_ses_aio/roles/setup_ses_configs/defaults/main.yml
rhafer: Done

rhafer:
> I would also change the role name into something else and move everything from the play.yml file into a role, but that can come later (something like `ses-osh` as the role and play.yml only includes both `ses` and `ses-osh` roles)

@Itxaka The latest push should have that.

2_deploy_ses_aio/roles/setup_ses_configs/tasks/main.yml
rhafer: done

2_deploy_ses_aio/play.yml
```diff
@@ -23,52 +23,9 @@
...
-      delegate_to: localhost
-      blockinfile:
-        path: "{{ suse_osh_deploy_user_variables }}"
-        create: yes
-        block: |
-          ceph_admin_keyring_b64key: {{ _clientadminkey.stdout_lines[0] }}
-          ceph_user_keyring_b64key: {{ _clientadminkey.stdout_lines[0]  }}
-          suse_osh_deploy_ceph_mons: [{% for ip in ansible_all_ipv4_addresses %}'{{ ip }}:6789'{% if not loop.last %},{% endif %}{% endfor %}] #list containing all ips of the host (which should be an AIO containing the mons).
+- name: Import SES configs
+  hosts: ses_nodes[0]
```
kantarun: Does this work in other's environment? 

I noticed this in my local environment.

```
PLAY [Import SES configs] ******************************************************************************************************************************************************
skipping: no hosts matched

```

2_deploy_ses_aio/roles/ses-osh/templates/ses-config.yml.j2
```diff
@@ -0,0 +1,28 @@
+ses_cluster_configuration:
+  ses_cluster_name: ceph
+{% if radosgw_keystone %}
+  ses_radosgw_url: "http://{{ hostvars[groups['ses_nodes'][0]].ansible_default_ipv4.address }}:8080/swift/v1"
+{% endif %}
+
+  conf_options:
+      ses_fsid: {{ hostvars[groups['ses_nodes'][0]].conf_options.stdout_lines.0 }}
```
kantarun: This also fails in my local environment where I have installed SES separately and now just running ses and ses-osh roles to configure it. 

Is there some other PR needed for this to work?

2_deploy_ses_aio/play.yml
rhafer: What does your inventory file look like?

2_deploy_ses_aio/roles/ses-osh/templates/ses-config.yml.j2
evrardjp: Yes I think this basically sets the expectations for other roles, and we then need to adapt everywhere else.

2_deploy_ses_aio/play.yml
evrardjp: It's probably named ses. It was a "forced" hostname before (due to ses-ansible handling).

I think I have to setup inventory expectations in the near future, this is what I have on my todolist after fixing step 7.

PR #68, opened 2019-02-05
Add clarification around python2/3 dependencies

rhafer:
Also add python-netaddr. It's required by the 'idaddr' filter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,8 +43,13 @@ Optionally (but recommended), you can preinstall the following:
 * ansible>=2.7.0
 * python-openstackclient
 * python-requests
-* python2-jmespath
-* openstacksdk
+* python-jmespath
+* python-openstacksdk
+* python-netaddr
+
+Make sure to install the variant of the packages that matches the Python
+release that ansible is using. (E.g. on Tumbleweed ansible is using Python 3,
+so install the "python3-" variant of the packages)
 
 If those aren't installed, they will be installed in a venv in ~/.socok8svenv
 


README.md
```diff
@@ -43,8 +43,13 @@ Optionally (but recommended), you can preinstall the following:
 * ansible>=2.7.0
 * python-openstackclient
 * python-requests
-* python2-jmespath
-* openstacksdk
+* python-jmespath
+* python-openstacksdk
+* python-netaddr
```
Itxaka: I thougth this was brougth by ansible already as dependency?

evrardjp: Nope, these are necessary by ansible, but ansible doesn't get them automatically.

evrardjp: Can we instead load the requirements.txt file? It would be easier, and more sustainable. :)

rhafer: Doesn't seem so. At least on Tumbleweed I needed to manually install it.

rhafer: > Can we instead load the requirements.txt file? It would be easier, and more sustainable. :)

Which requirements file? If they are really strictly required by ansible we should fix the ansible packages. I thought they were just optional dependencis for ansible.

evrardjp: In scripts library we have a requirements.txt file to bootstrap the env. It has this content in.
It's just that README.md can get out of sync.

I think it's worth using a real docs system, but in the meantime I propose we create a card for it.

evrardjp: Card created.

PR #69, opened 2019-02-06
ses aio: move bash code to ansible and merge playbooks

Itxaka:
Moves the old sh code into an asible playbook and merges the 3
different files into just one as the tasks are small and connected
so it makes no sense to have them separated


2_deploy_ses_aio/play.yml
```diff
@@ -0,0 +1,74 @@
...
+  remote_user: root
+  gather_facts: true
+  roles:
+    - ses
+  vars:
+    ses_openstack_config: True
+
+- hosts: ses_nodes[0]
+  vars:
+    ses_config_file_path: "~/suse-osh-deploy/ses_config.yml"
```
evrardjp: That could be great if this makes uses of vars/common-vars.yml. I suppose you want to do that in another PR?

On top of that, I haven't decided yet the filename we should have for this, as ansible-runner doesn't plan for that case. Either `env/extra_files/ses_config.yml`, or `artifacts/ses_config.yml`. In the meantime, we can keep it that way, but we should make sure this is tracked as somethimg we do (assign a case in jira).

2_deploy_ses_aio/play.yml
```diff
@@ -0,0 +1,74 @@
...
+  tasks:
+    - name: Ensure the OSH config folder exists
+      file:
+        state: directory
+        name: "{{ ses_config_file_path | dirname }}"
+      delegate_to: localhost
+
+    # The AIO ses creates /tmp/ses_config/ on localhost,
```
jsuchome: Should we actually follow this? As right now (with https://github.com/SUSE-Cloud/socok8s/pull/67) we're generating SES config ourselves, we could drop it directly under  `ses_config_file_path`, right?

evrardjp: I believe this could be considered as something to quickly merge.

We can iterate with the results of #67 when we are ready.

I agree with you, this needs cleanup too, though.

2_deploy_ses_aio/play.yml
Itxaka: > That could be great if this makes uses of vars/common-vars.yml. I suppose you want to do that in another PR?

Yup, this is only a small logical change so adding a different change under the same PR feels wrong to me

2_deploy_ses_aio/play.yml
Itxaka: that makes sense if that is changed and merged, will happily rebase once that gets in.

2_deploy_ses_aio/play.yml
evrardjp: OK.

2_deploy_ses_aio/play.yml
Itxaka: We may even integrate that PR with the play.yml file completely as the setup_ses_config doesnt seem like a proper role to me, or move everything into a ses-osh role and include it from the play.yml file

PR #74, opened 2019-02-08
Increase worker counts


evrardjp:
Decrease load on master, which will help us in the long run

### DIFF
--- a/3_caasp_nodes_on_openstack_heat/caasp-stack.yaml
+++ b/3_caasp_nodes_on_openstack_heat/caasp-stack.yaml
@@ -87,7 +87,7 @@ parameters:
   worker_count:
     type: number
     description: Number of Worker nodes to boot
-    default: 2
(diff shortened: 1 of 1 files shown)


evrardjp:
@jsuchome Not sure about that. It should be automatically done, what's the code you are thinking about?

jsuchome:
Hm, I was thinking `environment.json` but you are right, that should be generated

evrardjp:
This changes the defaults, but ansible is passing that as a variable anyway, so I am not 100% sure it's really needed.

Itxaka:
needs rebase!

evrardjp:
We can probably abandon this, now this parameter is given to the stack.
https://github.com/SUSE-Cloud/socok8s/blob/2875bdce944f5e1b730affd6baeedddb87475fdd/playbooks/openstack-deploy_caasp.yml#L47

PR #75, opened 2019-02-08
Inject the correct pool names into the "ses" role

rhafer:
This should make it create the pools that we actually need for osh.

### DIFF
--- a/2_deploy_ses_aio/play.yml
+++ b/2_deploy_ses_aio/play.yml
@@ -22,6 +22,27 @@
     - ses
   vars:
     ses_openstack_config: True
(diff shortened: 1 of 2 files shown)


jsuchome:
Who is actually executing `Create openstack pool(s)` task (that consumes `ses_openstack_pools`)?

rhafer:
> Who is actually executing `Create openstack pool(s)` task (that consumes `ses_openstack_pools`)?

That's happening here:
https://github.com/SUSE-Cloud/automation/blob/master/scripts/jenkins/ses/ansible/roles/ses/tasks/openstack_config.yml

PR #77, opened 2019-02-12
Added option to delete CaaSP only

jsuchome:
If CaaSP deployment goes wrong, user might need to redeploy it.
Instead of tearing down complete socok8s setup, it should be enough
to delete CaaSP before.

(This will need rebase after the merge of https://github.com/SUSE-Cloud/socok8s/pull/76)

### DIFF
--- a/script_library/deployment-actions-openstack.sh
+++ b/script_library/deployment-actions-openstack.sh
@@ -55,6 +55,7 @@ function deploy_osh(){
 }
 function teardown(){
     clean_openstack
+    delete_caasp
     clean_userfiles
 }
 function clean_k8s(){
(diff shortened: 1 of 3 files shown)


jsuchome:
Hm, I guess we'd need to delete the inventory items as well

script_library/deployment-actions-openstack.sh
```diff
@@ -64,12 +65,14 @@ function clean_k8s(){
...
-function clean_openstack(){
-    echo "Deleting on OpenStack"
-    ${socok8s_absolute_dir}/4_osh_node_on_openstack/delete.sh
+function delete_caasp(){
     echo "Delete Caasp nodes"
     ${socok8s_absolute_dir}/3_caasp_nodes_on_openstack_heat/delete.sh || true
     ${socok8s_absolute_dir}/3_caasp_nodes_on_openstack_manually/delete.sh || true
+}
+function clean_openstack(){
+    echo "Deleting on OpenStack"
```
evrardjp: couldn't that clean_openstack call delete_caasp as first step instead then?

jsuchome: Well, I wanted to split those, so `delete_caasp` is called from `teardown` (although maybe in the wrong order?)

jsuchome:
> In that case, a developer could simply delete the caasp environment on openstack by sourcing the 
> actions, and running delete_caasp.

So you're basically saying I should not put it into `run.sh`

> In that case, the delete_caasp should remove a few lines from the inventory too.

Yep, I realized that as well

evrardjp:
If we split the inventory dir into multiple files, that should be easy to do.

evrardjp:
Inventory has been split into multiple files, so it should be far easier to achieve that.

PR #81, opened 2019-02-12
Fix bash warts

Itxaka:
This should make it pass our shellcheck test on travis.

Also simplifies the helm delete command

Also removes a extra inventory var that was created but never used

### DIFF
--- a/playbooks/roles/deploy-osh/files/020-ingress.sh
+++ b/playbooks/roles/deploy-osh/files/020-ingress.sh
@@ -1,4 +1,7 @@
 #!/bin/bash
+# shellcheck disable=SC2086
+# disable checks for double quoting
+# this file comes from upstream + some changes, keep it as close as possible
 
 # Copyright 2017 The Openstack-Helm Authors.
 #
@@ -17,7 +20,7 @@
 set -xe
 
 #NOTE: Deploy global ingress
-: ${OSH_INFRA_PATH:="../openstack-helm-infra"}
+: "${OSH_INFRA_PATH:=../openstack-helm-infra}"
 tee /tmp/ingress-kube-system.yaml << EOF
 pod:
   replicas:
--- a/playbooks/roles/deploy-osh/files/085-horizon.sh
+++ b/playbooks/roles/deploy-osh/files/085-horizon.sh
@@ -1,4 +1,7 @@
 #!/bin/bash
+# shellcheck disable=SC2086
+# disable checks for double quoting
+# this file comes from upstream + some changes, keep it as close as possible
 
 # Copyright 2017 The Openstack-Helm Authors.
 #
@@ -23,7 +26,7 @@ set -xe
 make horizon
 
 #NOTE: Deploy command
-: ${OSH_EXTRA_HELM_ARGS:=""}
+: "${OSH_EXTRA_HELM_ARGS:=""}"
 helm upgrade --install horizon ./horizon \
     --namespace=openstack \
     --set manifests.network_policy=true \
--- a/playbooks/roles/setup-caasp-workers/files/caasp_cp_node_set_subvolumes.sh
+++ b/playbooks/roles/setup-caasp-workers/files/caasp_cp_node_set_subvolumes.sh
@@ -9,15 +9,15 @@ systemctl stop docker
 
 mounted_snapshot=$(mount | grep snapshot | gawk  'match($6, /ro.*@\/.snapshots\/(.*)\/snapshot/ , arr1 ) { print arr1[1] }')
 
-if [ -n "$mounted_snapshot" ]; then
-  btrfs property set -ts /.snapshots/$mounted_snapshot/snapshot ro false
+if [[ -n "${mounted_snapshot}" ]]; then
+  btrfs property set -ts /.snapshots/"${mounted_snapshot}"/snapshot ro false
 fi
 
 root_remounted_rw=false
 
 create_subvolume(){
-  if [ ! -d "$1" ]; then
-    if [ $root_remounted_rw = false ]; then
+  if [[ ! -d "$1" ]]; then
+    if [[ "${root_remounted_rw}" = false ]]; then
       mount -o remount, rw /
       root_remounted_rw=true
     fi
@@ -29,6 +29,6 @@ create_subvolume /var/lib/nova
 create_subvolume /var/lib/neutron
 create_subvolume /var/lib/libvirt
 
-if [ -n "$mounted_snapshot" ]; then
-  btrfs property set -ts /.snapshots/$mounted_snapshot/snapshot ro true
+if [[ -n "${mounted_snapshot}" ]]; then
+  btrfs property set -ts /.snapshots/"${mounted_snapshot}"/snapshot ro true
 fi
--- a/script_library/cleanup-k8s.sh
+++ b/script_library/cleanup-k8s.sh
@@ -5,7 +5,7 @@ set -x
 # TODO(evrardjp): Instead of removing ALL the charts, we should
 # list all the charts to remove and delete them in parallel
 # Delete helm charts releases
-helm ls -a | awk 'NR > 1 {print $1 }' | xargs helm delete $line --purge
+helm ls -a -q | xargs -n1 helm delete --purge
 
 # Delete dangling deployments
 for NS in openstack ceph; do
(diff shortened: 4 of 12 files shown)


Itxaka:
@evrardjp done!

evrardjp:
It looks good with my eyes, but I haven't tested the code.
Did you have a go at it, to see if it breaks the deployment? :)

Itxaka:
> It looks good with my eyes, but I haven't tested the code.
> Did you have a go at it, to see if it breaks the deployment? :)

currently running it to see if I broke everything!

Itxaka:
Updated with a small fix for the extra vars to be passed correctly to ansible-playbook even when they are empty

Still testing the rest

Itxaka:
Tested and it seems correct.

script_library/run-ansible.sh
```diff
@@ -29,21 +33,18 @@ function run_ansible(){
     fi
 
     if [[ -f ${extravarsfile} ]]; then
-        echo "Extra variables file exists: $(realpath ${extravarsfile}). Loading its vars in ansible-playbook call."
-        extra_vars="-e @${extravarsfile}"
+        echo "Extra variables file exists: $(realpath "${extravarsfile}"). Loading its vars in ansible-playbook call."
+        ansible_command="$ansible_command -e ${extravarsfile}"
```
evrardjp: If there is no extravars file, the ansible_command will not be set, is that right?

Itxaka: nope, if there is no extravars then the extravars switch wont be added to the command, the command `ansible_command` is already defined at the top of the file

playbooks/roles/deploy-osh/files/020-ingress.sh
```diff
@@ -17,7 +17,7 @@
 set -xe
 
 #NOTE: Deploy global ingress
-: ${OSH_INFRA_PATH:="../openstack-helm-infra"}
+: "${OSH_INFRA_PATH:="../openstack-helm-infra"}"
```
jhesketh: I'm not a bash expert, but this has double quotes already inside it; will it cause any issues? (it works in my testing so I suspect it's fine)

playbooks/roles/setup-caasp-workers/files/caasp_cp_node_set_subvolumes.sh
```diff
@@ -10,7 +10,7 @@ systemctl stop docker
 mounted_snapshot=$(mount | grep snapshot | gawk  'match($6, /ro.*@\/.snapshots\/(.*)\/snapshot/ , arr1 ) { print arr1[1] }')
 
 if [ -n "$mounted_snapshot" ]; then
-  btrfs property set -ts /.snapshots/$mounted_snapshot/snapshot ro false
+  btrfs property set -ts /.snapshots/"$mounted_snapshot"/snapshot ro false
```
jhesketh: (nit) Should this be `"${mounted_snapshot}"`?

playbooks/roles/deploy-osh/files/020-ingress.sh
evrardjp: It's probably working because we're not entering that part of the default assignment, but I would be glad to see an answer.

script_library/run-ansible.sh
evrardjp: Sorry I didn't see that :p Thanks!

script_library/run-ansible.sh
```diff
@@ -1,9 +1,13 @@
 #!/bin/bash
 
-
+# exit if vars is not set
+socok8s_absolute_dir=${socok8s_absolute_dir:?}
```
evrardjp: I was kinda abusing this by sourcing that script and be able to run the plays directly, so it means I can't do it anymore?

Shouldn't that be warned with a message? It seems failing without explaining why is unfriendly...

playbooks/roles/deploy-osh/files/020-ingress.sh
Itxaka: Im my testing it did work, which is something that surprised me, let me retest it just in case....

playbooks/roles/setup-caasp-workers/files/caasp_cp_node_set_subvolumes.sh
Itxaka: :+1:

script_library/run-ansible.sh
Itxaka: I guess you can as long as you set the `socok8s_absolute_dir` env manually?

Agreed on no message, let me add one to all the instances of this so its more user friendly

playbooks/roles/deploy-osh/files/020-ingress.sh
Itxaka: it actually works (would need some bash expert to explain why to me) but it also works without the double quotes inside it, so Im gonna remove those

Itxaka:
Should be better now!

evrardjp:
I am still testing this one out :)

script_library/run-ansible.sh
```diff
@@ -1,9 +1,13 @@
 #!/bin/bash
 
-
+# exit if vars is not set
```
evrardjp: I would still prefer to not exit the shell. I suppose this doesn't exit the shell but stop the process by returning an error, is that correct? In that case I am okay with it.

script_library/run-ansible.sh
evrardjp: the @ has disappeared. It needs to stay.

playbooks/roles/deploy-osh/files/020-ingress.sh
```diff
@@ -54,8 +54,8 @@ EOF
   helm upgrade --install ingress-${NAMESPACE} ${OSH_INFRA_PATH}/ingress \
     --namespace=${NAMESPACE} \
     --values=/tmp/ingress-${NAMESPACE}.yaml \
-  ${OSH_EXTRA_HELM_ARGS} \
-  ${!OSH_EXTRA_HELM_ARGS_INGRESS_NAMESPACE}
+  "${OSH_EXTRA_HELM_ARGS}" \
```
evrardjp: This seems to generate invalid input for helm CLI.

evrardjp: Same applies for horizon change.

script_library/run-ansible.sh
Itxaka: yup, it will stop the process and print the message

script_library/run-ansible.sh
Itxaka: ah yes, this failed for me when testing, Im not sure why so I had to remove it

Good catch!

Fake edit: YEs, I remember, it failed because I had the path but there was nothing in it or something like that so it was only adding the `@` at the command line which made it fail.

Itxaka: done

playbooks/roles/deploy-osh/files/020-ingress.sh
Itxaka: done, actually I disabled the checking for hte file as it comes from upstream, we should not modify for aesthetic reasons.

evrardjp:
This kinda behaves differently than what we have right now, so I will give it another go in my next rebuild.

script_library/cleanup-k8s.sh
```diff
@@ -37,29 +37,29 @@ findmnt --raw | awk '/^\/var\/lib\/kubelet\/pods/ { print $1 }' | xargs -r -L1 -
...
-        kubectl delete configmap $configmap -n $NS ;
+        kubectl delete configmap "$configmap" -n $NS ;
     done
 done
 
 
 echo "Removing suse socok8s files in /tmp"
 
-pushd /tmp
+pushd /tmp || exit
```
dirkmueller: at least exit with a nonzero exit code?

script_library/cleanup-k8s.sh
```diff
@@ -37,29 +37,29 @@ findmnt --raw | awk '/^\/var\/lib\/kubelet\/pods/ { print $1 }' | xargs -r -L1 -
...
   rm -f socok8s-*
   # TODO(evrardjp): When uniform filenames are used (starting with socok8s-) we can only keep the OSH templated files here.
   for  filename in suse-mariadb.yaml suse-rabbitmq.yaml suse-keystone.yaml suse-memcached.yaml suse-glance.yaml suse-horizon.yaml suse-cinder.yaml suse-ovs.yaml suse-libvirt.yaml suse-nova.yaml suse-ingress-kube-system.yaml suse-ingress-namespace.yaml suse-heat.yaml; do
       rm -f /tmp/$filename;
   done
-popd
+popd || exit
```
dirkmueller: same here

playbooks/roles/deploy-osh/files/085-horizon.sh
```diff
@@ -23,7 +26,7 @@ set -xe
 make horizon
 
 #NOTE: Deploy command
-: ${OSH_EXTRA_HELM_ARGS:=""}
+: "${OSH_EXTRA_HELM_ARGS:=""}"
```
evrardjp: shouldn't the nesting (`"<>=""}"`) be removed?

Should we really lint that upstream script?

evrardjp: I guess it's good to lint it if we carry our own version. But I believe we should fix upstream instead, and not carry this file.

script_library/cleanup-k8s.sh
evrardjp: agreed.

script_library/cleanup-k8s.sh
evrardjp: ditto.

script_library/run-ansible.sh
```diff
@@ -29,21 +33,18 @@ function run_ansible(){
...
-        source ${HOME}/.socok8svenv/ara.rc
+        # shellcheck disable=1090
+        source "${HOME}"/.socok8svenv/ara.rc
     fi
 
-    pushd ${socok8s_absolute_dir}
-        ansible-playbook ${extra_vars:-} -i ${inventorydir} $@ -v
-    popd
+    pushd "${socok8s_absolute_dir}" || exit
+        ${ansible_command} -i "${inventorydir}" "$@" -v
```
evrardjp: are we sure this is passing arguments properly, if multiple arguments in $@ are provided? I would have the impression they would be received as one, but don't trust my bash foo.

toabctl:
@Itxaka needs a rebase

jsuchome:
@Itxaka also, please rather fix upstream scripts of those we are using (and let's focus on getting the downstream ones away)

Itxaka:
I believe this should be reworked to just check our downstream files as there is no bash checks upstream *and* we will get rid of those hopefully soon and just use the upstream ones directly

evrardjp:
> I believe this should be reworked to just check our downstream files as there is no bash checks upstream _and_ we will get rid of those hopefully soon and just use the upstream ones directly

Agreed

Itxaka:
will reopen a new one when time permits it :D

PR #83, opened 2019-02-12
Utils repo is needed for gnu_parallel package

jsuchome:


playbooks/generic-build_images.yml
```diff
@@ -11,6 +11,12 @@
         - always
         - imagebuilder
         - upstream_patching
+  tasks:
+    - name: Add utilities repository
```
evrardjp: I would prefer if that would be included in the suse-build-image role.

That role is self contained about the setup that needs to happen on a host for building images.

On top of that, that repo should probably in a variable, so that it's easy to override.
That variable can be in the defaults/main.yml of the suse-build-images role.

Itxaka: agreed

jsuchome:
@Itxaka @evrardjp updated ... but it still does not work for me, `zypper_repository` fails with "Specified local path does not exist or is not accessible."
(although it works manually)

evrardjp:
@jsuchome Could you extend the debugging? I mean if it works manually but not in ansible, there must be a problem with the module -- We can only fix the module by knowing the reason it fails (For which we could technically create a PR, and point to that PR in the comment of this task "TODO():Move to zypper_repository when bug xxxx is fixed.)

evrardjp:
To debug, just export ANSIBLE_DEBUG=yes and ANSIBLE_VERBOSITY=3

playbooks/roles/suse-build-images/defaults/main.yml
```diff
@@ -21,3 +21,6 @@ build_infra_components_list:
 suse_build_images_packages:
   - docker
   - gnu_parallel
+
+sle_utils_repository_url:
+  - "https://download.opensuse.org/repositories/utilities/SLE_12"
```
evrardjp: Please put this on the same line (and remove the  - ) as the "sle_utils_repository_url" to be a string.

jsuchome:
Updated, now it works

Itxaka:
ah, you are using autom import keys, makes sense then :+1:

PR #84, opened 2019-02-12
Provide example WORKDIR


evrardjp:
Currently there is no complete example of what is needed to bring your
own WORKDIR.

This fixes the problem by completing the "examples" folder, so that it
contains all the necessary structure if you want to create your own
workdir, instead of letting it get generated.


jsuchome:
Could you put it into `examples/suse-osh-deploy/` ? Or write somewhere what WORKDIR is supposed to mean?

evrardjp:
> Could you put it into `examples/suse-osh-deploy/` ? Or write somewhere what WORKDIR is supposed to mean?

That's fair. It seems WORKDIR is not documented. Could we do this in a different PR though?

evrardjp:
My hope would be to later move $WORKDIR to a flexible location (that would depend on CI build ID, so that we can run multiple builds in parallel).

PR #86, opened 2019-02-12
Adapt to new ses_config structure


evrardjp:
The ses_config was changed in this tooling to match the latest
ardana changes. Therefore the set_fact would fail to get the proper mon
hosts.

This should fix the set_fact (when the user is not overriding its
value with extravars), for the new structure.


evrardjp:
If you don't provide your own suse_osh_deploy_ceph_mons list, this should fix it.

PR #90, opened 2019-02-13
Rework ssh wait loop and knownhosts update

rhafer:
The addition of the host keys to the knownhosts file frequently failed
for the SES and CaaSP nodes because ssh-keyscan didn't return any valid
keys for the nodes yet.

This commit changes the wait condition to wait until sshd is answering
on port 22 and additionally adds a retry to the "ssh-keyscan" call for
running the command until it actually returns something. (Unfortunately
"ssh-keyscan" always returns a 0 exit code it seems)

### DIFF
--- a/playbooks/openstack-deploy_caasp.yml
+++ b/playbooks/openstack-deploy_caasp.yml
@@ -183,18 +183,31 @@
 
         - meta: refresh_inventory
 
-        # Waiting for port is not enough, as the port might be up (systemd)
-        # while ssh is not ready. Wait for connection is not good either
-        # because the key is unknown.
-        - name: Wait for a while for SSH to be properly booted up
+        - name: Wait for a SSH to be properly started
           wait_for:
-            timeout: 120
+            host: "{{ hostvars[item]['ansible_host'] }}"
+            port: 22
+            search_regex: OpenSSH
+            state: started
+            delay: 10
+            timeout: 180
+          loop: "{{ groups['caasp-admin'] + groups['caasp-masters'] + groups['caasp-workers'] }}"
 
         - name: Get pubkey, and add it to known hosts
-          shell: ssh-keyscan -H {{ hostvars[item]['ansible_host'] | quote }} >> ${HOME}/.ssh/known_hosts
+          command: "ssh-keyscan -H {{ hostvars[item]['ansible_host'] | quote }}"
+          register: _caaspsshkeys
+          retries: 10
+          delay: 10
+          until: _caaspsshkeys.stdout and _caaspsshkeys.stdout.strip()
           loop: "{{ groups['caasp-admin'] + groups['caasp-masters'] + groups['caasp-workers'] }}"
-          args:
-            executable: /bin/bash
+
+
+        - name: Insert known host
+          known_hosts:
+            state: present
+            name: "{{ hostvars[item.item]['ansible_host'] }}"
+            key: "{{ item.stdout }}"
+          loop: "{{ _caaspsshkeys.results }}"
 
     - name: Handle deletion
       when:
--- a/playbooks/openstack-osh_instance.yml
+++ b/playbooks/openstack-osh_instance.yml
@@ -59,16 +59,22 @@
 
         - meta: refresh_inventory
 
-        # Waiting for port is not enough, as the port might be up (systemd)
-        # while ssh is not ready. Wait for connection is not good either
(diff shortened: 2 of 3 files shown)


playbooks/openstack-deploy_caasp.yml
```diff
@@ -183,19 +183,32 @@
...
+          retries: 10
+          delay: 10
+          until: _caaspsshkeys.stdout and _caaspsshkeys.stdout.strip()
           loop: "{{ groups['caasp-admin'] + groups['caasp-masters'] + groups['caasp-workers'] }}"
-          args:
-            executable: /bin/bash
 
+
+        - name: Insert known host
+          known_hosts:
```
jsuchome: Wow, there's `known_hosts` module, cool!

jsuchome:
Oh ... travis complains

playbooks/openstack-deploy_caasp.yml
```diff
@@ -183,19 +183,32 @@
...
         - meta: refresh_inventory
 
-        # Waiting for port is not enough, as the port might be up (systemd)
-        # while ssh is not ready. Wait for connection is not good either
-        # because the key is unknown.
-        - name: Wait for a while for SSH to be properly booted up
+        - name: Wait for a SSH to be properly started
           wait_for:
-            timeout: 120
+            host: "{{ hostvars[item]['ansible_host'] | quote }}"
```
evrardjp: Could you remove the `|quote` here?

playbooks/openstack-deploy_caasp.yml
evrardjp: :)

playbooks/openstack-deploy_caasp.yml
```diff
@@ -183,19 +183,32 @@
...
-            executable: /bin/bash
 
+
+        - name: Insert known host
+          known_hosts:
+            state: present
+            name: "{{ hostvars[item.item]['ansible_host'] }}"
+            key: "{{ item.stdout }}"
+          loop: "{{ _caaspsshkeys.results }}"
+
```
evrardjp: Could you remove the trailing whitespace, please?
It will break ansible-lint.

rhafer:
> I had come across the same problem and implemented a very similar solution. I've pushed another commit to your branch to also include the osh_instance, I hope you don't mind.

Cool, thanks. I must have missed that.

playbooks/openstack-deploy_caasp.yml
```diff
@@ -183,18 +183,31 @@
...
+        - name: Wait for a SSH to be properly started
           wait_for:
-            timeout: 120
+            host: "{{ hostvars[item]['ansible_host'] }}"
+            port: 22
+            search_regex: OpenSSH
+            state: started
+            delay: 10
+            timeout: 180
+          loop: "{{ groups['caasp-admin'] + groups['caasp-masters'] + groups['caasp-workers'] }}"
```
Itxaka: should we use a set here so we dont have the same node twice?

Is that even possible, that caasp admin and caasp master are the same for example?

playbooks/openstack-deploy_caasp.yml
```diff
@@ -183,18 +183,31 @@
...
 
         - name: Get pubkey, and add it to known hosts
-          shell: ssh-keyscan -H {{ hostvars[item]['ansible_host'] | quote }} >> ${HOME}/.ssh/known_hosts
+          command: "ssh-keyscan -H {{ hostvars[item]['ansible_host'] | quote }}"
+          register: _caaspsshkeys
+          retries: 10
+          delay: 10
+          until: _caaspsshkeys.stdout and _caaspsshkeys.stdout.strip()
```
Itxaka: shouldnt `_caaspsshkeys.stdout` be enough given than an empty string is falsy in python? Or does ansible understand that empty stdout is still stdout?

playbooks/openstack-deploy_caasp.yml
evrardjp: It is not possible. They are different roles, which don't overlap, IIRC.

playbooks/openstack-deploy_caasp.yml
rhafer: This is mainly to catch the case when stdout is just whitespaces (not sure if this can happen with ssh-keyscan, but "better safe than sorry" :) ) .

PR #91, opened 2019-02-15
Do not read possibly remote file with lookup.

jsuchome:


evrardjp:
I an not sure if I should accept that patch in the meantime, I probably should :p

jsuchome:
I understand this is temporary, but it's a fix for current usage

PR #93, opened 2019-02-18
Developer mode to build images

evrardjp:
This is a refactor of the process for developer mode in order to:
- Fix it, as it simply didn't work for me (inspired by PR#66)
- Increase simplicity by re-using variables
- Remove repetitions by using roles per topic
- Be closer to the openstack-infra way to build image, as this will be the
  future way to configure OSH-images
- Allow a secure registry to be configured by generating self-signed certs.
- Create a separation for CaaSP, for the developer BYO certs or BYO CaaSP case.


evrardjp:
Ready for primetime.

evrardjp:
Rebased -- solved conflicts.

playbooks/roles/suse-build-images/tasks/build-with-docker.yml
```diff
@@ -1,16 +1,29 @@
...
-    src: "{{ image }}-build.sh.j2"
-    dest: "/tmp/build_component.sh"
-    mode: "0755"
-
-# TODO(evrardjp): Cleanup this for idempotency
-# in the meantime, skip_ansible_lint
-- name: Build and push infra component - suse {{ image }} image
-  command: /tmp/build_component.sh
+- name: Push the docker image {{ image.context }}
+  command: docker push {{ image.repository }}:{{ tag }}
```
evrardjp: This {{ tag }} is incorrect and is fixed in a later commit.

playbooks/roles/install-docker/tasks/main.yml
```diff
@@ -0,0 +1,41 @@
...
+    port: "{{ ('http://' + build_registry)| urlsplit('port') }}"
+  docker_container:
+    name: "{{ ('http://' + build_registry)| urlsplit('hostname') }}"
+    image: "registry:2"
+    ports: "{{ port }}:{{ port }}"
+    volumes:
+      - "{{ build_registry_certs_folder }}:/certs"
+    restart_policy: "always"
+    env:
+      REGISTRY_HTTP_TLS_CERTIFICATE: "{{ build_registry_certs_folder }}/{{ build_registry_cert_name }}"
```
evrardjp: This is incorrect and is fixed in a later commit of this series.

playbooks/roles/setup-caasp-workers/tasks/developer-tasks.yml
```diff
@@ -0,0 +1,31 @@
...
+  setup:
+  delegate_to: "{{ item }}"
+  delegate_facts: True
+  loop: "{{ groups['osh-deployer'] }}"
+
+- name: Configure the hosts for local registry if developer mode is enabled
+  lineinfile:
+    path: /etc/hosts
+    line: |
+      "{{ hostvars[groups['osh_deployer'][0]]['ansible_default_ipv4']['address'] | ipaddr('address') }} {{ hostvars[groups['osh_deployer'][0]]['ansible_fqdn'] }} {{ hostvars[groups['osh_deployer'][0]]['ansible_hostname'] }}"
```
evrardjp: Readability of this is improved in another patch in this series.

playbooks/roles/suse-build-images/tasks/main.yml
```diff
@@ -20,91 +26,15 @@
...
-- name: Restart docker if needed
-  service:
-    name: docker
-    state: restarted
-  when: _upd_docker_daemon_conf is defined
-
-- name: Build infra components
-  include_tasks: build-infra-component.yml
-  with_items: "{{ build_infra_components_list }}"
+- name: Build and push the non-LOCI components
```
jsuchome: Wouldn't it be good to mention that those are actually openstack-helm-images

playbooks/roles/generate-certs/handlers/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+# handlers file for generate-certs
```
jsuchome: Is this currently necessary?

playbooks/roles/install-docker/defaults/main.yml
```diff
@@ -0,0 +1,10 @@
+---
+# defaults file for install-docker
+
+# We don't know which default python version will be configured on the remote
+# node, therefore we install both docker packages, so we'll be able to find
+# the python libs anyway.
+docker_packages:
+  - docker
+  - python2-docker
+  - python3-docker
```
jsuchome: Actually I think if you install just `python-dockerpty` it should pick the right version

playbooks/generic-build_images.yml
```diff
@@ -1,8 +1,12 @@
 ---
-- hosts: osh-deployer
+- hosts: osh-deployer[0]
```
jsuchome: There could be more deployers?

playbooks/generic-build_images.yml
```diff
@@ -11,10 +15,29 @@
...
+        tasks_from: copy-certificates
+      when:
+        - developer_mode
+
+    - name: If developer mode, install docker
+      include_role:
+        name: install-docker
+        apply:
+          tags:
+            - docker
```
jsuchome: Where's the check for `developer_mode` ?

playbooks/generic-build_images.yml
```diff
@@ -11,10 +15,29 @@
...
+      when:
+        - developer_mode
+
+    - name: If developer mode, build suse images given in extravars or suse-build-images defaults.
+      include_role:
+        name: suse-build-images
+        apply:
+          tags:
+            - imagebuilder
```
jsuchome: Where's the check for `developer_mode` ?

playbooks/roles/setup-caasp-workers/tasks/osh-deployer-in-hosts.yml
```diff
@@ -0,0 +1,18 @@
+---
+- name: Get the facts from the deployer node from the worker nodes
+  setup:
+  delegate_to: "{{ item }}"
+  delegate_facts: True
+  loop: "{{ groups['osh-deployer'] }}"
+
+- name: Configure the hosts for local registry if developer mode is enabled
```
jsuchome: Where's the check for `developer_mode` ?

playbooks/roles/suse-build-images/tasks/main.yml
evrardjp: They are all openstack-helm-images?

playbooks/roles/generate-certs/handlers/main.yml
evrardjp: It's generated by default by ansible-galaxy :p 
I don't mind removing it. I don't see a point either.

playbooks/roles/install-docker/defaults/main.yml
evrardjp: Mmm what is "the right" version, as it might totally be defined in userspace? I don't know how python-dockerpty is installed, neither if it provides the docker library needed for https://docs.ansible.com/ansible/2.7/modules/docker_container_module.html#docker-container-module . But I would be happy to do a follow up patch to clean that up. I just know this works for now :)

playbooks/generic-build_images.yml
evrardjp: I guess it will depend on how we do things with airship, either it's the same node or not, either we name those the same way or not.

But the idea behind this is to ensure this rolls on a single node at all time, and that would be the node for which we grab the facts. Not another one.

playbooks/generic-build_images.yml
evrardjp: It's an env variable DEVELOPER_MODE. That env var is loaded from the common vars. 
It hasn't changed for a while, and, for once... is in the readme :)

playbooks/generic-build_images.yml
evrardjp: ditto.

playbooks/roles/setup-caasp-workers/tasks/osh-deployer-in-hosts.yml
evrardjp: ditto.

playbooks/roles/suse-build-images/tasks/main.yml
```diff
@@ -94,11 +94,14 @@
...
-  with_items: "{{ build_infra_components_list }}"
+# We should not build from defaults
+# Instead developer should say what to build, and we build it.
+# Developer should say "How to use it", so it can later be tested in env.
+- name: Build the non-LOCI components
+  include_tasks: build-with-docker.yml
+  loop: "{{ build_images_with_dockerfiles_list }}"
   loop_control:
-    loop_var: component
+    loop_var: image
```
Itxaka: that is a really nice feature for clarity!

script_library/deployment-actions-kvm.sh
```diff
@@ -30,6 +30,10 @@ function build_images(){
     echo "Nothing will happen if developer mode is not set"
     run_ansible ${socok8s_absolute_dir}/playbooks/generic-build_images.yml -e "build_osh_images=yes"
 }
+function setup_caasp_workers_for_openstack(){
+    echo "Ensuring caasp workers can be used for openstack"
+    run_ansible ${socok8s_absolute_dir}/playbooks/generic-setup_caasp_workers_for_openstack.yml
```
Itxaka: This will fail the shellcheck travis test, needs to be quoted

script_library/deployment-actions-openstack.sh
```diff
@@ -28,8 +28,8 @@ function deploy_ccp_deployer() {
     run_ansible ${socok8s_absolute_dir}/playbooks/openstack-deploy_ccp_deployer.yml
 }
 function enroll_caasp_workers() {
-    echo "Enrolling caasp worker nodes into the cluster and ensuring they are ready for openstack"
-    run_ansible ${socok8s_absolute_dir}/playbooks/openstack-enroll_caasp_workers.yml
+    echo "Enrolling caasp worker nodes into the cluster"
+    run_ansible ${socok8s_absolute_dir}/playbooks/generic-enroll_caasp_workers.yml
```
Itxaka: ditto

script_library/deployment-actions-openstack.sh
```diff
@@ -41,6 +41,10 @@ function build_images(){
     echo "Nothing will happen if developer mode is not set"
     run_ansible -e "build_osh_images=yes" ${socok8s_absolute_dir}/playbooks/openstack-build_images.yml
 }
+function setup_caasp_workers_for_openstack(){
+    echo "Ensuring caasp workers can be used for openstack"
+    run_ansible ${socok8s_absolute_dir}/playbooks/generic-setup_caasp_workers_for_openstack.yml
```
Itxaka: ditto

playbooks/roles/suse-build-images/tasks/main.yml
jsuchome: I mean: unlike LOC components, these are coming from different source, openstack-helm-images project

Itxaka:
Could we remove all the empty files?

playbooks/roles/install-docker/defaults/main.yml
jsuchome: The right version is either python2-docker or python3-docker, I guess you do not need both.

It might be better to just use `python-docker` actually.

playbooks/generic-build_images.yml
jsuchome: I understand we can theoretically have more deployer nodes, I'm just suprised that only this playbook specifically chooses the first item

playbooks/generic-build_images.yml
Itxaka: FYI this gets into the playbooks from here: https://github.com/SUSE-Cloud/socok8s/blob/master/vars/common-vars.yml#L3

jsuchome: I understand that. I was missing `when:  - developer_mode` condition, I don't know how I missed that

playbooks/roles/install-docker/defaults/main.yml
evrardjp: oh python-docker is pointing to the right default python version?
That would be neat.

If someone is not running dual stack python, it means we are installing a package too much here. But that's not really the problem. The problem is that we can't predict reliably what will be used for the python on that machine because:
- Some people could have made the system dual stack by installing python3 (IIRC 15 is still python2 by default) 
- The deployer can decide with python version to use in the inventory
- Ansible will fallback to local ansible's python version if nothing is defined. Which can be python3 for a modern OS, or python2 for a less modern one :p . But we can't predict that -- IIRC.

What we could do instead, is have a stance: "Please install x on this machine, and configure inventory appropriately"

evrardjp: Edit: 15 has no python by default in my container, so I guess we can use whatever we want. It's probably worth testing though. Can we do that in a different PR?

playbooks/generic-build_images.yml
evrardjp: That's because that's the only one to rely on a fact that's set on the first node :)

playbooks/roles/install-docker/defaults/main.yml
jsuchome: I'm fine with that

playbooks/roles/suse-build-images/tasks/main.yml
evrardjp: Solved in PR96.

playbooks/roles/install-docker/defaults/main.yml
evrardjp: Solved in PR97.

evrardjp:
> Some minor nits.
> 
> I would prefer to have smaller PRs that tackle one thing at a time, so its easier to review them.
> 
> Also I have a feeling that some of this commit should be merged together, as if not, if we revert some of them we would lead the repo in a broken state, which is bad. IMO you should be able to apply a commit and revert it and it should still leave the repo working, but Im not gonna block this due to that as it would be a PITA to rework the commits :)

I think they are pretty much independant. I can isolate a few of those that would cleanly apply. But then there is a huge chunk that would need to go in, which is hard to review in one bit. This is why it's done in 4 commits or so, each one focusing on a single thing. "Add docker config/install" or "Configure registry". Sadly bringing only one of those would leave this repo in a bad state, which I believe is not the best thing. I think at the end, this focused on "one goal" (making the developer mode work for image b uilding in a simple way for dev), and therefore it's acceptable as one PR, IMO.

script_library/deployment-actions-kvm.sh
evrardjp: I can try having a go (pun intended as it's a bash script) at wrapping those.

evrardjp:
@Itxaka @jsuchome @rhafer should I work on this more, or should we go for extra patches in different PRs?

evrardjp:
> I would prefer if commits that fix issues in previous commits would be squashed, but it's probably better to be over this one already and focus on followups

I think that's done? All the "fix x" were squashed.

jsuchome:
> I think that's done? All the "fix x" were squashed.

It's possible, sorry :-) It's the number of commits that may have confused me

PR #95, opened 2019-02-19
Do not pip install openstack client, use regular SUSE packages

jsuchome:
Provide custom 010-setup-client.sh that does do installation,
only the configration part.

### DIFF
--- a/playbooks/roles/deploy-osh/tasks/main.yml
+++ b/playbooks/roles/deploy-osh/tasks/main.yml
@@ -130,17 +130,43 @@
     - horizon
     - run
 
-
-# TODO(evrardjp): Use mechanism like glance to re-use component-install.yml code and allow any override.
-# TODO(evrardjp): Check stdout to ensure idempotency. In the meantime skip_ansible_lint
-- name: Setup openstack CLIs
-  command: ./tools/deployment/multinode/010-setup-client.sh
-  args:
-    executable: /bin/bash
-    chdir: /opt/openstack-helm
+- name: Install openstack client
+  package:
+    name: python-openstackclient
+    state: present
+  register: _pkginstall
+  until: _pkginstall is success
+  retries: 5
+  delay: 2
   tags:
     - run
-    - skip_ansible_lint
+
+- name: Create .config/openstack directory
+  file:
+    path: /root/.config/openstack
+    state: directory
+    owner: root
+    group: root
+    mode: '0755'
+  tags:
+    - run
+
+- name: Save default clouds.yaml file
+  template:
+    src: clouds.yaml.j2
+    dest: /root/.config/openstack/clouds.yaml
+    owner: root
+    mode: '0640'
+
+- name: Build openstack-helm charts
+  make:
+    chdir: /opt/openstack-helm
+    target: all
+  when:
+    - not developer_mode
+    - _gitclone is changed
+  tags:
+    - run
 
 # TODO(evrardjp): Follow on https://review.openstack.org/#/c/605005/
 - name: Use a different ingress script as long as upstream does not support overrides
--- a/playbooks/roles/deploy-osh/templates/clouds.yaml.j2
+++ b/playbooks/roles/deploy-osh/templates/clouds.yaml.j2
@@ -1,0 +1,11 @@
+clouds:
+  openstack_helm:
+    region_name: RegionOne
+    identity_api_version: 3
+    auth:
+      username: 'admin'
+      password: 'password'
+      project_name: 'admin'
+      project_domain_name: 'default'
+      user_domain_name: 'default'
+      auth_url: 'http://keystone.openstack.svc.cluster.local/v3'


playbooks/roles/deploy-osh/tasks/main.yml
```diff
@@ -130,8 +130,25 @@
     - horizon
     - run
 
+- name: Install openstack client
+  package:
+    name: python-openstackclient
```
evrardjp: could you convert that to a list with two packages?
- python-openstackclient
- python-heatclient

As the latter has a dependency on the former, not the other way around, IIRC.

playbooks/roles/deploy-osh/files/010-setup-client.sh
```diff
@@ -0,0 +1,36 @@
...
+      username: 'admin'
+      password: 'password'
+      project_name: 'admin'
+      project_domain_name: 'default'
+      user_domain_name: 'default'
+      auth_url: 'http://keystone.openstack.svc.cluster.local/v3'
+EOF
+
+#NOTE: Build charts
+make all
```
evrardjp: I would prefer if we make all the charts explicitly, rather than in here. We are building the osh-infra charts in main anyway, it would be just nice to have the osh charts built afterwards, and more explicit. The airship squad could add their repos too.

evrardjp: That's more a nit than a problem for me, if we keep the file. Just saying :)

playbooks/roles/deploy-osh/tasks/main.yml
dirkmueller: heatclient certainly does not depend on openstackclient, its the other way, aka the way jiri did it.

playbooks/roles/deploy-osh/files/010-setup-client.sh
jsuchome: OK

playbooks/roles/deploy-osh/tasks/main.yml
```diff
@@ -130,17 +130,44 @@
...
+    owner: root
+    group: root
+    mode: 0755
+  tags:
+    - run
+
+- name: Save default clouds.yaml file
+  copy:
+    srt: clouds.yaml
+    dest: /etc/openstack/clouds.yaml
```
dirkmueller: why do we need to install that system wide rather than in the home directory?

jsuchome: I guess we don't really _need_ it. This is just copying the functionality of helm's `010-setup-client.sh`

jsuchome: That said, I think I could put it under `~/.config/openstack/` , unless there really is some reason to have it in global location.

playbooks/roles/deploy-osh/tasks/main.yml
evrardjp: that's interesting, I didn't know openstackclient bundled everything in this packaging. I double checked and it's the case indeed.

playbooks/roles/deploy-osh/tasks/main.yml
evrardjp: I think both are fine. Let's use home, it seems cleaner :)
We'll see if it eventually conflicts with the expectations of other things upstream too, and we could improve upstream then.

playbooks/roles/deploy-osh/tasks/main.yml
```diff
@@ -130,17 +130,44 @@
...
+    path: /etc/openstack
+    state: directory
+    owner: root
+    group: root
+    mode: 0755
+  tags:
+    - run
+
+- name: Save default clouds.yaml file
+  copy:
```
evrardjp: I would use template: here, so that we can change it later with a few overrides, should that path be pursued.

playbooks/roles/deploy-osh/tasks/main.yml
```diff
@@ -130,17 +130,44 @@
...
+    srt: clouds.yaml
```
evrardjp: I think that's `src:` here

jsuchome: yep, i've spotted it as well

playbooks/roles/deploy-osh/tasks/main.yml
jsuchome: ok, I'll adapt it

playbooks/roles/deploy-osh/tasks/main.yml
jsuchome: It is not bundled in one package, it's brought by dependency

playbooks/roles/deploy-osh/tasks/main.yml
jsuchome: OK

playbooks/roles/deploy-osh/tasks/main.yml
```diff
@@ -130,17 +130,43 @@
     - horizon
     - run
 
+- name: Install openstack client
+  package:
+    name: python-openstackclient
+    state: present
+  register: _pkginstall
```
Itxaka: do we really need all this stuff with retries? This should only be used if we expect issues but why would we expect issues on a simple package install?

jsuchome: usually not, but it could theoretically happen that some zypper process is already running...

jsuchome: So I'm just using same values I've found elsewhere

Itxaka: shouldnt then we use the zypper module with `environment:
    ZYPP_LOCK_TIMEOUT: 20` as options to suse-ify it more??

Itxaka: substitute 20 with the wanted value of course :)

jsuchome: Might be good idea to do extra PR that would replace all package installations in our plays

evrardjp: > do we really need all this stuff with retries? This should only be used if we expect issues but why would we expect issues on a simple package install?

network related issues. That's why we brought that into ansible-lint.
ansible-lint will complain if you don't bring that.

Let's not make it extra hard with locks, just make it retries, so we don't have to change that everywhere. If we need to change the module to have this ZYPP_LOCK_TIMEOUT by default, we can too.

playbooks/roles/deploy-osh/tasks/main.yml
```diff
@@ -130,17 +130,43 @@
...
-    executable: /bin/bash
+- name: Create .config/openstack directory
+  file:
+    path: /root/.config/openstack
+    state: directory
+    owner: root
+    group: root
+    mode: '0755'
+  tags:
+    - run
```
evrardjp: nit: technically that can still be install.

playbooks/roles/deploy-osh/tasks/main.yml
```diff
@@ -130,17 +130,43 @@
...
+
+- name: Save default clouds.yaml file
+  template:
+    src: clouds.yaml.j2
+    dest: /root/.config/openstack/clouds.yaml
+    owner: root
+    mode: '0640'
+
```
evrardjp: nit: while this should be tagged `run`.

playbooks/roles/deploy-osh/tasks/main.yml
```diff
@@ -130,17 +130,43 @@
...
+- name: Build openstack-helm charts
+  make:
     chdir: /opt/openstack-helm
+    target: all
+  when:
+    - not developer_mode
+    - _gitclone is changed
```
evrardjp: that should probably go away. If developer mode is used, the charts need to be built anyway, and the clone would not be changed. But that can be in a different commit.

PR #97, opened 2019-02-19
Do not try to dual stack python


evrardjp:
Just install the one that works by default.


evrardjp:
#93 needs to be in first. Another vote is required though.

jsuchome:
You can rebase and merge

PR #98, opened 2019-02-19
Fix registry URL for LOCI builds


evrardjp:
The LOCI registry link contains two slashes in a row, which would not
be allowed by docker. This fixes it.


evrardjp:
This is not enough to "fix" the building of images using LOCI, but is a first start.
The next required work would be to merge the following patches:
- https://review.openstack.org/#/c/637963/
- https://review.openstack.org/#/c/637168/
and then add the extra arguments for insecure SSL in the environment variables of `playbooks/roles/suse-build-images/tasks/build-with-loci.yml`

PR #99, opened 2019-02-20
Minor fixes for image building documentation

jsuchome:
Feel free to merge these into different documentation PR if there's one

### DIFF
--- a/playbooks/roles/suse-build-images/defaults/main.yml
+++ b/playbooks/roles/suse-build-images/defaults/main.yml
@@ -11,7 +11,7 @@
 # Full story on how to use this:
 # 0) Set developer mode. It should be already like this if you're entering
 #    this role.
-# 1) Override this variable by providing a `docker_image`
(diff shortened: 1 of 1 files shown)


playbooks/roles/suse-build-images/defaults/main.yml
```diff
@@ -65,7 +67,7 @@ docker_images: []
 # suse_osh_deploy_libvirt_yaml_overrides:
 #   images:
 #     tags:
-#       libvirt: "<insert build server url here, probably local-registry:5000>/openstackhelm/libvirt:latest-opensuse_15"
+#       libvirt: "{{ build_registry }}openstackhelm/libvirt:latest-opensuse_15"
```
jsuchome: Actually `{{ build_registry }}` is not known at this step, so until we fix it, there needs to be full url, like `$USERNAME-osh.openstack.local:5000`

PR #100, opened 2019-02-20
Restore caasp automation from kubic-project

Itxaka:
It seems that the repo has gone missing so restore the files while we
look for a replacement


jsuchome:
+1 until we find out what's wrong ... does anyone know where to ask?

Itxaka:
I asked on the #kubic room in freenode, everything seems to point there as a contact point

Itxaka:
oh my, we also need to copy the `velum-interactions` tool :crying_cat_face:

evrardjp:
That behaviour is very bad. I am kinda disappointed nobody checked on us before completely deleting (vs archiving).

Itxaka:
from the #kubic chat:

`<fvogt> itxaka: It was removed because it wasn't used by and for kubic - you should ask about that internally`

so we may need to escalate this internally with the kubic team :)

evrardjp:
I think it's fair, I just hoped they would "redirect" or "change ownership" vs deleting it.

jsuchome:
Do they guarantee the functionality, so we can assume it works if we have it in our own repo?

Itxaka:
This seems to be ready for review, I tested the modified step `enroll_caasp_workers` and it works as it did before.

Currently testing `deploy_caasp`

Itxaka:
deploy_caasp also run correctly

PR #101, opened 2019-02-21
Ensure helm charts are built


evrardjp:
Whether the developer mode is enabled or not, the helm charts need to be built.
On top of that, when developer mode is used, it currently skips the build
as the git clone isn't changed.


evrardjp:
I will rebase this.

PR #102, opened 2019-02-21
Add base docs

evrardjp:


jsuchome:
So the README content was _not_ changed in this PR, right?

evrardjp:
Nope, just converted from md to rst using pandoc.

doc/test-requirements.txt
```diff
@@ -0,0 +1 @@
+#
```
Itxaka: can we drop this file?

evrardjp: I guess I used what is the convention, but we can ofc remove it in a later patchset.

PR #103, opened 2019-02-21
disable debug for ansible on developer mode

Itxaka:
Currently the developer mode is linked into an increase in ansible
verbose AND a change to the stdout callback

Instead of that, have a different var for it as developer mode is
needed for some roles (build_images) but you dont necessarily want
extra verbosity or you stdout_callback to be overwritten

### DIFF
--- a/doc/source/user/ansible_tips.rst
+++ b/doc/source/user/ansible_tips.rst
@@ -1,0 +1,32 @@
+==============================
+Ansible tips
+==============================
+
+
+There is several variables that you can use to get extra debugging from the ansible playbooks as shown in https://docs.ansible.com/ansible/latest/reference_appendices/config.html
+
+Some examples:
+
+.. code-block:: console
+
+    export ANSIBLE_VERBOSITY=3
+    export ANSIBLE_STDOUT_CALLBACK=debug
+
+Both of this environment variables will enable a lot of verbosity and debug for all the playbooks being run
+
+
+You can also enable the debugger for failed tasks as shown in https://docs.ansible.com/ansible/latest/user_guide/playbooks_debugger.html
+
+.. code-block:: console
+
+    export ANSIBLE_ENABLE_TASK_DEBUGGER=True
+
+
+This will drop you into the debugger when a task fails so you can examine the task, vars, retry it and so on. Make sure to check the ansible docs linked for all available options
+
+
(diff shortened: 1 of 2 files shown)


Itxaka:
Seems fair @evrardjp 

I'll wait for the doc PRs to get in and rework the PR to drop + document if needed those vars (they come from ansible so IMO they should not be documented on our side)

evrardjp:
@Itxaka I agree. Maybe in a tips and tricks in user stories, pointing to ansible's docs?

evrardjp:
the docs PR merged, so it might be good to revisit this?

Itxaka:
Updated!

Itxaka:
> Could you add the doc change, or at least point to another PR for the doc change, please?

I dont think that part has a place in our docs being generic to ansible. I mean, why stop on those vars? there is plenty of helpful ansible docs that could apply to our repo, so I dont think we should have any copy of those here.

Itxaka:
added a doc for debugging with some small examples and links to ansible docs. Dont look too good for me but at least we can probably expand the docs with more debugging tips for other parts (caasp, k8s, helm?)

evrardjp:
> added a doc for debugging with some small examples and links to ansible docs. Dont look too good for me but at least we can probably expand the docs with more debugging tips for other parts (caasp, k8s, helm?)

That's exactly what I believe our docs could help. We can't presume a deep knowledge of all the tools, so why not make user's experience better here?

doc/source/user/debugging.rst
```diff
@@ -0,0 +1,32 @@
+==============================
+Debugging
```
evrardjp: Could you rename this to:
`Increase Ansible verbosity` or `Ansible tips`?

doc/source/user/debugging.rst
```diff
@@ -0,0 +1,32 @@
...
+==============================
+
+
+There is several variables that you can use to get extra debugging from the ansible playbooks as shown in https://docs.ansible.com/ansible/latest/reference_appendices/config.html
+
+Some examples:
+
+.. code-block:: console
+
+    ANSIBLE_VERBOSITY = 3
```
evrardjp: I think you should remove the spaces, else it might be a problem.

You should also add `export` at the beginning in each line.

doc/source/user/debugging.rst
Itxaka: Yeah I wanted a bit of feedback here. This may be just the debugging page and then a subsection on ansible or we could have different pages for debugging the different components.

Im ok with either, maybe separated is better in case some pages get too long

PR #104, opened 2019-02-22
Restructured documentation


evrardjp:
To be applied on top of PR#102.

PR #106, opened 2019-02-25
Document helm chart values discovery


evrardjp:
Here is how I check which helm charts values are available for
overriding.


jsuchome:
Is the new `discover-helm-charts-overrides.rst` file linked from any other place?

evrardjp:
Automatically from a glob.

evrardjp:
Added DNM so that we know #104  needs to merge first.

PR #109, opened 2019-02-27
travis: publish docs

Itxaka:
Publish docs when merged into the master branch to github pages

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,12 +2,27 @@ language: python
 cache: pip
 dist: xenial
 
-matrix:
+jobs:
   include:
     - name: "shell syntax"
+      stage: tests
       script: find . -path "./submodules" -prune -o -type f -name "*.sh" -print | xargs -n1 shellcheck -x --format=gcc
     - name: "playbook syntax"
+      stage: tests
       script: ansible-lint playbooks/*.yml
+    - name: "build and publish docs"
+      stage: publish docs
+      install:
+        - pip install tox
+      script: tox -e docs
+      deploy:
+        provider: pages
+        skip_cleanup: true
+        github_token: $token # set in the Travis dashboard
+        local_dir: doc/build/html
+        on:
+          branch: master
+
   allow_failures:
     - name: "shell syntax"
 
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -27,6 +27,7 @@ extensions = [
     'sphinxcontrib.blockdiag',
     'sphinxcontrib.seqdiag',
     'sphinxcontrib.nwdiag',
+    'sphinx.ext.githubpages',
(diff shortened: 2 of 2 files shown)


Itxaka:
This *needs* a token from a user with access to the repo as it pushes to the `gh-pages` branch. The token should be created by a bot/automated user and added to travis as `token` var

Not sure who to talk to or who is in charge of the bots

Itxaka:
Also, it works really nice: https://itxaka.github.io/socok8s/deploy-guide/index.html

.travis.yml
```diff
@@ -2,12 +2,27 @@ language: python
...
+      stage: tests
       script: find . -path "./submodules" -prune -o -type f -name "*.sh" -print | xargs -n1 shellcheck -x --format=gcc
     - name: "playbook syntax"
+      stage: tests
       script: ansible-lint playbooks/*.yml
+    - name: "build and publish docs"
+      stage: publish docs
+      install:
+        - pip install tox
+      script: tox
```
evrardjp: can you change this to tox -e docs ?
Tox might grow with reno and linting.

.travis.yml
```diff
@@ -2,12 +2,27 @@ language: python
 cache: pip
 dist: xenial
 
-matrix:
```
evrardjp: Could you explain why this movesfrom matrix to jobs?

.travis.yml
Itxaka: :+1:

.travis.yml
evrardjp: I meant in the commit message, not here :)

Itxaka: To move this into different stages to we dont build the docs if there is a failure on the tests, mostly it gains us time and we can reuse it later for extra steps that we dont necessarily want to run unless the linting is passed first (like unit tests that take longer? molecule testing?)

Kind of like the jenkins pipeline, if step X fails do not move into the next one.

Itxaka: oh, that makes sense....

evrardjp:
@Itxaka still don't merge?

Itxaka:
> @Itxaka still don't merge?

yup yup, need to find out who is the bot owner for github to obtain a token and put it on the repo secret vars in travis so it can push to github pages. Will try to do it today as its currently using my own token, which is bad :)

toabctl:
@Itxaka any update on this?

Itxaka:
> @Itxaka any update on this?

@toabctl unfortunately nope, I could not get the owner of the susegithubapi account to appear :/

I guess we can move this forward and use one of my tokens meanwhile? That usually leads to forgetting about it until it breaks thought :D

Itxaka:
This will be moved into jenkins instead

PR #112, opened 2019-02-28
Updated ses_config.yml file generation to DeepSea openstack.integrate format

matthew-fuller:
Modified ses_config.yml.j2 template and added a task to make the output ses_config.yml file from the ses-osh role match the DeepSea openstack.integrate output format. Tested and verified locally (KVM), but not on eng cloud...

### DIFF
--- a/playbooks/roles/ses-osh/tasks/main.yml
+++ b/playbooks/roles/ses-osh/tasks/main.yml
@@ -11,6 +11,15 @@
   with_items:
     - "{{ hostvars[groups['ses_nodes'][0]].ceph_files.results }}"
 
+- name: Get key values for ses-config.yml
+  set_fact:
+    "ceph_{{ item | regex_replace('[\\.\\-]', '_') }}_key":
+       "{{lookup('ini', 'key section={{ item }} file={{ socok8s_workspace }}/ceph.{{ item }}.keyring') }}"
+  with_items:
+    - "client.cinder"
+    - "client.cinder_backup"
+    - "client.glance"
+
 - name: Generate ses-config.yml
   delegate_to: localhost
   template:
--- a/playbooks/roles/ses-osh/tasks/main.yml
+++ b/playbooks/roles/ses-osh/tasks/main.yml
@@ -11,15 +11,6 @@
   with_items:
     - "{{ hostvars[groups['ses_nodes'][0]].ceph_files.results }}"
 
-- name: Get key values for ses-config.yml
-  set_fact:
-    "ceph_{{ item | regex_replace('[\\.\\-]', '_') }}_key":
-       "{{lookup('ini', 'key section={{ item }} file={{ socok8s_workspace }}/ceph.{{ item }}.keyring') }}"
-  with_items:
-    - "client.cinder"
-    - "client.cinder_backup"
-    - "client.glance"
-
 - name: Generate ses-config.yml
   delegate_to: localhost
   template:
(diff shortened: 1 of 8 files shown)


matthew-fuller:
PR #113 addresses this problem (and accounts for a problem parsing the variable name 'cinder-backup' in the new format due to the hyphen).

playbooks/roles/ses-osh/tasks/main.yml
```diff
@@ -11,6 +11,15 @@
   with_items:
     - "{{ hostvars[groups['ses_nodes'][0]].ceph_files.results }}"
 
+- name: Get key values for ses-config.yml
+  set_fact:
+    "ceph_{{ item | regex_replace('[\\.\\-]', '_') }}_key":
+       "{{lookup('ini', 'key section={{ item }} file={{ socok8s_workspace }}/ceph.{{ item }}.keyring') }}"
+  with_items:
```
evrardjp: Shouldn't this be based on the ceph_files.results too, to avoid having an item list that's out of date, and ensuring everything is properly fetched?

playbooks/roles/ses-osh/tasks/main.yml
```diff
@@ -11,6 +11,15 @@
   with_items:
     - "{{ hostvars[groups['ses_nodes'][0]].ceph_files.results }}"
 
+- name: Get key values for ses-config.yml
+  set_fact:
+    "ceph_{{ item | regex_replace('[\\.\\-]', '_') }}_key":
```
evrardjp: Where is this fact used? I am confused why this needs to be a set fact.

matthew-fuller:
I closed PR 113 and moved its changes into this PR since they all pertained to the new ses_config format. Per JP's request in PR 113, I changed the "Get ses_cluster_configuration" task in deploy-osh/tasks/main.yml to use include_vars rather than using filtering.

playbooks/roles/ses-osh/tasks/main.yml
matthew-fuller: I tried to leave most of the original PR (36) intact. These variables are used in the ses_config.yml.j2 template when constructing the new ses_config file, and they were originally based on the names of the ses_openstack_keys in the ses-ansible role, which are hard coded to "client.glance", "client.cinder", and "client.cinder_backup". However, those keys don't seem to be accessible from the ses-osh role, so I simply made a list of the names. I'm certainly open to changing / improving this, however.

evrardjp: Oh yeah I see now. 

Wouldn't it be simpler to do the lookups directly in the template? It would be far easier to read/parse.

playbooks/roles/ses-osh/templates/ses-config.yml.j2
```diff
@@ -1,28 +1,24 @@
...
-  ses_cluster_name: ceph
+---
+ceph_conf:
+  cluster_network: {{ hostvars[groups['ses_nodes'][0]].conf_options.stdout_lines.4 }}
+  fsid: {{ hostvars[groups['ses_nodes'][0]].conf_options.stdout_lines.0 }}
+  mon_host: {{ hostvars[groups['ses_nodes'][0]].conf_options.stdout_lines.2 }}
+  mon_initial_members: {{ hostvars[groups['ses_nodes'][0]].conf_options.stdout_lines.1 }}
+  public_network: {{ hostvars[groups['ses_nodes'][0]].conf_options.stdout_lines.3 }}
+cinder:
+  key: {{ ceph_client_cinder_key }}
```
evrardjp: I meant with the comment above you can replace the set_fact with the lookup directly here, and avoid the messing up with the input/set fact :)

evrardjp:
Close to perfect to me. I'd remove the set fact for future reviewability

matthew-fuller:
Following JP's recommendation, I've removed the task "Get key values for ses-config" entirely and moved the client key lookups to the ses_config.yml.j2 template directly.

evrardjp:
I haven't tested it, but it's simple and looks great.

Itxaka:
merge?

PR #113, opened 2019-02-28
Revise osh templates and example ses_config.yml file

matthew-fuller:
These changes revise the cinder, glance, libvirt, and nova templates to account for the new ses_config.yml format. The example file in workdir/examples has also been updated.

### DIFF
--- a/playbooks/roles/deploy-osh/tasks/main.yml
+++ b/playbooks/roles/deploy-osh/tasks/main.yml
@@ -9,7 +9,8 @@
     - always
 
 - name: Get ses_cluster_configuration
-  include_vars: "{{ socok8s_ses_pools_details }}"
+  set_fact:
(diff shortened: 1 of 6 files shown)


playbooks/roles/deploy-osh/tasks/main.yml
```diff
@@ -9,7 +9,8 @@
     - always
 
 - name: Get ses_cluster_configuration
-  include_vars: "{{ socok8s_ses_pools_details }}"
```
evrardjp: Could we keep `include_vars:` I think it would be advisable instead of filtering.

The code would be:
```
include_vars:
  file: "{{ socok8s_ses_pools_details }}"
  name: ses_cluster_configuration

evrardjp:
On top of my previous comment, this will effectively break tests, as the interface is changed, but the template generation was not adapted. I would add a commit on top of these to convert the existing template generation to the new format.

PR #114, opened 2019-03-01
Appliying conflicting changes from ucp branch to master

kantarun:
Syncing common variable change from ucp branch into master.
Now variables are named to socok8s instead of having osh_ or
airship_ namespace in prefix of their names.
Variables which are not shared between and are unique to osh
or airship deployment are left untouched.

Also getting changes from roles which are common and are changed in
ucp branch to bring attention in this review.


kantarun:
Added PR from wrong branch by mistake so closing it.

PR #115, opened 2019-03-01
Applying conflicting changes from ucp branch to master

kantarun:
Syncing common variable change from ucp branch into master.
Now variables are named to socok8s instead of having osh_ or
airship_ namespace in prefix of their names.
Variables which are not shared between and are unique to osh
or airship deployment are left untouched.

Also getting changes from roles which are common and are changed in
ucp branch to bring attention in this review.

Adding group_vars file for common variables shared across various
roles so that they don't need to be explictly included in various
roles where needed. Currently common_vars are used for this purpose
which can be phased out if group_vars work for our need.


doc/source/deploy-guide/configure-environment.rst
```diff
@@ -104,15 +104,15 @@ You can find examples of the required files in `examples/workdir`.
 Configure the VIP that will be used for OpenStack
 -------------------------------------------------
 
-Add `suse_osh_deploy_vip_with_cidr:` with its appropriate value for your
+Add `suse_socok8s_deploy_vip_with_cidr:` with its appropriate value for your
```
jgu17: Based on the references below to this vip variable, I think you meant "socok8s_deploy_vip_with_cidr"? I don't have strong opinion on names, so this would be a nitpick you can ignore: socok8s_deploy even seems to a little redundant, socok8s_vip_with_cidr seems to be pretty clear. Also fyi, I think Scott added multiple vips this afternoon, so we have public/external vip, dcm vip, similar to the names in Cloud 9.

playbooks/roles/deploy-osh/tasks/kubectl-install.yml
```diff
@@ -61,12 +63,20 @@
...
+        path: "{{ ansible_user_dir }}/.kube/config"
+      delegate_to: localhost
+      register: old_config
 
     - name: Copy local runner kubeconfig file to deploy node
       copy:
         src: "{{ kubeconfig_file_path }}"
         dest: "{{ ansible_user_dir }}/.kube/config"
         mode: "0600"
+      when: new_config.stat.exists and (old_config.stat.exists == False or old_config.stat.mtime < new_config.stat.mtime)
```
jgu17: lint complains this line -- looks like my bad though... I think the lint wants it to "not old_config.stat.exists" instead.

script_library/run-ansible.sh
```diff
@@ -6,8 +6,8 @@ function run_ansible(){
     set -x
     # ansible-runner default locations
     if [[ -z ${ANSIBLE_RUNNER_DIR+x} ]]; then
-        echo "ANSIBLE_RUNNER_DIR env var is not set, defaulting to '~/suse-osh-deploy'"
-        export ANSIBLE_RUNNER_DIR="${HOME}/suse-osh-deploy"
+        echo "ANSIBLE_RUNNER_DIR env var is not set, defaulting to '~/suse-socok8s-deploy'"
```
jgu17: is suse and socok8s redundant? "soc" could even be short and sweet.  but again, I don't want to be too nitpicky on names, so ignore me if you want to :-).

doc/source/deploy-guide/configure-environment.rst
evrardjp: In that case, we can probably have 
``socok8s_external_vip_with_cidr``, ``socok8s_dcm_vip_with_cidr``, and so on?

examples/workdir/env/extravars
```diff
@@ -6,4 +6,8 @@ ceph_user_keyring_b64key: QVFDZWoySmNBQUFBQUJBQWlhSTZJZEw2Nml4S1FINEVlOTBTbXc9PQ
 suse_osh_deploy_ceph_mons: ['192.168.50.10:6789']
 # This is the VIP that will be used for the openstack deployment on kubernetes
 # (ingress controller)
-suse_osh_deploy_vip_with_cidr: 192.168.50.46/24
+# Deployment goal is either 'airship' or 'osh'
+socok8s_deployment_goal: airship
```
evrardjp: Two things here:
1) If this PR is about common naming, we shouldn't add this variable here, but in a later commit.
2) Why does this variable exist? I fail to see the grand scheme of things for multiple reasons: A) deployers can run the appropriate code by running the right commands (see docs). B) It will be more complex and error prone for users. C) This variable is not documented.

Can you remove this variable, and the variables in L12-L13?

playbooks/generic-build_images.yml
```diff
@@ -10,7 +10,7 @@
     #http://jinja.pocoo.org/docs/2.10/templates/#default
     - name: Check if developer mode should be enabled
       set_fact:
-        developer_mode: "{{ (lookup('env','OSH_DEVELOPER_MODE') | default('False', true) ) | bool }}"
+        developer_mode: "{{ (lookup('env','SOCOK8S_DEVELOPER_MODE') | default('False', true) ) | bool }}"
```
evrardjp: nit: I think a follow-up patch can change that by loading the common vars, and setting fact for it, to prevent duplication. We could even remove the set_fact if developer_mode is changed appropriately everywhere and loading the common vars. But it's okay for now.

playbooks/group_vars/all.yml
```diff
@@ -0,0 +1,49 @@
+---
```
evrardjp: While I agree a group_var is very convenient to use, I don't see a reason to load that on all nodes.

Likewise, there seem to be a big overlap with common-vars.

Could you edit common-vars instead, focusing on the existing vars to migrate to this new location?

playbooks/group_vars/all.yml
```diff
@@ -0,0 +1,49 @@
...
+upstream_repos_clone_folder: "/opt"
+suse_osh_registry_location: "{{ lookup('env','OPENSTACK_HELM_IMAGE_REGISTRY_LOCATION') | default('docker.io', true) }}"
```
evrardjp: Note: I don't see why we set up this var, as the registry location is set for developer mode already. It works.

playbooks/group_vars/all.yml
```diff
@@ -0,0 +1,49 @@
...
+suse_osh_image_version: "{{ lookup('env','OPENSTACK_HELM_IMAGE_VERSION') | default('pike', true) }}"
```
evrardjp: This is, for me, very brittle.
A) It means the group vars hold version, which is less semantically correct than the deploy-osh role. If we want a common variable, can we use something in the manifest?
B) We should make it very intentional to override the default branch, so it needs to be in extra vars IMO. On top of that, a deployer won't store in a file the current deployment version if it's defined as an env var (it ends up by reducing reproducibility for the env var case)
C) This is a new feature and should be added in a separate commit.

playbooks/group_vars/all.yml
```diff
@@ -0,0 +1,49 @@
...
+suse_libvirt_image_version: "{{ lookup('env','OPENSTACK_HELM_LIBVIRT_IMAGE_VERSION') | default('ubuntu-xenial-1.3.1-1ubuntu10.24', true) }}"
```
evrardjp: ditto

playbooks/group_vars/all.yml
```diff
@@ -0,0 +1,49 @@
...
+suse_ovs_image_version: "{{ lookup('env','OPENSTACK_HELM_OPENVSWITCH_IMAGE_VERSION') | default('v2.8.1', true) }}"
+
+developer_mode: "{{ (lookup('env','SOCOK8S_DEVELOPER_MODE') | default('False', true) ) | bool }}"
```
evrardjp: already defined in common vars.

playbooks/group_vars/all.yml
```diff
@@ -0,0 +1,49 @@
...
+
+socok8s_workspace: "{{ lookup('env','ANSIBLE_RUNNER_DIR') | default('~/suse-socok8s-deploy', true) }}"
```
evrardjp: can you use the common vars here?

playbooks/group_vars/all.yml
```diff
@@ -0,0 +1,49 @@
...
+socok8s_extravars: "{{ socok8s_workspace }}/env/extravars"
+socok8s_ses_pools_details: "{{ socok8s_workspace }}/ses_config.yml"
+socok8s_registry_cert: "{{ socok8s_workspace }}/certs/domain.crt"
+socok8s_registry_certkey: "{{ socok8s_workspace }}/certs/domain.key"
+socok8s_caasp_environment_details: "{{ socok8s_workspace }}/environment.json"
+socok8s_libvirtuuid: "{{ socok8s_workspace }}/libvirt.uuid"
+
+socok8s_deploy_config_location: "{{ socok8s_workspace }}"
+socok8s_airship_deploy_site_name: soc-minimal
```
evrardjp: new feature, different PR?

playbooks/group_vars/all.yml
```diff
@@ -0,0 +1,49 @@
...
+
+ucp_namespace_name: "{{ ucp_namespace | default('ucp') }}"
```
evrardjp: shouldn't that be a default in your role(s)/playbooks to deploy airship?

That sound unnecessary for a group var.

playbooks/group_vars/all.yml
```diff
@@ -0,0 +1,49 @@
...
+
+suse_distro_identifier: opensuse
```
evrardjp: Why is this here? New feature -> new PR

playbooks/group_vars/all.yml
```diff
@@ -0,0 +1,49 @@
...
+suse_airship_components_base_image: "opensuse/leap:15.0"
```
evrardjp: Can this be a role default? New feature -> new PR

playbooks/group_vars/all.yml
```diff
@@ -0,0 +1,49 @@
...
+
+# check if argument is set in commandline for airship_local_images_flag value
+# otherwise use corresponding environment variable to determine it
+suse_airship_build_local_images: "{%- if airship_local_images_flag is defined -%}
```
evrardjp: FYI you can use the ternary filter to simplify this. However, do not use env vars for this, as it limits reproducibility. We should limit the usage of env vars to a minimum... And new feature -> new pr.

playbooks/group_vars/all.yml
```diff
@@ -0,0 +1,49 @@
...
+                                    {{ airship_local_images_flag | bool }}
+                                  {%- else -%}
+                                    {{ lookup('env','AIRSHIP_BUILD_LOCAL_IMAGES') | default('False', true) | bool }}
+                                  {%- endif -%}
+                                 "
+
+# Flag to be used when just want to use previously built local airship component images
+# without need of re-building them.
+suse_airship_use_local_images: "{{ lookup('env','AIRSHIP_USE_LOCAL_IMAGES') | default('False', true) | bool }}"
```
evrardjp: ditto.

playbooks/openstack-deploy_caasp.yml
```diff
@@ -178,8 +178,8 @@
         - name: Add vip with cidr in extravars
           lineinfile:
             path: "{{ socok8s_extravars }}"
-            regexp: "^suse_osh_deploy_vip_with_cidr.*"
-            line: "suse_osh_deploy_vip_with_cidr: {{ _vipip.stdout }}/24"
+            regexp: "^socok8s_deploy_vip_with_cidr.*"
+            line: "socok8s_deploy_vip_with_cidr: {{ _vipip.stdout }}/24"
```
evrardjp: please see my comment on naming the vip. I am fine with this, but if we have multiple, it might be better to give it the right name directly.

playbooks/roles/deploy-osh/defaults/main.yml
```diff
@@ -2,7 +2,7 @@
 # defaults file for deploy-osh
 
 # Location of the kubeconfig file, fetched from velum UI.
-kubeconfig_file_path: "~/suse-osh-deploy/kubeconfig"
+kubeconfig_file_path: "{{ socok8s_workspace }}/kubeconfig"
```
evrardjp: Can this use the common vars?

playbooks/roles/deploy-osh/tasks/kubectl-install.yml
```diff
@@ -1,9 +1,11 @@
 ---
 - name: Download latest stable kubectl
+  become: yes
   shell: >-
-    kubectl version || (curl -LO https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl &&
+    kubectl version --client || (curl -LO https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl &&
     chmod 0755 ./kubectl &&
-    mv ./kubectl /usr/local/bin/kubectl)
+    mv ./kubectl /usr/local/bin/kubectl &&
```
evrardjp: It seems unrelated to this patch, can you separate it into a new PR and add a reason for the extra link?

playbooks/roles/deploy-osh/tasks/kubectl-install.yml
evrardjp: ```when: x and y ``` should be rewritten to:

```
when:
  - "x"
  - "y"
```

playbooks/roles/deploy-osh/tasks/kubectl-install.yml
```diff
@@ -61,12 +63,20 @@
       stat:
         path: "{{ kubeconfig_file_path }}"
       delegate_to: localhost
+      register: new_config
+
+    - name: Find if an existing .kube/config file
```
evrardjp: Could you explain why and spin that up in a new commit?

playbooks/roles/dev-patcher/tasks/main.yml
```diff
@@ -29,6 +29,17 @@
 
 - name: Get patch details
   include_tasks: patch-repos.yml
-  loop: "{{ dev_patcher_patches }}"
+  loop: "{{ dev_patcher_patches | default([], True) }}"
```
evrardjp: dev_patcher_patches is always defined in defaults, no need to default it.

playbooks/roles/dev-patcher/tasks/main.yml
```diff
@@ -29,6 +29,17 @@
...
   loop_control:
     loop_var: patchnumber
+
+- name: Check if there is local code change to cloned repos on deployer
+  stat:
+    path: "{{ role_path }}/files/repos/"
+  register: local_change_repos
+
+- name: Copy local code changes to cloned repos on deployer
+  synchronize:
```
evrardjp: I am not sure why this exists? The feature currently works, no need for sync....

run.sh
```diff
@@ -2,7 +2,7 @@
 
 set -o errexit
 
-if [[ ${OSH_DEVELOPER_MODE:-"False"} == "True" ]]; then
+if [[ ${SOCOK8S_DEVELOPER_MODE:-"False"} == "True" ]]; then
```
evrardjp: good catch. You might also want to wrap the left hand side of the `==` with ```""```

script_library/run-ansible.sh
evrardjp: Agreed

vars/common-vars.yml
```diff
@@ -1,10 +1,19 @@
 ---
-upstream_repos_clone_folder: /opt
-developer_mode: "{{ (lookup('env','OSH_DEVELOPER_MODE') | default('False', true) ) | bool }}"
-socok8s_workspace: "{{ lookup('env','ANSIBLE_RUNNER_DIR') | default('~/suse-osh-deploy', true) }}"
+upstream_repos_clone_folder: "/opt"
+suse_osh_registry_location: "{{ lookup('env','OPENSTACK_HELM_IMAGE_REGISTRY_LOCATION') | default('docker.io', true) }}"
```
evrardjp: please add this in a different patch, this is adding new feature.

vars/common-vars.yml
```diff
@@ -1,10 +1,19 @@
...
+
+socok8s_workspace: "{{ lookup('env','ANSIBLE_RUNNER_DIR') | default('~/suse-socok8s-deploy', true) }}"
 socok8s_extravars: "{{ socok8s_workspace }}/env/extravars"
 socok8s_ses_pools_details: "{{ socok8s_workspace }}/ses_config.yml"
 socok8s_registry_cert: "{{ socok8s_workspace }}/certs/domain.crt"
 socok8s_registry_certkey: "{{ socok8s_workspace }}/certs/domain.key"
 socok8s_caasp_environment_details: "{{ socok8s_workspace }}/environment.json"
 socok8s_libvirtuuid: "{{ socok8s_workspace }}/libvirt.uuid"
+
+socok8s_deploy_config_location: "{{ socok8s_workspace }}"
```
evrardjp: ditto.

vars/manifest.yml
```diff
@@ -12,3 +12,18 @@ upstream_repos:
   loci:
     src: https://github.com/openstack/loci
     version: c94e1dfc86cae044a4582f321bb1721846419a00
+  airship-armada:
+    src: https://github.com/openstack/airship-armada
+    version: master
```
evrardjp: you should probably freeze this -- but that's your business.

doc/source/deploy-guide/configure-environment.rst
kantarun: We can certainly improve things but it might be better if we do those as post-merge actions. Any improvement (variable name change) in this patch means, that needs to be tested/verified in ucp branch side as well.

playbooks/roles/deploy-osh/tasks/kubectl-install.yml
kantarun: Will try to address in next patch.

script_library/run-ansible.sh
kantarun: I will suggest to do this as a post-merge action.
IMHO suse prefix makes it quite obvious to users (non-developers) when looking for things and then its kind of a product qualifier.
For me, it works either way.

examples/workdir/env/extravars
kantarun: My thinking was that this flag may help in making 'state' decisions what kind of deployment is targeted or current type of deployment is already orchestrated. 
If we are just doing deploy_osh or deploy_airship , then commands itself are obvious . I was thinking it may be useful when we are doing peacemeal actions (subset of 7 or 8 or 1-6 actions) e,g. creating subvolumes when adding compute (lifecycle steps) or creating inventory file for osh or airship deployment . 
IMHO, i am sure there might be better way or too premature at current state of things. Just added more as placeholder and can be removed easily if its not needed as post-merge cleanup.

playbooks/generic-build_images.yml
kantarun: make sense

playbooks/group_vars/all.yml
kantarun: For me, explicitly including common-vars wherever needed seems like additional step. It may not be obvious to someone who is new to code. Also common-vars are intended to be shared across all roles and in a way indirectly shared across all nodes.

I have commented out common_vars and may be we can remove the file later if group_vars seems to work for us.

playbooks/group_vars/all.yml
kantarun: I have to check the code for details but iirc those locations are pre-defined to use deployer node registry. This options seems to allow other public or internal registry if needed.

run.sh
kantarun: Wrap 'SOCOK8S_DEVELOPER_MODE' ? e.g. 
${"SOCOK8S_DEVELOPER_MODE":-"False"}. 

I see this can be a post-merge cleanup / improvement action if its okay.

vars/manifest.yml
kantarun: Agreed, we will have to do that. I think, current thinking was that as we are still in development mode so don't want to miss out fixes happening in upstream for problems we may see or already observed.

doc/source/deploy-guide/configure-environment.rst
evrardjp: lgtm then.

playbooks/roles/deploy-osh/tasks/kubectl-install.yml
evrardjp: Don't forget the wrapping with double quotes :)

script_library/run-ansible.sh
evrardjp: Fine for me to do it in another patch.

examples/workdir/env/extravars
evrardjp: Well I kinda agree and don't agree.

I agree on the fact that it's maybe too premature to call on the state of things. But then, if it's premature, why creating a new variable on some predictions of things that are premature to tell?

The addition of this var should be in a subsequent commit.

playbooks/group_vars/all.yml
evrardjp: I don't think it's a good idea.

These group_vars would apply to all the node, therefore overloading the ansible worker of data that are pretty much useless. Instead, loading the var at runtime would be easier to scale.

The idea of common-vars is indeed to load them from the roles. It is, according to me, simple to understand: When you read the role, it explicitly tells to load the vars.

With group vars + load vars, it becomes less obvious.

playbooks/group_vars/all.yml
evrardjp: The extra vars have precedence overall. If you say "I want to use my own registry", then you override that variable in extravars and it works.

run.sh
evrardjp: ``"${SOCOK8S_DEVELOPER_MODE:-False}" `` is what I meant. I am afraid of false positives.

evrardjp: But it can be done in a different commit if you prefer.

vars/manifest.yml
evrardjp: Fine for me :)

examples/workdir/env/extravars
kantarun: Okay..will remove it as currently its not used anywhere.

playbooks/roles/deploy-osh/tasks/kubectl-install.yml
kantarun: Okay..will add double quotes in next patch

run.sh
kantarun: Thanks. Will make this change in next patch

playbooks/group_vars/all.yml
kantarun: Online, when I looked for pattern around this problem, came across many recommendation of using group_vars for sharing data across roles based on group of hosts.

In general, ansible already provides this mechanism to define variables and to propagate them across group of nodes or all nodes. For simplicity and with current state of things (conflicting host group_names), just added to all. 
We can certainly add vars to group_vars/osh-workers if those are only intended for osh worker nodes during osh mode of deployment.

common-vars.yml is not a obvious place for looking variables for a newcomer as its not a standard practice and then require extra step of including it in specific roles before you can use it (came across issue myself as assumed common-vars is already included somewhere but it was not).

playbooks/group_vars/all.yml
kantarun: I don't see that variable currently defined in extravars.

And having that as environment variable sometimes can be easier as user don't have to touch the file.

I don't see obvious side-effect of including this way.

playbooks/group_vars/all.yml
kantarun: yes..so intention is to replace common-vars with **group_vars/{group_of_hosts}.yml** of sharing variables across group of hosts.
So whatever is defined under common-vars is current places in group_vars/all.yml for the reasons mentioned above.

playbooks/group_vars/all.yml
kantarun: same reply as above

playbooks/group_vars/all.yml
kantarun: This is what's there in ucp branch. And it should not conflict with master branch.

Also variables are moved from common-vars.yml to group_vars/all.yml .

playbooks/group_vars/all.yml
kantarun: Actually this can be a good place for defining base image.

Like for building images of airship components, we can change the base image from leap 15 to SLE 15 and dockerfile will use that to build the component image.
Having at a central place, can make this switch quite easy . May be we can build openstack service images that way as well.

playbooks/openstack-deploy_caasp.yml
kantarun: Currently I don't see that change in ucp branch so not sure we need to worry about that aspect. This can be enhanced in post-merge tasks.

playbooks/roles/deploy-osh/tasks/kubectl-install.yml
kantarun: This is we already have in ucp branch and have made life quite easier as now no need to refresh kube config every 2nd day in long running caasp cluster
So what's the concern other than part of a branch.

playbooks/roles/dev-patcher/tasks/main.yml
kantarun: I saw this error locally when we don't have any review added in list, then there is error reported. It just handles that case which can/will happen down the line.

playbooks/roles/dev-patcher/tasks/main.yml
kantarun: This is quite a life-saver when we are developing/ triaging issue locally. 

So I can make local changes in specific files of airship or osh project and copy those files and file structure to files/repos/ path. Then those changes are used and made available in my actual deployment. I heavily use it and 
can see benefit for others as well.

playbooks/roles/deploy-osh/tasks/kubectl-install.yml
kantarun: We have this change in ucp branch. I think change is to check explicitly for client version here as version command shows both server and client version. We are updating client aspect so it kind of makes sense though not sure if there is any other reason as change was made by somebody else.

playbooks/roles/deploy-osh/defaults/main.yml
kantarun: Its already coming from a global shared variable defined in group-vars.

playbooks/group_vars/all.yml
```diff
@@ -0,0 +1,48 @@
...
+
+developer_mode: "{{ (lookup('env','SOCOK8S_DEVELOPER_MODE') | default('False', true) ) | bool }}"
+
+socok8s_workspace: "{{ lookup('env','ANSIBLE_RUNNER_DIR') | default('~/suse-socok8s-deploy', true) }}"
+socok8s_extravars: "{{ socok8s_workspace }}/env/extravars"
+socok8s_ses_pools_details: "{{ socok8s_workspace }}/ses_config.yml"
+socok8s_registry_cert: "{{ socok8s_workspace }}/certs/domain.crt"
+socok8s_registry_certkey: "{{ socok8s_workspace }}/certs/domain.key"
+socok8s_caasp_environment_details: "{{ socok8s_workspace }}/environment.json"
+socok8s_libvirtuuid: "{{ socok8s_workspace }}/libvirt.uuid"
```
evrardjp: Even if I very much like this line, I think it should be in a different commit. The problem here is that we won't explain the `why` of this operation. It will be hard to understand 2 months from now.

playbooks/roles/deploy-osh/defaults/main.yml
evrardjp: I am fine with the reuse.

playbooks/roles/deploy-osh/tasks/kubectl-install.yml
evrardjp: I am sorry to be a pain but `this change was made by somebody else` is not a valid `why` reason for me. I can't review without a why. It's also very unrelated to the scope of this patch, so it should be a different patch. If you made this series of different patches with the why, they would have been merged already. It's slowing us down to have `mega patches`.

playbooks/roles/deploy-osh/tasks/kubectl-install.yml
evrardjp: I am confused, it seems exactly like https://github.com/SUSE-Cloud/socok8s/pull/115/files#diff-46a36719526ac2ed40c3a997043b868dR35 . 

Do you mean when you don't use automation, you still want to test the age of the kubeconfig, and have a way to copy the file?

In that case, please create a new PR, and update the logic I stated above.
It should updated like this to avoid repeating ourselves:
1) Check if kubeconfig is recent
2) If automation is present, update the kubeconfig with velum-interactions
3) If not present, pdate the kubeconfig by copying a local file (so about the same logic as now, just reusing the stat in 1) and overriding the file.

Still worth doing in another PR.

playbooks/roles/dev-patcher/tasks/main.yml
evrardjp: This is because you edited to remove something from the role. Assuming users don't edit the roles (because they never have to, nor they should), it should never be a problem.

It's defined by default, no need to redefine a default, that's what `defaults/main.yml` are for.

playbooks/roles/dev-patcher/tasks/main.yml
evrardjp: So you want a new feature in developer mode to override local changes without an upstream patch. I don't think this should be done in this commit. I could be fine in another PR.

vars/common-vars.yml
```diff
@@ -1,10 +1,21 @@
...
-developer_mode: "{{ (lookup('env','OSH_DEVELOPER_MODE') | default('False', true) ) | bool }}"
-socok8s_workspace: "{{ lookup('env','ANSIBLE_RUNNER_DIR') | default('~/suse-osh-deploy', true) }}"
-socok8s_extravars: "{{ socok8s_workspace }}/env/extravars"
-socok8s_ses_pools_details: "{{ socok8s_workspace }}/ses_config.yml"
-socok8s_registry_cert: "{{ socok8s_workspace }}/certs/domain.crt"
-socok8s_registry_certkey: "{{ socok8s_workspace }}/certs/domain.key"
-socok8s_caasp_environment_details: "{{ socok8s_workspace }}/environment.json"
-socok8s_libvirtuuid: "{{ socok8s_workspace }}/libvirt.uuid"
+# Proposing to phase out common_vars in favor of group_vars to define usage of global
+# variables and auto loading/availability of variables in roles
```
evrardjp: Please do this in another commit.

As said above, I don't agree with that in the same commit, and on top of it, I think it's a very bad idea.

playbooks/roles/deploy-osh/tasks/kubectl-install.yml
kantarun: I have provided the reason in my earlier response. I just mentioned that change was not made by me but reason provided earlier kinda makes sense to me.

Now to question of whether its related or not and require a separate PR ..that's a separate discussion (related to issue of having a branch with big change). When we merge that branch, it will bring many additions but expectation is that those should not impact existing CI or ECP mode of dev deployment. 

If you see this will impact existing code in master branch, then I am all ears about the reason and way to address it.

playbooks/roles/deploy-osh/tasks/kubectl-install.yml
kantarun: I am not sure I still got the understanding of the concern about the way its implemented. We are using k8s supported way of refreshing bearer tokens which is supported by  caasp generated kube-config.

Now what's suggested might be another approach, does not seems like better approach as there is additional extra step needed to use velum interaction. Whereas in current implemented approach, its works behind the scenes.

Before this, we used to see expired token error when playbook play executed after a day or two and that does not with this fix.

Now having a separate PR is a separate discussion and we have already talked about why its not done in this case and not worth repeating here.

https://kubernetes.io/docs/reference/access-authn-authz/authentication/#option-1-oidc-authenticator

playbooks/roles/dev-patcher/tasks/main.yml
kantarun: Why we would not have case where there are no values defined for https://github.com/SUSE-Cloud/socok8s/blob/master/playbooks/roles/dev-patcher/defaults/main.yml#L10 . This can happen when we don't have any more upstream reviews to consume.

I had this case initially when mentioned patch was merged, so I removed it from value list and now list needed to be empty, so removed it and saw this error when executed the play. I did not know how to define empty list in yaml for that variable so ended removing it. So here we are trying to handle that case.

playbooks/roles/dev-patcher/tasks/main.yml
kantarun: We already have this change in ucp branch and used. Having a separate PR for every UCP branch change is going to inefficient, error-prone and time-consuming. I don't see any downside of having this change slide-in or added benefit of having as separate PR as its already reviewed/used multiple times.

jgu17:
Can we please focus on the big picture that 1) we need to get this merge done as quick as we can, otherwise, more changes are coming as Airship team starts new Sprint, and 2) ensure the merge of Airship deployment code doesn't break/disrupt the OSH deployment that the Helm squad is currently using? We started to note non critical improvement/refactoring items on the etherpad so we file Jira ticket to track them post merge, but those don't necessarily have to be addressed in this PR.

examples/workdir/env/extravars
kantarun: Its already removed so resolving it again

examples/workdir/env/extravars
```diff
@@ -4,6 +4,17 @@ ceph_user_keyring_b64key: QVFDZWoySmNBQUFBQUJBQWlhSTZJZEw2Nml4S1FINEVlOTBTbXc9PQ
 # This is a list of the Ceph MONs, should the ses_config.yml to override the
 # value provided by the ses_config.yml (in case of dual stack, floating IPs...)
 suse_osh_deploy_ceph_mons: ['192.168.50.10:6789']
+
 # This is the VIP that will be used for the openstack deployment on kubernetes
 # (ingress controller)
-suse_osh_deploy_vip_with_cidr: 192.168.50.46/24
```
evrardjp: If this is removed, then please update the code to not break the deploy-osh, as it seems not updated.

playbooks/roles/deploy-osh/tasks/kubectl-install.yml
evrardjp: The problem is not to provide the reason or not in the review, but in the code/commit message.

But that comment didn't cover the link anyway.

Let's stop nitpicking on this, you have said you will do things after this merge, so I expect this would be part of it?

playbooks/roles/deploy-osh/tasks/kubectl-install.yml
evrardjp: I just meant to reuse a little of the code above. If you don't want, fine for me.

playbooks/roles/deploy-osh/tasks/main.yml
```diff
@@ -13,12 +13,12 @@
...
     - always
 
-- name: Pre-fligt checks for the role
+- name: Pre-flight checks for the role
   assert:
     that:
       - ceph_admin_keyring_b64key is defined
       - ceph_user_keyring_b64key is defined
-      - suse_osh_deploy_vip_with_cidr is defined
+      - socok8s_deploy_vip_with_cidr is defined
```
evrardjp: please update with change above.

playbooks/roles/dev-patcher/tasks/main.yml
evrardjp: If you have no values, you should do
``dev_patcher_patches: []`` , not comment things. Totally worth documenting though.

playbooks/roles/dev-patcher/tasks/main.yml
evrardjp: Ok for me, if the others are okay. I will not be pedantic here.

playbooks/roles/deploy-osh/tasks/kubectl-install.yml
```diff
@@ -1,9 +1,11 @@
 ---
 - name: Download latest stable kubectl
+  become: yes
```
Itxaka: This should not be needed as we are using the root user I believe

playbooks/roles/deploy-osh/templates/ingress-kube-system.yaml.j2
```diff
@@ -13,6 +13,6 @@ network:
     # possible options: routed, keepalived
     mode: routed
     interface: ingress-vip
-    addr: {{ suse_osh_deploy_vip_with_cidr }}
+    addr: {{ socok8s_ext_vip }}
```
Itxaka: is this correct? shouldnt it be `socok8s_deploy_vip_with_cidr` ?

playbooks/roles/dev-patcher/tasks/main.yml
Itxaka: kind of ok for me, can you add tags to the tasks so we can skip them if needed?

Would have loved to have this as a different Pr as this is a different logical change, shouldnt be paired here as its a new fature.

playbooks/roles/dev-patcher/tasks/main.yml
```diff
@@ -29,6 +29,17 @@
...
 - name: Get patch details
   include_tasks: patch-repos.yml
-  loop: "{{ dev_patcher_patches }}"
+  loop: "{{ dev_patcher_patches | default([], True) }}"
   loop_control:
     loop_var: patchnumber
+
+- name: Check if there is local code change to cloned repos on deployer
+  stat:
+    path: "{{ role_path }}/files/repos/"
```
Itxaka: is role_path/files a proper path to store this? We can possible have repos inside the roles???

Seems a bit awkward to me, this should point to an external, different dir which is out of the way. We already have a dir for things out of the roles scope/external and that is `"{{ playbook_dir }}/../files/`

Can you point this there instead as to be in sync?

vars/common-vars.yml
```diff
@@ -1,10 +1,45 @@
...
+
+######################################################
+# Following variables are used by airship logic only
+######################################################
+socok8s_deploy_config_location: "{{ socok8s_workspace }}"
+ucp_namespace_name: "{{ ucp_namespace | default('ucp') }}"
+
+# Variables used for building airship images locally
+suse_distro_identifier: opensuse
+suse_airship_components_base_image: "opensuse/leap:15.0"
```
Itxaka: This is kind of absurd. socok8s vars start with socok8s, shouldnt airship vars start with airship? why suse_airship? Do we plan on having airship_* vars coming from a different place?

Itxaka:
alos would love to see those commits actually fixed. We have 2 commit of ` Merge remote-tracking branch 'upstream/master' into ucp...` which will *not* get in socok8s because its noise and git history should not be noise + 1 commit that fixes a previous commit, which should be fixed up into the commit that is fixing.

Basically the commits from this PRs are terrible. Im not gonna ask for splitting them into proper commits as I know its a PITA but can we at least squash them together and have a full list of everything changed on the commit message so its clearer down the line what was introduced here please?

playbooks/roles/deploy-osh/templates/ingress-kube-system.yaml.j2
kantarun: This is some merge issue. I have addressed it locally but for some reason, update is not reflected here.
socok8s_ext_vip  is updated name for suse_osh_deploy_vip_with_cidr .

vars/common-vars.yml
kantarun: Updated file and should not impact osh-deployment.

vars/common-vars.yml
kantarun: There is no standard naming convention identified and adopted so far. 

May be we can revisit variables used in this file and standardize as a post ucp branch merge.

playbooks/roles/dev-patcher/tasks/main.yml
kantarun: This is dev-only functionality and we are NOT likely to check-in temporary code changes to branch. Hopefully, we will have a upstream review after successful local testing out of that.

The reason this is in dev-patcher as its only needed during dev-work and does not have to visible externally. 
 

This is kind of file structure I have to use local-only changes

arun:~/myFolder/myWork/suse_cloud/arun_forks/socok8s/playbooks/roles/dev-patcher/files$ tree repos
repos
├── airship-deckhand
│   ├── images
│   │   └── deckhand
│   │       ├── Dockerfile.opensuse
│   │       └── Dockerfile.ubuntu
│   └── Makefile
├── airship-shipyard
│   ├── charts
│   │   └── shipyard
│   │       └── templates
│   ├── images
│   │   └── airflow
│   └── src
│       └── bin
│           └── shipyard_airflow
│               └── entrypoint.sh
└── openstack-helm
    └── keystone
        └── templates
            └── bin

playbooks/roles/dev-patcher/tasks/main.yml
kantarun: There is check if file/repos/ is not there, then nothing is executed .

Not sure about which tag you are referring to use here. 

Also making changes here trigger change in ucp branch as it needs to be re-merged after this PR is merged. Trying to minimize merge-conflict in a way

PR #117, opened 2019-03-01
Added some latest patches for loci and openstack-helm

jsuchome:
Removed patch 621206 that has been already merged


playbooks/roles/dev-patcher/defaults/main.yml
```diff
@@ -8,4 +8,6 @@ dev_patcher_packages:
 #unrelated patches will fail (because not cloned)
 #conflicting patches will fail (because need resolving)
 dev_patcher_patches:
-  - 621206
+  - 637963
```
Itxaka: can we have a comment above with the patch description so we dont have to go look them up?

jsuchome: Done, I hope ansible handles this

PR #118, opened 2019-03-01
Fix the variable name containing loci build command.

jsuchome:
Also, stay on ocata for now.


jsuchome:
> Also, stay on ocata for now.

Well, that might not be good idea, building for ocata seems to fail, I'll remove this part of the commit

evrardjp:
yeah but master fails too :p 
Should we move to queens?

jsuchome:
Current script does point to queens

evrardjp:
We should probably build the wrapper for rocky in upstream too, and move to it. But it's a little premature, as Queens is not working yet in upstream...

PR #122, opened 2019-03-06
Removed brackets from suse_osh_deploy_ses_mons value and .join from storageclass tempate

matthew-fuller:
Since the storageclass template is expecting a string and not an array for the monitors key, I've removed the brackets from suse_osh_deploy_ses_mons. Also, since the port number and comma is being appended to each potential IP in the list during the "Create a list of monitors..." task in main.yml, I've removed the .join from the k8s-ses-storageclasses.yml.j2 template.


rajdharwadkar:
Looks good.

matthew-fuller:
Unable to duplicate previous errors that prompted this PR, so I'm closing it as unnecessary

PR #126, opened 2019-03-07
Removing uwsgi param and changing postgres pod user back to root

kantarun:
uwsgi_read_timeout parameter is not recognized by nginx server running
in ucp ingress pod. So removing this config param setting which was
added earlier to resolve shipyard cli 504 timeout issue.

Updating postgres pod to run as root user instead of upstream change
of using 999 (postgres user) as we are seeing permission denied
when 999 user tries to change permission of directory
/var/lib/postgreslq/data directory.

Updated openSUSE image tag with version.

### DIFF
--- a/playbooks/roles/airship-deploy-ucp/templates/airship.yaml.j2
+++ b/playbooks/roles/airship-deploy-ucp/templates/airship.yaml.j2
@@ -199,6 +199,11 @@ data:
     pod:
       replicas:
         server: 1
+      # TODO(arunkant) Temporary workaround for upstream change and failure seen on caasp node
+      # https://review.openstack.org/#/c/635070/  (More investigation needed)
+      security:
+        server:
+          runAsUser: 0
   source:
     type: local
     location: /armada/airship-components/openstack-helm-infra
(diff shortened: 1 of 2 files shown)


playbooks/roles/airship-deploy-ucp/templates/airship.yaml.j2
```diff
@@ -78,6 +78,31 @@ data:
...
+  namespace: ucp
+  wait:
+    timeout: 1800
+    labels:
+      release_group: airship-ucp-postgresql
+  values:
+    pod:
+      security:
+        server:
+          runAsUser: 0
```
jgu17: Is root user the permanent solution? If not, can we add a todo comments?

kantarun: At the moment, we don't know. Ideally we need to figure out why its issue for us with caasp node.
May be we need to investigate more later.

Is there standard convention for TODO here to follow.

jgu17: I am not aware. I followed the existing #TODO comments in the socok8s.

PR #129, opened 2019-03-11
Added temporary workaround to proceed with shipyard gateway 504 error.

rajdharwadkar:


playbooks/roles/airship-deploy-osh/tasks/main.yml
```diff
@@ -115,7 +115,8 @@
   environment:
     SHIPYARD_IMAGE: "{{ shipyard_image }}"
   register: shipyard_create_config
-  failed_when: shipyard_create_config.stdout.find('Error') == 0
+#  TODO(Raj) This is temporary workaround to proceed with shipyard error.
+#  failed_when: shipyard_create_config.stdout.find('Error') == 0
```
jgu17: Did we need a wait/sleep in this workaround so the create doc action finishes before the commit action happens?

kantarun: How we are handling timeout error which will happen when config docs insert/update happens via shell execution.

We need to cover 2 aspects
1) Wait for total of 3 minutes
2) And ignore timeout error

jgu17:
no longer needed

PR #130, opened 2019-03-11
Generate libvirt UUID as a separate task, and replace all static UUIDs

matthew-fuller:
This change replaces all static UUIDs in the OSH value override templates with a dynamically generated UUID that is created as a separate task in the airship-deploy-osh role.

### DIFF
--- a/playbooks/roles/airship-deploy-osh/tasks/main.yml
+++ b/playbooks/roles/airship-deploy-osh/tasks/main.yml
@@ -17,6 +17,18 @@
     - always
     - run
 
+# This is an arbitrary UUID value that's passed to the libvirt daemonset setup job for configuring a libvirt secret XML file
+- name: Create Libvirt ceph cinder secret uuid
+  shell: "uuidgen > {{ socok8s_libvirtuuid }}"
+  args:
+    creates: "{{ socok8s_libvirtuuid }}"
+  delegate_to: localhost
+
+- name: Get libvirt secret
+  set_fact:
+    libvirt_ceph_cinder_secret_uuid: "{{ lookup('file', socok8s_libvirtuuid) }}"
+  run_once: True
+
 - name: Set site path
   set_fact:
     site_path: "{{ upstream_repos_clone_folder }}/airship-treasuremap/site/{{ socok8s_site_name }}"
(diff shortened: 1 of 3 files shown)


playbooks/roles/airship-deploy-osh/tasks/main.yml
```diff
@@ -17,6 +17,18 @@
     - always
     - run
 
+# This is an arbitrary UUID value that's passed to the libvirt daemonset setup job for configuring a libvirt secret XML file
+- name: Create Libvirt ceph cinder secret uuid
+  shell: "uuidgen > {{ socok8s_libvirtuuid }}"
+  args:
+    creates: "{{ socok8s_libvirtuuid }}"
+  delegate_to: localhost
```
kantarun: Do we want to generate UUID with each run of that play in same environment? Is it okay to have different generated UUID values and used if play is re-run ?

If its okay to have different UUID with each run, then this logic is fine. We just need to check if its okay and also will it trigger site update anyway if that's the only change ?

jgu17: I don't think it makes sense to regenerate it every time.  yes, when the uuid changes, the site is considered changed, pretty expensive.

matthew-fuller: "creates" will skip the action if the referenced file already exists, so as long as {{ socok8s_workspace }}/libvirt.uuid hasn't been removed since the last run, the UUID will never be regenerated. Each subsequent run will refer back to that file and reuse the same UUID.

playbooks/roles/airship-deploy-osh/templates/site/soc/software/charts/osh/openstack-compute-kit/libvirt.yaml
```diff
@@ -28,5 +28,5 @@ data:
         cinder:
           user: {{ ses_cluster_configuration.cinder.rbd_store_user }}
           keyring: {{ ses_cluster_configuration.cinder.key }}
-          secret_uuid: 76f12f7a-16a6-11e9-864e-52540033f343
+          secret_uuid: {{ libvirt_ceph_cinder_secret_uuid }}
```
toabctl: Why not something like `{{ 99999999 | random | to_uuid }}` instead of calling shell that creates a file. Or is that uuid persistent and shouldn't change while running multiple times?

jgu17: I prefer it not to change when the playbook reruns. Otherwise it causes the site manifest content to change which triggers the Airship to create a new version of the site config.

matthew-fuller: Exactly--if a new UUID is generated and applied with each run of the playbook, even a totally unrelated change would trigger an unnecessary (and as James points out, expensive) redeployment of cinder, nova, and libvirt since the new UUID would be seen as a new configuration change for those charts.

PR #131, opened 2019-03-12
Adding latest helm client install when min version is specified

kantarun:
Upstream airship is updated to latest client version and existing
logic does not update when client and server version matches.

Now forcing latest client install when both client and server version
are matching and min version is secified (non empty value)

Also to apply update, helm serve needs to be stopped in case of
helm client update. Gives error as it says install location file is
busy.


jgu17:
Don't you only want to update if the current version is older than specified minimum version?  And Should we pin down the helm version, instead of getting the latest? Helm has beening doing good job of keeping backwards compatibility with the 2.x so far, but it may not be so friendly when it goes to 3.0.  the helm get script the playbook is using supports -v option for you to pass in the desired version.

kantarun:
This is no longer applicable as helm is installed as rpm package.

PR #132, opened 2019-03-12
Run dist-upgrade on CaaSP nodes

jhesketh:
To ensure we have the latest updates for CaaSP, add the repos and run
dist-upgrade as a transactional-update (after which a restart is required
to apply).


jsuchome:
Have you considered registering the nodes in SCC and getting the repos this way?

jhesketh:
> Have you considered registering the nodes in SCC and getting the repos this way?

No I haven't. I'm not sure what would be required here. We don't seem to do this for SES or SLE, but perhaps we should? I'm also not sure how complex that might make things when this is really just for developer environments at this point.

jsuchome:
I've described it here https://jira.suse.de/browse/SCRD-8178 so I think it's not that difficult.
However that was at the time I was not able to find the repos ... so if you add the repos directly, it's OK I guess.

jhesketh:
> I've described it here https://jira.suse.de/browse/SCRD-8178 so I think it's not that difficult.
> However that was at the time I was not able to find the repos ... so if you add the repos directly, it's OK I guess.

To be honest, I'm not sure if these are the most correct repositories. I believe they are from my understanding of IBS but I could be mistaken.

I'm not opposed to using the SCC but it does look more convoluted. Maybe that's a future improvement to look at across the various products we're consuming?

PR #135, opened 2019-03-13
An extra writable subvolume is needed for libvirt pod

jsuchome:
Lousily copied from commit 7d078e0b8252154c6fc766bcc6b5cfca630d4bdd

Ideally we would drop this `caasp_cp_node_set_subvolumes.sh` and use `roles/airship-configure-worker/tasks/subvolumes-setup.yml`


Itxaka:
I believe we cannot use the airship piece as our inventories are different. Hopefully we can bring this differences on the monday sync meeting and maybe agree on a shared inventory? Then we could move that ahead to DRY

PR #136, opened 2019-03-14
Add overrides for neutron

Itxaka:
When deploying the compute kit we cannot override the images used
for neutron currently as we are not passing any values to the
compute-kit script

This adds a task that will deploy the socok8s-neutron.yaml config
into the deployer and uses as part of the environment variables
when executing the nova compute-kit deploy script

### DIFF
--- a/playbooks/roles/deploy-osh/tasks/main.yml
+++ b/playbooks/roles/deploy-osh/tasks/main.yml
@@ -341,6 +341,17 @@
     - libvirt
     - run
 
+# we need to deploy the extra values for neutron so on the next task they are picked up
+- name: Deploy neutron overrides
+  import_tasks: component-install.yml
+  vars:
+    _cmpnt_template: neutron.yaml
+    _cmpnt_overrides: "{{ suse_osh_deploy_neutron_yaml_overrides }}"
+    _cmpnt_runcommand: "/bin/true" # we only want to create the template, not run anything
+  tags:
+    - neutron
+    - run
+
 # TODO(evrardjp): Check if need to have our own dummy overrides
 # In that case, try to split the shell script upstream in two scripts.
 # Compute kit deploys nova and neutron but only nova needs overrides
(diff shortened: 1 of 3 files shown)


playbooks/roles/deploy-osh/tasks/main.yml
```diff
@@ -341,6 +341,17 @@
     - libvirt
     - run
 
+# we need to deploy the extra values for neutron so on the next task they are picked up
+- name: Deploy neutron overrides
+  import_tasks: component-install.yml
+  vars:
+    _cmpnt_template: neutron.yaml
+    _cmpnt_overrides: "{{ suse_osh_deploy_neutron_yaml_overrides }}"
+    _cmpnt_runcommand: "/bin/true" # we only want to create the template, not run anything
```
jsuchome: This is smart, but maybe we can add some variable that would mean "Run the override file" is skipped?

Maybe if `_cmpnt_runcommand` is empty , some "when" clause would not run it?

I know the result is the same, but it might look better

Itxaka: Yes, I didnt want to go further or anything with this as its kind of a workaround currently pending any further work upstream to divide that compute-kit?

So I wanted to make it work *now* fix it later if we need to go upstream to change it

jsuchome: Well... it might have some other usage in theory, even if this particular script is splited upstream

playbooks/roles/deploy-osh/tasks/main.yml
```diff
@@ -352,6 +363,7 @@
     _cmpnt_runcommand: "./tools/deployment/multinode/140-compute-kit.sh"
   environment:
     OSH_EXTRA_HELM_ARGS_NOVA: "{{ lookup('env', 'OSH_EXTRA_HELM_ARGS_NOVA') | default('', True) }} --values /tmp/socok8s-nova.yaml"
+    OSH_EXTRA_HELM_ARGS_NEUTRON: "{{ lookup('env', 'OSH_EXTRA_HELM_ARGS_NOVA') | default('', True) }} --values /tmp/socok8s-neutron.yaml"
```
jhesketh: I think you might mean 'OSH_EXTRA_HELM_ARGS_NEUTRON'

Itxaka: :+1:

playbooks/roles/deploy-osh/tasks/main.yml
evrardjp: You can simply call config_template, like L175.

evrardjp:
No need for something a little bit obscure :)

evrardjp:
The improvement to use config_template directly can be done later. Also it's simple.

PR #142, opened 2019-03-18
Upgrade CaaSP nodes after velum registration

jsuchome:


jsuchome:
Yeah, velum seems to work well. I'm just not yet sure if it will end with stable cluster

evrardjp:
Do we still need this?

jsuchome:
Nope

PR #144, opened 2019-03-18
Allow adding a list of patches for dev-patcher in extravars

Itxaka:
currently we either use the default list of patches that comes
with socok8s or we override the list on extravars and have to add
all of those patches are they are not merged together.

This introduces a var dev_patcher_user_patches that allows you to specify
any upstream patches to add without overriding the default list so
you no longer need to keep the extravars in sync with what we have in the
playbook

### DIFF
--- a/doc/source/user/use-custom-patches.rst
+++ b/doc/source/user/use-custom-patches.rst
@@ -1,0 +1,25 @@
+===================
+Use custom patches
+===================
+
+
+If you want to apply upstream patches in your environment, set your patch numbers
+under the `dev_patcher_user_patches` key on `${WORKDIR}/env/extravars`:
+
+.. code-block:: yaml
+
+  dev_patcher_user_patches:
+    # test patch for kyestone
+    - 12345
+    # test patch for cinder
+    - 12345
+
+
+This way, the patches will only be carried in your environment.
+If you want to change the product instead (for developer mode or not),
+please propose a PR to the socok8s repo.
+
+.. note::
+
+    This list of patches provided via extravars will be appended to the default
+    patches list available on the dev-patcher role vars.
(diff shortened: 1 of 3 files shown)


playbooks/roles/dev-patcher/tasks/main.yml
```diff
@@ -7,6 +7,10 @@
   tags:
     - always
 
+- name: Include extra patches from user vars
```
evrardjp: No need for set fact:

playbooks/roles/dev-patcher/tasks/main.yml
```diff
@@ -29,7 +33,7 @@
 
 - name: Get patch details
   include_tasks: patch-repos.yml
-  loop: "{{ dev_patcher_patches | default([], True) }}"
+  loop: "{{ dev_patcher_user_patches + dev_patcher_patches | default([], True) }}"
```
evrardjp: Can just replace ```loop: "{{ dev_patcher_patches | default([], True) }}"``` with:

```
loop: "{{ (dev_patcher_patches) + (dev_patcher_user_patches | default([])) }}"
```

dev_patcher_patches: should always be defined in defaults/ so no need to default it. And The uter can just be defaulting to a list.

Itxaka:
I think we may need some docs or at least an entry in the example extravars or this is gonna be forgotten

evrardjp:
Please add a user story in the docs?

doc/source/user/use-custom-patches.rst
```diff
@@ -0,0 +1,23 @@
+===================
+Use custom patches
+===================
+
+
+Sometimes there may be upstream patches that you want to apply with the dev-patcher
+but creating a PR and send it to the soco8ks repo may delay your work, or you are
```
evrardjp: That's not the reason. Please remove this line.

You could rephrase to simply:

```If you want to apply upstream patches in your environment, set the following: ...```

```This way, the patches will only be carried in your environment. If you want to change the product instead (for developer mode or not), please propose a PR to the socok8s repo.```

PR #145, opened 2019-03-18
Fix issues when running OSH deploy steps in parallel

flaviodsr:
Running `deploy_ses`, `deploy_caasp` and `deploy_ccp_deployer` in parallel (using a jenkins job) raised the following issues:
  - deploy-osh: In some cases the cluster is not ready to reply back the cluster-info
query.
  - delete-caasp-stack: Although the `Get a floating ip for the workers` task is skipped it
requires to template the loop parameter which fails as one of the varaibles
is defined by a task that is skipped and does not have `stdout`.
  - deploy-caasp: Running deploy CaaSP in parallel with other steps fails as the extravars
file is not created yet

This PR contains fixes for those issues.

### DIFF
--- a/playbooks/openstack-caasp_instance.yml
+++ b/playbooks/openstack-caasp_instance.yml
@@ -60,7 +60,7 @@
         # Do not reuse to prevent ssh issues or race conditions when running
         # multiple of those jobs at the same time.
         - name: Get a floating ip for the workers
-          loop: "{{ _workernodeswithnofloating.stdout | from_json }}"
+          loop: "{{ _workernodeswithnofloating.stdout | default('[]') | from_json }}"
           os_floating_ip:
             cloud: "{{ deploy_on_openstack_cloudname }}"
             server: "{{ item.ID }}"
--- a/playbooks/roles/deploy-osh/tasks/kubectl-install.yml
+++ b/playbooks/roles/deploy-osh/tasks/kubectl-install.yml
@@ -84,6 +84,9 @@
   command: kubectl cluster-info
   changed_when: false
   register: _kubectlclusterinfo
+  retries: 10
+  until: _kubectlclusterinfo is succeeded
+  delay: 5
 
 - name: Show kubectl cluster-info
   debug:


playbooks/openstack-caasp_instance.yml
```diff
@@ -180,6 +180,7 @@
             path: "{{ socok8s_extravars }}"
             regexp: "^socok8s_ext_vip.*"
             line: "socok8s_ext_vip: {{ _vipip.stdout }}"
+            create: yes
```
evrardjp: While I agree this would be fixing the parallel work, running this in parallel could technically enter a race condition where this task run at the same time as the templating of ses osh vars, which could then cause of or the other to get non-written. Same applies to inventory patching above.

These plays are not meant for running in parallel. I don't think it's a wise idea to do this in parallel, I am not sure how much we gain by doing so....

I am not strongly against, but for other reviewers, keep in mind adding this parameter basically opens us to race conditions.

evrardjp: If we make those changes on different files, I am in.

evrardjp: I think we should have an assert at the end of the play that checks the content of the extra vars, to ensure the vip was well set (by doing a lookup('file',...)) . That could work enough for me, and ensure the file has the vip.

flaviodsr:
Rebased to master, but to be able to deploy OSH running steps in parallel when possible we still need to address the issue with the task [1] as the `extravars` file is only created at the end of ses-deploy.

playbooks/roles/deploy-osh/tasks/kubectl-install.yml
```diff
@@ -84,6 +84,9 @@
   command: kubectl cluster-info
   changed_when: false
   register: _kubectlclusterinfo
+  retries: 10
+  until: _kubectlclusterinfo is succeeded
+  delay: 5
```
jhesketh: Rather than waiting for another process to bring up the cluster we should move everything that is dependent on the cluster (such as this) to another stage in the jenkins pipeline that can depend on the previous task being complete.

evrardjp: It technically is...  The deploy osh only runs after the caasp enrollment.

We still need to keep that code there to prevent the deployment if cluster-info is not in a good state (when we'll pass that code to customers).

However, I believe we should also have a success condition of the deployment of the caasp cluster, and caasp enrollment. It can be set to this for example,or the ability to run other kubectl commands. But that's outside the scope of this PR imo.

evrardjp: That said, I agree that we shouldn't hide issues here, we should assume a working state, and just test the state. If it fails, it means the cluster don't have the requirements for deploying OSH.

evrardjp:
If we bring the file create:yes I would be fine with bringing this in, even if it introduces a possible race condition. The value it brings exceeds its costs.

toabctl:
Needs a rebase. sorry

dirkmueller:
please reopen when rebased

PR #146, opened 2019-03-18
Added cinder-volume service account to PrivilegedRoleBinding template to allow required privilege escalation

matthew-fuller:
This change adds the service account 'cinder-volume' to the 'PrivilegedRoleBinding' template. The cinder-volume deployment now requires privilege escalation due to an upstream change to the cinder deployment-volume.yaml template. (https://github.com/openstack/openstack-helm/commit/f5bf6ec2deb0bf3a3803a4cb441c197fe19abe3b)


matthew-fuller:
This change was already added to the deploy-osh role back on Feb 6th with commit a653e35, so it should only be necessary to add it to the airship-configure-caasp role.

PR #147, opened 2019-03-19
Refactor Airship ucp site deployment

jgu17:
Moved the ucp site manifest from a single, standalone Armada site manifest yaml to be a true Airship site that shares the global treasuremap definitions, version control, endpoints, secrets etc. This change brings consistency in how we deploy ucp and osh, and similar experience to the end user for configuration mgmt of the two layers. Following work is needed to enable scale profile for the ucp control plane and tls endpoints.


evrardjp:
Very big patch, @jsuchome or @Itxaka could you have a look at the update of common-vars?

vars/common-vars.yml
```diff
@@ -47,3 +48,11 @@ suse_airship_components_image_tag: "{%- if suse_airship_build_local_images or su
                                     {%- else -%}
                                       latest
                                     {%- endif -%}"
+suse_osh_image_version: "pike"
```
jsuchome: Does it make sense to set this differently than the value in playbooks/roles/deploy-osh/defaults/main.yml ?

Itxaka: you are duplicating this value. Shouldn't we all use the same value for osh+airship? If not, cant we name them differently? because on the osh-deploy role in the vars there is the same var which one will be used when we deploy X or Y? This is confusing and could lead to problems so please use a different var name or lets agree in using the same version for both airship and osh

Itxaka:
Can we please please please *please* not have those terrible commits "Merge branch.." in there?

Also just as advice, but it feels to me than having 1 real commit with 1k changes (literally!) is pretty wrong. First this groups a lot of commits that might be different logic or touch different parts which have no business being together. Then there is the total breakage of things like git revert or git bisect, as its a huge blob so its all or nothing.

Currently if this PR breaks something and we need to revert it asap because its blocking us we would need to do the full revert of the 1000 lines, instead of reverting just 1 small commit. It just doesn't make sense to me.

jgu17:
> Very big patch, @jsuchome or @Itxaka could you have a look at the update of common-vars?
> 
> why would you do this to me 😿 a 1000 lines PR to check argh
> On it...

I am sorry :-( ...this is pretty much the minimum change to create a working "site" though... majority of line of code change is from site file location move so we share the common site artifacts and have a single site manifest structure for both osh and ucp.

jgu17:
> Can we please please please please not have those terrible commits "Merge branch.." in there?
> Also just as advice, but it feels to me than having 1 real commit with 1k changes (literally!) is pretty wrong. First this groups a lot of commits that might be different logic or touch different parts which have no business being together. Then there is the total breakage of things like git revert or git bisect, as its a huge blob so its all or nothing.
> Currently if this PR breaks something and we need to revert it asap because its blocking us we would need to do the full revert of the 1000 lines, instead of reverting just 1 small commit. It just doesn't make sense to me.

see my expalanation to a previous comment.

vars/common-vars.yml
jgu17: Good question. I see that playbooks/roles/deploy-osh/defaults/main.yml uses ocata. Is Ocata osh only deployment actually deploys? on Ubuntu or SUSE? 

Airship deployment has been using Pike for a quite a while and works pretty well for us. If the osh squad only uses Ocata for Ubuntu, can I suggest we align on Pike? This is also what demo at SUSECON, so I really don't think we want to use Ocata.

I cam imagine OSH squad could work on a newer version of openstack before it is ready for the Airship squad to adopt. Actually that's the case now, since you are on Rocky. So maybe we can add an override in OSH to the common var value? I am not against either to the idea Itxaka mentioned below to use a different variable. Whatever works.

Itxaka: My main problem with this is that there is the same var name in 2 places which can be set to different values, leading to problems when running this, so either we sync on the same branch (rocky for OSH currently) or we set different names as to not confuse ourselves :+1:

jsuchome: I would thin it makes sense to use same release, but if it is not  possible for some time, let's have different variable names...

jgu17: Airship+OSH deploy is using Pike on Ubuntu for development an susecon demo. I think OSH only deployment is working on Rocky on Leap.  I see the default value in deploy-osh is set to Ocata. Do you only use Ocata on Ubuntu? I mentioned in the comments to Jiri, Pike on Ubuntu has worked without issue for us. Can we move to Pike on Ubuntu on the osh deployment as well?  Do you want me to change that?

If we can't or have reasons not to align, I am not against to use different variable names either.

Itxaka: yeah I think we should have them separated for the moment but one next to the other in `vars/common-vars.yml` so we can merge them in the future and we dont forget about them

jgu17: I am fine with different version number for now. let me push a new commit. I'll keep "suse_osh_image_version" for deploy_osh role, and have a new var name "suse_airship-openstack_helm_image_version" for the airship deployment.

jgu17: actually, I am going to back out the change that adds the osh image version in the common vars. This can/should be in a separate commit as it is not required for the UCP site refactoring. And we will talk about convergence in theinter-squad meeting.

Itxaka: sounds good to me

PR #152, opened 2019-03-19
Update neutron setup for Rocky

jsuchome:
See https://docs.openstack.org/releasenotes/neutron/queens.html


jsuchome:
> Shouldn't this be done upstream?

Upstream charts are not expecting Rocky, right? So I think yes, it should be upstream but only once there's a way to have release-dependent code

playbooks/roles/deploy-osh/defaults/main.yml
```diff
@@ -96,4 +96,10 @@ suse_osh_deploy_memcached_yaml_overrides: {}
 suse_osh_deploy_nova_yaml_overrides: {}
 suse_osh_deploy_ovs_yaml_overrides: {}
 suse_osh_deploy_rabbitmq_yaml_overrides: {}
-suse_osh_deploy_neutron_yaml_overrides: {}
\ No newline at end of file
+suse_osh_deploy_neutron_yaml_overrides: {
```
Itxaka: is this ok or should the `{}` be removed instead?

I have it like this on my extravars:
```
suse_osh_deploy_neutron_yaml_overrides: 
  # api-paste entrypoint neutron.api.versions:Versions.factory was deprecated in Queens
  conf:
    paste:
      app:neutronversions:
        paste.app_factory: neutron.pecan_wsgi.app:versions_factory
```

jsuchome: Hm, I'm not sure. I also tested with extravars only, and added this PR so that whoever is testing neutron can use it...

jsuchome: Meaning = I did not test this PR myself

Itxaka: I did test it with those values, in any case I think they are both valid yaml so it should work?

Itxaka: actually let me retract that, it does not seem to be valid yaml :o

jsuchome:
@Itxaka better?

playbooks/roles/deploy-osh/defaults/main.yml
```diff
@@ -96,4 +96,11 @@ suse_osh_deploy_memcached_yaml_overrides: {}
...
 suse_osh_deploy_rabbitmq_yaml_overrides: {}
-suse_osh_deploy_neutron_yaml_overrides: {}
\ No newline at end of file
+
+suse_osh_deploy_neutron_yaml_overrides:
+  # api-paste entrypoint neutron.api.versions:Versions.factory was deprecated in Queens
+  # We must replace it for Rocky deployments
+  conf:
+    paste:
+      app:neutronversions:
```
evrardjp: shouldn't this need to be wrapped in double quotes?

jsuchome: I don't know ... it's the same value as in `values.yaml`

evrardjp: Semi Complex keys in yaml should be wrapped with "". And yes, "" is removed when parsing anyway, but it's for clarity of the key:value, here the key contains ":"

But ok, if it's like that in values, let's keep it.

Itxaka: works for me like that, tested. I was also surprised :o

jsuchome:
I would expect it is correct - the other PR was just setting `conf:` to empty value, this one is replacing some values inside.
That said, I was not able to test it myself, yet

Itxaka:
This should work, I have it like that on my environment and not having any issues deploying neutron :D

playbooks/roles/deploy-osh/defaults/main.yml
evrardjp: Can we make this part of the template instead? I don't see a reason we should not provide this by default in the template nowadays.

jsuchome: True

playbooks/roles/deploy-osh/templates/neutron.yaml.j2
```diff
@@ -1 +1,23 @@
...
+    bootstrap: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+    db_init: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+    db_drop: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+    ks_user: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+    ks_service: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+    ks_endpoints: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+    neutron_db_sync: "{{ suse_osh_registry_location }}/openstackhelm/neutron:{{ suse_openstack_image_version }}"
+    neutron_dhcp: "{{ suse_osh_registry_location }}/openstackhelm/neutron:{{ suse_openstack_image_version }}"
+    neutron_l3: "{{ suse_osh_registry_location }}/openstackhelm/neutron:{{ suse_openstack_image_version }}"
+    neutron_linuxbridge_agent: "{{ suse_osh_registry_location }}/openstackhelm/neutron:{{ suse_openstack_image_version }}"
```
evrardjp: Do you think we should add:
```

    neutron_sriov_agent: docker.io/openstackhelm/neutron:ocata-sriov-1804
    neutron_sriov_agent_init: docker.io/openstackhelm/neutron:ocata-sriov-1804
```

to remember we still need to do those for the technical preview?

jsuchome: Well, if it is not replaced it is still the same value.

If we have to remember those, I believe they should be tracked in Jira, not in the code.

evrardjp: They should be both tracked in Jira and code. Jira is internal, we don't know who wants to help in here.

playbooks/roles/deploy-osh/templates/neutron.yaml.j2
```diff
@@ -1 +1,23 @@
-{{ lookup('template','files/common-images-overrides.yml') | from_yaml | to_nice_yaml(indent=2) }}
\ No newline at end of file
+{% if developer_mode %}
```
evrardjp: We should list here all the upstream patches necessary, for helping the other squad, and tracking.

jsuchome: There's no current upstream patch for neutron.

Besides, I do not think we should communicate via comments in the code. Also we should not use code for tracking bugs and TODOs

PR #158, opened 2019-03-20
Change the armada chart deployment order

jgu17:
Moved armada chart deployment to the last in UCP site since we are using
armada-api pod status as the check if UCP is fully up.


evrardjp:
I am curious, isn't deckhand deployable with helm charts?

That looks good to me, but if deckhand needs armada, that's kinda weird :)

PR #160, opened 2019-03-20
Add a override workaround to disable memcached on nova

Itxaka:
Seems like the oslo_cache lib has some problems in python3

Lets override and disable the cache until the upstream fix[0] is
accepted

[0] https://review.openstack.org/#/c/640500/


Itxaka:
instead of disabling lets move to `dogpile.cache.memcached` which seems to work.

Closing this and sending a different PR

PR #162, opened 2019-03-20
Fixes a bug in ansible template to add overrides to nova

Itxaka:
due to a bug in ansible config template[0] we cannot currently pass
multiline overrides onto a merged template as they will be mangled so
we need to have a different file to pass to helm

This depends on https://review.openstack.org/#/c/642067/ to work. If the
pathc is not applied it wont break anything but also dont expect the placement
api to work :D

[0] https://bugs.launchpad.net/openstack-ansible/+bug/1819974


jsuchome:
As this extra overrides file is temporary anyway, don't you want to rename it to nova-placement-overrides so it is clear from name what is its purpose??

Itxaka:
I dont plan on merging this tbh, its just there as an example to avoid the current block

Itxaka:
Closing this as templates improvements have been merged already

PR #163, opened 2019-03-20
Register and upgrade CaaSP after the deployment

jsuchome:
Based on the official guidelines from the CaaSP team. 

Alternative to https://github.com/SUSE-Cloud/socok8s/pull/164


playbooks/openstack-prepare_caasp_workers_for_update.yml
```diff
@@ -0,0 +1,36 @@
...
+    - name: Pre-flight checks for the role
+      assert:
+        that:
+          - caasp_registration_key is defined
+
+    - name: Register node in SCC
+      command: "SUSEConnect -r {{ caasp_registration_key }}"
+
+    - name: Disable the transactional-update timer
+      command: "systemctl disable --now transactional-update.timer"
```
Itxaka: should we re-enable it afterwards? Is there a reason to disable it?

playbooks/openstack-update_caasp.yml
```diff
@@ -0,0 +1,28 @@
...
+    - name: Load variables
+      include_vars: "{{ item }}"
+      loop:
+        - "{{ playbook_dir }}/../vars/common-vars.yml"
+        - "{{ playbook_dir }}/../vars/manifest.yml"
+      tags:
+        - always
+
+    - name: Use Velum tooling to update CaaSP admin node
+      shell: "./velum-interactions --update-admin --update-minions -e {{ upstream_repos_clone_folder }}/ccp/environment.json"
```
Itxaka: do we need update-minions here? shouldnt that be update-admin only?

jsuchome: I splitted it in to tasks in updated commit

playbooks/openstack-prepare_caasp_workers_for_update.yml
jsuchome: TBH I have no idea

Itxaka:
@jsuchome this needs a rebase? or are you gonna do a different approach?

jsuchome:
Closing, we may not need it

PR #165, opened 2019-03-20
airship ci: Add setup_airship target to main run.sh

aagate:
* Add 'setup_airship' target in run.sh to enable
  airship ci.

* Updated 'deploy_airship' in deployment_actions_openstack.sh
  to call generic_deploy_playbook playbook instead of
  non-existent playbook.

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -72,6 +72,10 @@ case "$deployment_action" in
         build_images
         deploy_osh
         ;;
+    "setup_airship")
+        setup_caasp_workers_for_openstack
+        deploy_airship
+        ;;
     "deploy_airship")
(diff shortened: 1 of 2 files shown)


run.sh
```diff
@@ -72,6 +72,10 @@ case "$deployment_action" in
         build_images
         deploy_osh
         ;;
+    "setup_airship")
+        setup_caasp_workers_for_openstack
```
jhesketh: We should perhaps rename this role if it is not unique to openstack.

aagate: Yes that will be good, may be as a part of overall reorganizing of osh and airship deploys

PR #166, opened 2019-03-21
Add a new deploy_network step to run.sh

toabctl:
This step does deploy a network stack (only OpenStack, no KVM
currently) which can then be used for further deployments (eg SES,
CaaSP, ...).
You can use the new step with:

./run.sh deploy_network

Given that the network setup is a Heat stack, the cleanup is pretty
easy and you don't have to manually delete the network/subnet/router .
Cleanup is also integrated when calling

./run.sh teardown

The variables INTERNAL_NETWORK and INTERNAL_SUBNET are no longer
needed. Theses names are calculated from SOCOK8S_ENVNAME. If needed,
the ansible variables for theses network/subnet names can still be
overwritten.

Note: The new step (deploy_network) is enabled by default. So if you
run "./run.sh", a new network stack will be created.

### DIFF
--- a/doc/source/deploy-guide/prepare-localhost.rst
+++ b/doc/source/deploy-guide/prepare-localhost.rst
@@ -241,30 +241,12 @@ Set this for **all** the following scripts in a deployment:
 
 .. code-block:: console
 
+   export SOCOK8S_ENVNAME='foctodoodle'
+   # 'engcloud' is the name in the `clouds.yaml`
    export OS_CLOUD=engcloud
-   # 'engcloud' is the name in the `clouds.yaml`,
    # Set the name of the keypair you created
    export KEYNAME=foctodoodle-key
-   # Set the name of the network you will use in the deployment
-   export INTERNAL_NETWORK=foctodoodle-net
-   # Set the name of the subnet you will use in the deployment
-   export INTERNAL_SUBNET=foctodoodle-subnet
-   # Set the name of the floating ip network you will use in the deployment
    export EXTERNAL_NETWORK=floating
-
-If you haven't created the internal networks/subnets and appropriate routers
-already, do it now (you only have to do it once):
-
-.. code-block:: console
-
-   openstack network create ${INTERNAL_NETWORK}
-   openstack subnet create --network ${INTERNAL_NETWORK} --subnet-range 192.168.100.0/24 ${INTERNAL_SUBNET}
-   openstack router create ${INTERNAL_NETWORK}-router
-   openstack router set --external-gateway floating ${INTERNAL_NETWORK}-router
-   openstack router add subnet ${INTERNAL_NETWORK}-router ${INTERNAL_SUBNET}
-
-Reconfirming that you’ve done all the previous steps to set up now will
-save you some time later.
 
 With this done, proceed to next section of the documentation,
 :ref:`targethosts`.
--- a/heat-templates/openstack-network.yml
+++ b/heat-templates/openstack-network.yml
@@ -1,0 +1,58 @@
+heat_template_version: 2015-04-30
+
+description: >
+  Template to generate the socok8s network including a router
+  to access the external network
+
+parameters:
+  prefix:
+    type: string
+    label: Stack name
+    description: Prefix for the resource names
+  network_external:
+    type: string
+    label: Public network name
+    description: The name of the public network
+    default: floating
+
+resources:
+  network:
+    type: OS::Neutron::Net
+    properties:
+      name:
+        list_join: ['-', [ { get_param: prefix }, 'net']]
+
+  subnet:
+    type: OS::Neutron::Subnet
+    properties:
+      network_id: { get_resource: network }
+      cidr: "192.168.100.0/24"
+      ip_version: 4
+      gateway_ip: 192.168.100.1
+      name:
+        list_join: ['-', [ { get_param: prefix }, 'subnet']]
+
+  router:
+    type: OS::Neutron::Router
+    properties:
+      name:
+        list_join: ['-', [ { get_param: prefix }, 'router']]
+      external_gateway_info:
+        network: { get_param: network_external }
+
+  router_interface:
+    type: OS::Neutron::RouterInterface
+    properties:
+      router_id: { get_resource: router }
+      subnet_id: { get_resource: subnet }
+
+outputs:
+  network_name:
+    description: Network name
+    value: { get_resource: network }
+  subnet_name:
+    description: Subnet name
+    value: { get_resource: subnet }
+  router_name:
+    description: Router name
+    value: { get_resource: router }
--- a/playbooks/openstack-cleanup_network.yml
+++ b/playbooks/openstack-cleanup_network.yml
@@ -1,0 +1,18 @@
+---
+- hosts: localhost
+  gather_facts: no
+  connection: local
+  post_tasks:
+    - name: Load generic vars
+      include_vars: "{{ playbook_dir }}/../vars/common-vars.yml"
+    - name: Load openstack vars
+      include_vars: "{{ playbook_dir }}/../vars/deploy-on-openstack.yml"
+
+    - name: Delete openstack network
+      block:
+        - name: delete network stack
+          os_stack:
+            cloud: "{{ deploy_on_openstack_cloudname }}"
+            state: "absent"
+            name: "{{ deploy_on_openstack_network_stackname }}"
+
--- a/playbooks/openstack-deploy_network.yml
+++ b/playbooks/openstack-deploy_network.yml
@@ -1,0 +1,26 @@
+---
+- hosts: localhost
+  gather_facts: no
+  connection: local
+  post_tasks:
+    - name: Load generic vars
+      include_vars: "{{ playbook_dir }}/../vars/common-vars.yml"
+    - name: Load openstack vars
+      include_vars: "{{ playbook_dir }}/../vars/deploy-on-openstack.yml"
+
+    - name: Create/Update openstack network
+      block:
+        - name: create network stack
+          register: network_stack_create_output
+          os_stack:
+            cloud: "{{ deploy_on_openstack_cloudname }}"
+            state: "present"
+            name: "{{ deploy_on_openstack_network_stackname }}"
+            template: "{{ playbook_dir }}/../heat-templates/openstack-network.yml"
+            parameters:
+              prefix: "{{ socok8s_envname }}"
+              network_external: "{{ deploy_on_openstack_external_network }}"
+
+        - name: show network stack output
+          debug:
+            var: network_stack_create_output
--- a/run.sh
+++ b/run.sh
@@ -38,6 +38,9 @@ source ${scripts_absolute_dir}/deployment-actions-${DEPLOYMENT_MECHANISM}.sh
 deployment_action=${1:-"setup_everything"}
 
 case "$deployment_action" in
+    "deploy_network")
+        deploy_network
+        ;;
     "deploy_ses")
         deploy_ses
         ;;
@@ -71,6 +74,7 @@ case "$deployment_action" in
         setup_caasp_workers_for_openstack
         ;;
     "setup_hosts")
+        deploy_network
         deploy_ses
         deploy_caasp
         deploy_ccp_deployer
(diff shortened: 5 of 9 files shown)


playbooks/openstack-deploy_network.yml
```diff
@@ -0,0 +1,30 @@
+---
+- hosts: localhost
+  gather_facts: no
+  connection: local
+  roles:
```
evrardjp: you don't need ansible-config_template here.

run.sh
```diff
@@ -34,6 +34,9 @@ source ${scripts_absolute_dir}/deployment-actions-${DEPLOYMENT_MECHANISM}.sh
 deployment_action=${1:-"setup_everything"}
 
 case "$deployment_action" in
+    "deploy_network")
```
evrardjp: It's kinda weird to have a new interface for "deploying networks". But I could be good with it. @jgu17 do you see its value on the airship side to have a user interface to deploy networks, or should we hide this?

doc/source/deploy-guide/prepare-localhost.rst
```diff
@@ -182,11 +182,7 @@ floating network (you only have to do it once):
 
 .. code-block:: console
 
-   openstack network create ${PREFIX}-net
-   openstack subnet create --network ${PREFIX}-net --subnet-range 192.168.100.0/24 ${PREFIX}-subnet
-   openstack router create ${PREFIX}-router
-   openstack router set --external-gateway floating ${PREFIX}-router
-   openstack router add subnet ${PREFIX}-router ${PREFIX}-subnet
+   ./run.sh deploy_network
```
evrardjp: This is fine, but it hides the necessary input for deploy_network (the need to pass prefix for example). How do you pass prefix btw?

run.sh
```diff
@@ -34,6 +34,9 @@ source ${scripts_absolute_dir}/deployment-actions-${DEPLOYMENT_MECHANISM}.sh
...
+        deploy_network
+        ;;
```
evrardjp: This means it's not on the standard deploy path, which is fine for CI (which would reuse networks), and convenient for devs. :+1:

doc/source/deploy-guide/prepare-localhost.rst
toabctl: WIth the heat parameter called prefix and then I set this in playbooks/openstack-deploy_network.yml via parameters

evrardjp: I totally missed that L21. Cool thanks.

playbooks/openstack-deploy_network.yml
toabctl: Done

playbooks/openstack-deploy_network.yml
```diff
@@ -0,0 +1,28 @@
...
+              prefix: "{{ deploy_on_openstack_prefix }}"
+              network_external: "{{ deploy_on_openstack_external_network }}"
+
+        - name: show network stack output
+          debug:
+            var: network_stack_create_output
+
+        - name: Get network name
+          debug:
+            msg: "{{ (network_stack_create_output.stack.outputs | selectattr('output_key', 'search', 'network_name') | list | first).output_value }}"
```
jhesketh: We should probably write the stack name out to a file like we do for ~/suse-socok8s-deploy/stackname. Then we can check if the stack already exists before creating it to make this more idempotent.
It would also be good to have a teardown function for this.

evrardjp: Wouldn't the stackname (deploy_on_openstack_prefix) be enough to know what to delete? 

We can probably use the same approach as caasp for the deletion indeed.

toabctl: Both fixed I think. I wrote a file and integrated the cleanup in the teardown function.

script_library/deployment-actions-openstack.sh
```diff
@@ -89,6 +94,8 @@ function clean_openstack(){
     run_ansible ${socok8s_absolute_dir}/playbooks/openstack-deploy_caasp.yml -e caasp_stack_delete=True || true
     echo "Delete SES node"
     run_ansible ${socok8s_absolute_dir}/playbooks/openstack-ses_aio_instance.yml -e ses_node_delete=True
+    echo "Delete network stack"
+    run_ansible ${socok8s_absolute_dir}/playbooks/openstack-cleanup_network.yml
```
evrardjp: This will actively clean things in gates, as we are running this function. So it means the gates will remove the network, and not recreate it. That's not ideal.

evrardjp:
I don't think we should merge this, as it will make the CI worse, not better.

script_library/deployment-actions-openstack.sh
toabctl: Then I would prefer to create networks in the default steps instead of not cleaning up networks. I don't see a reason why we should treat networks different to any other resources.

run.sh
toabctl: I changed this. Why would CI reuse networks?

run.sh
```diff
@@ -80,6 +83,7 @@ case "$deployment_action" in
         deploy_airship
         ;;
     "setup_everything")
+	deploy_network
```
jesusaurus: We should also add this to "setup_hosts", right? And maybe we should be calling "setup_hosts" in "setup_everything" instead of repeating the same list of tasks?

evrardjp: Agreed on the fact to add it into setup_hosts (in other words not expose it to end users). (See comment above).

For the setup_everything, I kinda like your idea. Want to propose a PR for that?

run.sh
evrardjp: Because there is no point of re-creating networks? I don't think we should have the overhead of creating networks, while it brings no positive value (it's not like we are making this multi tenant or anything). But I can be convinced otherwise.

toabctl: It's about creating multiple envs in parallel (as we discussed yesterday, we'll have a single variable (currently `PREFIX` but could/should be something like `ENV_NAME`). So deploying the network for the given `ENV_NAME` is needed during a standard deployment imo

run.sh
toabctl: @jesusaurus added it to `setup_hosts` step.
In terms of `setup_everything` - I think we should try to split between
a) the setup of the infrastructure (network, servers/volumes for caasp, ceph, installation of caasp, ceph) 
b) the actual installation on a given environment

A customer will not install this on openstack or libvirt but (I assume) on a already deployed caasp and ceph cluster. So the steps that do that deployment (on a given env) are imo more important (and customer facing) than the other stuff (which is nice for demos, testing, devs, CI, ...)

script_library/deployment-actions-openstack.sh
evrardjp: I guess that's where I differ: Networks don't need to be unique. We could all share the same networks. I don't see a reason to isolate those, because it takes longer to build things in CI for no positive value.

heat-templates/openstack-network.yml
```diff
@@ -0,0 +1,58 @@
...
+  prefix:
+    type: string
+    label: Stack name
+    description: Prefix for the resource names
+  network_external:
+    type: string
+    label: Public network name
+    description: The name of the public network
+    default: floating
+
```
evrardjp: I would prefer if this would be creating network_external, network, subnet directly within the parameters, and in ansible be smart and use the prefix if necessary.

run.sh
evrardjp: Well, we can create multiple environment in parallel while still keeping the efficiency of not creating networks. Although this feature is, for me, interesting. It's a convenience.

script_library/deployment-actions-openstack.sh
toabctl: - It doesn't take that much longer (this is in the range of tens of seconds I would say) to create a new network
- having isolated networks is nice because you know that only your deployed services are on this network (and no other dhcp/pxe/etcd/whatever service is running beside the one you want to have)

Or is your plan to run multiple envs in the same network?

rhafer: > I guess that's where I differ: Networks don't need to be unique. We could all share the same networks

While it might be possible. I really think it's a bad idea. Especially when done just to safe ~10-15s of deployment time. We should have different deployment isolated networkwise.

run.sh
jesusaurus: @toabctl I agree that we should maintain a logical split between (a) infrastructure setup, and (b) the actual installation; but `setup_everything` already blurs that line. My understanding is that we have `setup_hosts` for (a) and then either `setup_airship` or `setup_openstack` for (b); and that `setup_everything` is equivalent to calling `setup_hosts` followed by `setup_openstack`.

run.sh
toabctl: The updated code now checks if there is already a network stack (it's not checking the network/subnet/router/gateway).
So if you have already a network stack (assuming that it is working correctly), then the creation will be skipped.
So it's possible to reuse the current stack now.

script_library/deployment-actions-openstack.sh
toabctl: Current code creates a network stack if the stack is not already available. @evrardjp @rhafer is that ok for now?

playbooks/openstack-deploy_network.yml
```diff
@@ -0,0 +1,40 @@
...
+  post_tasks:
+    - name: Load generic vars
+      include_vars: "{{ playbook_dir }}/../vars/common-vars.yml"
+    - name: Load openstack vars
+      include_vars: "{{ playbook_dir }}/../vars/deploy-on-openstack.yml"
+
+    - name: Check if the network stack already exist
+      command: openstack stack show "{{ deploy_on_openstack_network_stackname }}" -c id -f value
+      # rc=1 means stack does not exist, rc=0 means stack does exist
+      ignore_errors: yes
```
evrardjp: you probably want to remove ignore_errors: yes

and replace it by:
`failed_when:  network_stack_available.rc != 0 or network_stack_available.rc != 1` or `failed_when: False`

You probably want to add `changed_when: False` too, because it's a read only operation, you'll never change things here.

However, I believe that the os_stack is idempotent, so you probably don't need to do all of this.

toabctl: Ah. Thanks for the hints. I thought `os_stack` would create a new stack whenever you call it. But it is updating the stack when the stack is already available. Great!

toabctl: Fixed

heat-templates/openstack-network.yml
rhafer: Is this still relevant? If yes, I think external networks usually can only be created by admin users and need to exist before. So we can create that with the stack.

toabctl: @rhafer I guess you mean we **can not** (instead of can) in case of the external network.
The parameter `network_external` is there to tell the router the external gateway info.

run.sh
toabctl: Update: `os_stack` automatically creates (if stack does not exist) or update (if stack is already there) the stack.

run.sh
jsuchome: I do not see reason for hiding, some tasks are for developers only anyway

playbooks/openstack-cleanup_network.yml
```diff
@@ -0,0 +1,22 @@
...
+      block:
+        - name: delete network stack
+          os_stack:
+            cloud: "{{ deploy_on_openstack_cloudname }}"
+            state: "absent"
+            name: "{{ deploy_on_openstack_network_stackname }}"
+
+        - name: Delete network stackname file
+          file:
+            path: "{{ deploy_on_openstack_network_stacknamefile }}"
```
jsuchome: I do not see who is creating this file...

toabctl: leftover from previous commit. fixed!

heat-templates/openstack-network.yml
evrardjp: I think it's fine.

script_library/deployment-actions-openstack.sh
evrardjp: It's okay.

heat-templates/openstack-network.yml
rhafer: > @rhafer I guess you mean we **can not** (instead of can) in case of the external network?

Ooops, yes.

PR #167, opened 2019-03-21
rabbitmq: Don't use common-images-overrides.yml

rhafer:
The RabbitMQ charts in openstack-helm-infra rely on a different images
than what we define in our common overrides (which seems to be mostly
targetting openstack images).
Specifically the "scripted_test" image for rabbitmq needs
"rabbitmq-management" an not  the heat image we're using in the common
overrides file.

### DIFF
--- a/playbooks/roles/deploy-osh/templates/rabbitmq.yaml.j2
+++ b/playbooks/roles/deploy-osh/templates/rabbitmq.yaml.j2
@@ -1,2 +1,1 @@
 ---
-{{ lookup('template','files/common-images-overrides.yml') | from_yaml | to_nice_yaml(indent=2) }}


rhafer:
> Just want to more clarification... is this about sticking to the default image in the RabbitMQ chart?

It's mainly about using the correct image for the cluster-wait pod of the RabbitMQ chart. The default overrides was breaking that. AFAICs you're not using `common-images-overrides.yml` so airship should be fine.

evrardjp:
@jgu17 was right, it's important to be in sync, should we switch to a different override.

Here we didn't include any new override, so we are good. But in the future, we should definitely communicate the "hey, we've set this as default because of x". We can patch the two code paths, as it will be just the same kind of yaml in both places, which will be both easy to find. We can even think of having a single file for overrides, but I guess that could be conflicting (we don't have the same deploy profiles)

PR #168, opened 2019-03-21
Support control plane scale out/in

jgu17:
The change specifically involves the cleanup of site directory before redeploy. otherwise, there can be older manifest files left and cause issues. 

SCRD-7618

### DIFF
--- a/site/soc/software/charts/osh/openstack-compute-kit/nova.yaml
+++ b/site/soc/software/charts/osh/openstack-compute-kit/nova.yaml
@@ -18,6 +18,8 @@ metadata:
         path: .values.conf
       - method: replace
         path: .values.pod
+      - method: merge
+        path: .upgrade.pre
   storagePolicy: cleartext
   substitutions:
     - src:
--- a/playbooks/roles/airship-deploy-ucp/tasks/main.yml
+++ b/playbooks/roles/airship-deploy-ucp/tasks/main.yml
@@ -7,13 +7,19 @@
   set_fact:
     site_path: "{{ upstream_repos_clone_folder }}/airship-treasuremap/site/{{ socok8s_site_name }}"
 
+- name: Ensure site directory for {{ socok8s_site_name }} is empty
+  become: yes
+  file:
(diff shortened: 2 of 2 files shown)


site/soc/software/charts/osh/openstack-compute-kit/nova.yaml
```diff
@@ -63,6 +65,15 @@ metadata:
...
 data:
+  upgrade:
+    pre:
+      delete:
+        - type: pod
+          name: airship-nova-test
+          labels:
+            application: nova
+            component: test
+            release_group: airship-nova
```
kantarun: Looks fine though question is why we need to delete test pod only for nova . I don't see this change for neutron test.

kantarun: Seems like James already have upstream review for this and included in dev-patcher so we may no longer need this.
Upstream review: https://review.openstack.org/#/c/645218

evrardjp: That review is clearer than this  :) 

Hard to understand without context. Thanks @kantarun !

toabctl:
@jgu17 can you remove the merge commits and rebase on top of master please?

jgu17:
> @jgu17 can you remove the merge commits and rebase on top of master please?

Can you explain more? I am more used to gerrit.

toabctl:
> Would squash commit work the same purpose in this specific case? It's really a single commit at the end that I want. Apparently, I shouldn't do a git pull as git asks me to o when I git push to the remote branch the second or third time. It is pulling from master instead of remote branch as I had thought. using rebase will mke more sense for sure.

Not sure about the squash commit. But it would be nice to be able to review what will be merged at the end (without the squash when pressing the merge button) imo.
You can just rebase your commit on top of current master (`git rebase -i upstream/master` assuming that upstream is the remote name for the SUSE-Cloud remote) and then do a `git push -f origin` (assuming origin is your remote)

PR #164, opened 2019-03-20
Upgrade caasp without scc

jsuchome:
The alternative to https://github.com/SUSE-Cloud/socok8s/pull/163

PR #172, opened 2019-03-21
Use CaaSP 3.0.0 image with latest updates

jsuchome:
An alternative to https://github.com/SUSE-Cloud/socok8s/pull/164


jsuchome:
@evrardjp sorry, but where in the deployment guide are we currently sharing that information?

evrardjp:
We are currently not doing it, because we said "deploy caasp". Now that we have more requirements, we should express them in docs. https://github.com/SUSE-Cloud/socok8s/blob/master/doc/source/deploy-guide/prepare-targets.rst would be a good place.

Itxaka:
@jsuchome close this? or do we want to still investigate this path?

jsuchome:
Closing, at least for now

PR #173, opened 2019-03-22
Fix for forbidden error for mariadb service account

rajdharwadkar:
Error:
"HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"forbidden: User \"system:serviceaccount:ucp:airship-ucp-mariadb-mariadb\" cannot get path \"/version/\"","reason":"Forbidden","details":{},"code":403} "


kantarun:
> I don't see a problem with that, but if you want to ensure compatibility with previous naming you should duplicate the ServiceAccount definition, and point to the code where this has changed upstream. So that we can track WHEN to remove the previous naming convention.

No backward compatibility concern here as its still in dev phase and new convention is going to be followed

PR #174, opened 2019-03-22
Allow vendor change when updating osh deployer

jsuchome:
It is safe because it's us who are setting the repositories.

Current error:

```
Computing distribution upgrade...
5 Problems:
Problem: problem with installed package cloud-init-18.2-lp150.11.1.x86_64
Problem: problem with installed package cloud-init-config-suse-18.2-lp150.11.1.x86_64
Problem: problem with installed package growpart-0.30-lp150.19.1.noarch
Problem: problem with installed package python-rpm-macros-2017.12.22.d9968ab-lp150.52.1.noarch
Problem: problem with installed package python3-jmespath-0.9.3-lp150.33.1.noarch

Problem: problem with installed package cloud-init-18.2-lp150.11.1.x86_64
 Solution 1: install cloud-init-18.2-lp150.1.1.x86_64 (with vendor change)
  obs://build.opensuse.org/Cloud  -->  openSUSE
 Solution 2: keep obsolete cloud-init-18.2-lp150.11.1.x86_64
```

### DIFF
--- a/playbooks/openstack-osh_hostconfig.yml
+++ b/playbooks/openstack-osh_hostconfig.yml
@@ -14,6 +14,14 @@
         state: present
       loop: "{{ deploy_on_openstack_osh_repos_per_imagename[deploy_on_openstack_oshnode_image] }}"
 
+    - name: Add known vendors
+      blockinfile:
+        path: /etc/zypp/vendors.d/opensuse
+        block: |
+          [main]
+          vendors = suse,opensuse,obs://build.opensuse.org/Cloud:OpenStack,obs://build.opensuse.org/devel:languages:python
+        create: yes
+
     - name: Zypper dist-upgrade
       zypper:
         name: '*'


evrardjp:
Risky but fine for me.

toabctl:
@jsuchome can you describe (in the commit msg) why this is useful and what problem this solves please?

jsuchome:
@toabctl updated. At least I think this is the issue...

evrardjp:
We could probably be more surgical, but does it matter?

evrardjp:
Need to split into two tasks, or change the ansible module.

playbooks/openstack-osh_hostconfig.yml
```diff
@@ -14,12 +14,17 @@
         state: present
       loop: "{{ deploy_on_openstack_osh_repos_per_imagename[deploy_on_openstack_oshnode_image] }}"
 
+    - name: Refresh all repositories
```
evrardjp: I think we should do this instead:

- name: Add known vendors
  blockinfile:
    path: /etc/zypp/vendors.d/opensuse
    block: |
      [main]
      vendors = "suse,opensuse,obs://build.opensuse.org/Cloud:OpenStack,obs://build.opensuse.org/devel:languages:python"
    create: yes

jhesketh:
I like JP's improvement, but happy either way

jsuchome:
OK, I've used @evrardjp proposal instead

evrardjp:
I have tried both in my env (the --alow-vendor-change, and the file with the vendors), and both fail temporarily in the upgrade. I don't know what to do from that on.

jsuchome:
How did ` --alow-vendor-change` fail? I mean after the removal of that refresh...

evrardjp:
The problem is not the task itself, is the dup which failed.

jsuchome:
> is the dup which failed.

Yes, but how? It did not fail because of vendor change, right?

evrardjp:
I am double checking with a fresh environment.

PR #175, opened 2019-03-22
Horizon: add overrides + patches for suse images

Itxaka:
Adds 2 upstream patches to the dev patcher for horizon to work
plus provides the proper overrides in the horizon template for helm

Also removes an old horizon patch that is now abandoned in favor of
the new ones

### DIFF
--- a/playbooks/roles/deploy-osh/templates/horizon.yaml.j2
+++ b/playbooks/roles/deploy-osh/templates/horizon.yaml.j2
@@ -1,2 +1,31 @@
 ---
-{{ lookup('template','files/common-images-overrides.yml') | from_yaml | to_nice_yaml(indent=2) }}
+{% if developer_mode %}
+conf:
+  apache2:
+    binary: apache2ctl
+    start_parameters: -DFOREGROUND -k start
+    site_dir: /etc/apache2/vhosts.d
+    conf_dir: /etc/apache2/conf.d
+    a2enmod:
+      - rewrite
+      - version
+  horizon:
+    security: |
+      <Directory "/var/www">
+      Options Indexes FollowSymLinks
+      AllowOverride All
+      <IfModule !mod_access_compat.c>
+        Require all granted
+      </IfModule>
+      <IfModule mod_access_compat.c>
+        Order allow,deny
+        Allow from all
+      </IfModule>
+      </Directory>
+images:
+  tags:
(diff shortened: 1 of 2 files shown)


Itxaka:
not working due to the ansible-config-template bug :cry:

playbooks/roles/deploy-osh/templates/horizon.yaml.j2
```diff
@@ -1,2 +1,27 @@
...
+      <IfModule !mod_access_compat.c>
+        Require all granted
+      </IfModule>
+      <IfModule mod_access_compat.c>
+        Order allow,deny
+        Allow from all
+      </IfModule>
+      </Directory>
+{%  endif %}
 {{ lookup('template','files/common-images-overrides.yml') | from_yaml | to_nice_yaml(indent=2) }}
```
evrardjp: You can probably remove this common images overrides. It's useless now.

playbooks/roles/deploy-osh/templates/horizon.yaml.j2
```diff
@@ -1,2 +1,27 @@
 ---
+{% if developer_mode %}
```
evrardjp: We should probably add an override here to point to your built image if developer mode.

playbooks/roles/deploy-osh/templates/horizon.yaml.j2
Itxaka: out of scope for this patch? Separate patch better?

playbooks/roles/deploy-osh/templates/horizon.yaml.j2
Itxaka: out of scope for this patch? Separate patch better?

Itxaka:
not merge until https://github.com/SUSE-Cloud/socok8s/pull/178 gets in and we rework the templates

playbooks/roles/deploy-osh/templates/horizon.yaml.j2
evrardjp: The template could mask any of your overrides, and create duplicate keys. it's a bad bad idea.

jsuchome: It's not related to this change, I would agree with separate PR (or at least separate commit)

Itxaka:
For me cleaning up my k8s resources and deploying it from scratch made it work as expected. Currently redeploying it several times to see if I can hit that again (may have something to do with leftovers from old deploys?)

PR #177, opened 2019-03-22
images: do not override the tempest+rally images

Itxaka:
We are not building nor plan to do for the moment so we should just
use the default tempest and test(xrally) images


evrardjp:
About the commit message, tempest should still be built IIRC. xrally is an official one, tempest isn't.

Itxaka:
> About the commit message, tempest should still be built IIRC. xrally is an official one, tempest isn't.

I believe that more work is needed as loci doesnt support tempest for now no? Due to it not following the naming releases I guess? (so there is no rocky release for tempest)

PR #178, opened 2019-03-22
Simplify templating


evrardjp:
Config template has a bug in multiline, and we don't know when
it will get solved.

We can skip config_template and rely on kubernetes has merge
behaviour. For that, we just need to pass the files in the cli,
on the right of the rest.

This adds this feature, and at the same times, cleans the hacks
we had around to use /bin/true to generate templates.

### DIFF
--- a/playbooks/roles/deploy-osh/tasks/component-install.yml
+++ b/playbooks/roles/deploy-osh/tasks/component-install.yml
@@ -1,21 +1,24 @@
 ---
-- name: "Prepare override file {{ _cmpnt_template }}"
-  config_template:
-    src: "{{ _cmpnt_template }}.j2"
-    dest: "/tmp/socok8s-{{ _cmpnt_template }}"
-    config_overrides: "{{ _cmpnt_overrides }}"
-    config_type: yaml
-  register: _usertemplatedata
+- name: Generate SUSE defaults
+  loop: "{{ _component_overrides }}"
+  template:
+    src: "{{ item.suse_file }}.j2"
+    dest: "/tmp/socok8s-susedefaults-{{ item.suse_file }}"
+
+- name: Generate user overrides
+  loop: "{{ _component_overrides }}"
+  copy:
+    content: "{{ item.user_overrides | to_nice_yaml }}"
+    dest: "/tmp/socok8s-useroverrides-{{ item.suse_file }}"
 
 # This needs specific executable, and therefore needs shell module,
 # which causes an ansible-lint issue (no idempotency test). In the meantime
 # that idempotency tests are created, we skip_ansible_lint
 - name: Run the override file
-  shell: "{{ _cmpnt_runcommand }}"
+  shell: "{{ _component_runcommand }}"
   args:
     executable: /bin/bash
-    chdir: "{{ _cmpnt_chdir | default('/opt/openstack-helm') }}"
-  environment:
-    OSH_INFRA_PATH: /opt/openstack-helm-infra
+    chdir: "/opt/openstack-helm"
+  environment: "{{ base_env_vars | combine(_component_run_env) }}"
   tags:
     - skip_ansible_lint
--- a/playbooks/roles/deploy-osh/tasks/main.yml
+++ b/playbooks/roles/deploy-osh/tasks/main.yml
@@ -177,14 +177,14 @@
   vars:
     _component_runcommand: ./tools/deployment/multinode/020-ingress.sh
     _component_run_env:
-      OSH_EXTRA_HELM_ARGS_INGRESS_OPENSTACK: >
+      OSH_EXTRA_HELM_ARGS_INGRESS_openstack: >
         --values=/tmp/socok8s-susedefaults-ingress-namespace.yaml
         --values=/tmp/socok8s-useroverrides-ingress-namespace.yaml
-        {{ lookup('env', 'OSH_EXTRA_HELM_ARGS_INGRESS_OPENSTACK') | default('', True) }}
-      OSH_EXTRA_HELM_ARGS_INGRESS_CEPH: >
+        {{ lookup('env', 'OSH_EXTRA_HELM_ARGS_INGRESS_openstack') | default('', True) }}
+      OSH_EXTRA_HELM_ARGS_INGRESS_ceph: >
         --values=/tmp/socok8s-susedefaults-ingress-namespace.yaml
         --values=/tmp/socok8s-useroverrides-ingress-namespace.yaml
-        {{ lookup('env', 'OSH_EXTRA_HELM_ARGS_INGRESS_CEPH') | default('', True) }}
+        {{ lookup('env', 'OSH_EXTRA_HELM_ARGS_INGRESS_ceph') | default('', True) }}
       OSH_EXTRA_HELM_ARGS_INGRESS_KUBE_SYSTEM: >
         --values=/tmp/socok8s-susedefaults-ingress-kube-system.yaml
         --values=/tmp/socok8s-useroverrides-ingress-kube-system.yaml
--- a/playbooks/roles/deploy-osh/tasks/main.yml
+++ b/playbooks/roles/deploy-osh/tasks/main.yml
@@ -172,23 +172,28 @@
     - ingress
     - run
 
-- name: Template file for the namespaces ingress
-  config_template:
-    src: ingress-namespace.yaml.j2
-    dest: /tmp/socok8s-ingress-namespace.yaml
-    config_overrides: "{{ suse_osh_deploy_ingress_namespace_yaml_overrides }}"
-    config_type: yaml
-
 - name: Setup Basic ingress and namespace
   import_tasks: component-install.yml
   vars:
-    _cmpnt_template: ingress-kube-system.yaml
-    _cmpnt_overrides: "{{ suse_osh_deploy_ingress_kube_system_yaml_overrides }}"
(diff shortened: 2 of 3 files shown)


playbooks/roles/deploy-osh/tasks/component-install.yml
```diff
@@ -1,21 +1,39 @@
...
-    config_overrides: "{{ _cmpnt_overrides }}"
-    config_type: yaml
-  register: _usertemplatedata
+# Example structure
+#_cmpnt_runcommand: ./tools/deployment/multinode/020-ingress.sh
+#_cmpnt_overrides:
+#  - suse: ingress-namespace.yaml
+#    user: "{{ suse_osh_deploy_ingress_namespace_yaml_overrides }}"
+#    envvars_getting_those_overrides:
+#      - "OSH_EXTRA_HELM_ARGS_INGRESS_OPENSTACK"
```
jhesketh: Can't we extract these envvars from the override files rather than having to define them separately?

playbooks/roles/deploy-osh/tasks/component-runtime-setup.yml
```diff
@@ -0,0 +1,27 @@
+---
+
+- name: "Generate suse override for {{ override.suse }}"
+  template:
+    src: "{{ override.suse }}.j2"
```
jsuchome: Why `suse`? The meaning is really path/file name

playbooks/roles/deploy-osh/tasks/component-runtime-setup.yml
```diff
@@ -0,0 +1,27 @@
...
+    dest: "/tmp/socok8s-suse-{{ override.suse }}"
+
+- name: Generate user overrides
+  copy:
+    content: "{{ override.user | to_nice_yaml }}"
+    dest: "/tmp/socok8s-user-{{ override.suse }}"
```
jsuchome: Also seen here: it would make more sense if the variable would be named file/path

playbooks/roles/deploy-osh/tasks/main.yml
```diff
@@ -172,56 +172,56 @@
...
   import_tasks: component-install.yml
   vars:
-    _cmpnt_template: ingress-kube-system.yaml
-    _cmpnt_overrides: "{{ suse_osh_deploy_ingress_kube_system_yaml_overrides }}"
     _cmpnt_runcommand: ./tools/deployment/multinode/020-ingress.sh
-  environment:
-    OSH_EXTRA_HELM_ARGS_INGRESS_KUBE_SYSTEM: "{{ lookup('env', 'OSH_EXTRA_HELM_ARGS_INGRESS_KUBE_SYSTEM') | default('', True) }} --values /tmp/socok8s-ingress-kube-system.yaml"
-    OSH_EXTRA_HELM_ARGS_INGRESS_OPENSTACK: "{{ lookup('env', 'OSH_EXTRA_HELM_ARGS_INGRESS_OPENSTACK') | default('', True) }} --values /tmp/socok8s-ingress-namespace.yaml"
-    OSH_EXTRA_HELM_ARGS_INGRESS_CEPH: "{{ lookup('env', 'OSH_EXTRA_HELM_ARGS_INGRESS_CEPH') | default('', True) }} --values /tmp/socok8s-ingress-namespace.yaml"
+    _cmpnt_overrides:
```
jsuchome: Could you just use `component` instead of `cmpnt` everywhere?

playbooks/roles/deploy-osh/tasks/main.yml
```diff
@@ -172,56 +172,56 @@
...
+      - suse: ingress-namespace.yaml
+        user: "{{ suse_osh_deploy_ingress_namespace_yaml_overrides }}"
+        envvars_getting_those_overrides:
```
jsuchome: Why such cryptic variable name?

playbooks/roles/deploy-osh/tasks/component-runtime-setup.yml
evrardjp: it's the suse-ification?

jsuchome: It is, in general
-  but there's no non-suse user of socok8s and plenty of other pieces of code are SUSE specific
- I think variable name should explain it's meaning, and "suse" for a name of file just does not

playbooks/roles/deploy-osh/tasks/component-install.yml
evrardjp: What does this bring? I think it would mask some thing and therefore be harder to understand?

playbooks/roles/deploy-osh/tasks/component-runtime-setup.yml
evrardjp: would `susefile` and `uservar` to replace `suse` and `user` be better?

playbooks/roles/deploy-osh/tasks/main.yml
evrardjp: Could you propose an alternative?

playbooks/roles/deploy-osh/tasks/component-runtime-setup.yml
jsuchome: susefile sounds better, yes

playbooks/roles/deploy-osh/tasks/main.yml
jsuchome: `extra_overrides`?

playbooks/roles/deploy-osh/tasks/component-install.yml
jhesketh: Maybe I'm confused by their purpose, but aren't these defined within the supplied override.suse/user files? If so, aren't we repeating them here?

playbooks/roles/deploy-osh/tasks/main.yml
evrardjp: it's not extra overrides, it's to know where the overrides apply.

playbooks/roles/deploy-osh/tasks/component-install.yml
evrardjp: the overrides is just template generation. We need to say where and how to apply the overrides. This is done through env vars. So basically the script x can be passed env vars y and z getting the overrides for components y and z.

evrardjp:
With the amount of reviews, it seems the simplification is not _simple_. I will go back to the drawing board, but we've spent already too much time on this.

evrardjp:
To be fixed.

playbooks/roles/deploy-osh/tasks/main.yml
```diff
@@ -172,56 +172,76 @@
...
+    _component_runcommand: ./tools/deployment/multinode/020-ingress.sh
+    _component_run_env:
+      OSH_EXTRA_HELM_ARGS_INGRESS_OPENSTACK: >
+        --values=/tmp/socok8s-susedefaults-ingress-namespace.yaml
+        --values=/tmp/socok8s-useroverrides-ingress-namespace.yaml
+        {{ lookup('env', 'OSH_EXTRA_HELM_ARGS_INGRESS_OPENSTACK') | default('', True) }}
+      OSH_EXTRA_HELM_ARGS_INGRESS_CEPH: >
+        --values=/tmp/socok8s-susedefaults-ingress-namespace.yaml
+        --values=/tmp/socok8s-useroverrides-ingress-namespace.yaml
+        {{ lookup('env', 'OSH_EXTRA_HELM_ARGS_INGRESS_OPENSTACK') | default('', True) }}
```
jsuchome: OSH_EXTRA_HELM_ARGS_INGRESS_CEPH

evrardjp: Fixed.

playbooks/roles/deploy-osh/tasks/main.yml
evrardjp: Used different name, should be clearer.

playbooks/roles/deploy-osh/tasks/main.yml
evrardjp: Done.

playbooks/roles/deploy-osh/tasks/component-runtime-setup.yml
evrardjp: Done.

playbooks/roles/deploy-osh/tasks/component-runtime-setup.yml
evrardjp: Done.

playbooks/roles/deploy-osh/tasks/component-install.yml
evrardjp: Should be clearer now.

jsuchome:
My understanding is that `suse_file` are those we, as SUSE are providing with socok8s and are different from upstream OSH.
And `user_overrides` are basically meant for developers that want to provide yet something extra, basically for testing. In my understanding such change will most likely land in `suse_file` later. And it's also possible some change we have in `suse_file` is gonna moved upstream at some point of time

evrardjp:
Correct, `suse_file` is our overrides, `user_overrides` are the deployer's. Deployer can provide a simple dict notation (for quick hacks) in extravars, or a file content using a lookup.

Assuming `user_overrides` are shared from deployers to this repo, some could eventually become part of the "official" overrides in `suse_file`. We can decide what is deployer specific, and what is the official stance of SUSE.

PR #180, opened 2019-03-22
add new compute node to existing cluster

rajdharwadkar:
First add new compute node in inventory file and then run following command

Command to add compute :
./run.sh add_compute
This will configures new compute node and run os_deploy.


playbooks/add_compute.yml
```diff
@@ -0,0 +1,15 @@
+---
+- hosts: airship-openstack-compute-workers
+  gather_facts: yes
+  any_errors_fatal: true
+  roles:
+    - role: airship-configure-worker
+      when: redeploy_osh_only is defined and redeploy_osh_only
```
kantarun: Why this condition is needed ? There is no relation to flag with add compute logic

playbooks/add_compute.yml
```diff
@@ -0,0 +1,15 @@
+---
```
kantarun: Why we need this as separate file. Cannot we just add tag and use related logic in deploy_airship flow

rajdharwadkar: Yeah but this is cleaner way of doing because that playbook does other tasks.

playbooks/add_compute.yml
rajdharwadkar: yeah I have fixed that one.

PR #182, opened 2019-03-25
Keystone: add overrides + patch for suse images

Itxaka:
Adds upstream patch to the dev patcher for keystone to work
plus provides the proper overrides in the keystone template for helm

### DIFF
--- a/playbooks/roles/deploy-osh/templates/keystone.yaml.j2
+++ b/playbooks/roles/deploy-osh/templates/keystone.yaml.j2
@@ -1,2 +1,46 @@
 ---
-{{ lookup('template','files/common-images-overrides.yml') | from_yaml | to_nice_yaml(indent=2) }}
+#TODO(itxaka): Remove the developer mode when 647403 is merged
+{% if developer_mode %}
+conf:
+  apache2:
+    binary: apache2ctl
+    start_parameters: -DFOREGROUND -k start
+    site_dir: /etc/apache2/vhosts.d
+    conf_dir: /etc/apache2/conf.d
+    a2enmod:
+      - version
+    security: |
+      <Directory "/var/www">
+      Options Indexes FollowSymLinks
+      AllowOverride All
+      <IfModule !mod_access_compat.c>
+        Require all granted
+      </IfModule>
+      <IfModule mod_access_compat.c>
+        Order allow,deny
+        Allow from all
+      </IfModule>
+      </Directory>
+pod:
+  security_context:
+    keystone:
+      pod:
+        runAsUser: 0
+      container:
+        keystone_api:
+          readOnlyRootFilesystem: false
+images:
+  tags:
+    bootstrap: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+    db_init: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+    db_drop: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+    keystone_api: "{{ suse_osh_registry_location }}/openstackhelm/keystone:{{ suse_openstack_image_version }}"
+    keystone_bootstrap: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+    keystone_credential_rotate: "{{ suse_osh_registry_location }}/openstackhelm/keystone:{{ suse_openstack_image_version }}"
+    keystone_credential_setup: "{{ suse_osh_registry_location }}/openstackhelm/keystone:{{ suse_openstack_image_version }}"
+    keystone_db_sync: "{{ suse_osh_registry_location }}/openstackhelm/keystone:{{ suse_openstack_image_version }}"
+    keystone_domain_manage: "{{ suse_osh_registry_location }}/openstackhelm/keystone:{{ suse_openstack_image_version }}"
+    keystone_fernet_rotate: "{{ suse_osh_registry_location }}/openstackhelm/keystone:{{ suse_openstack_image_version }}"
+    keystone_fernet_setup: "{{ suse_osh_registry_location }}/openstackhelm/keystone:{{ suse_openstack_image_version }}"
+    ks_user: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+{%  endif %}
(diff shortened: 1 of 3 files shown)


Itxaka:
not merge until #178 gets in and we rework the templates

playbooks/roles/deploy-osh/templates/keystone.yaml.j2
```diff
@@ -1,2 +1,22 @@
...
+      <IfModule !mod_access_compat.c>
+        Require all granted
+      </IfModule>
+      <IfModule mod_access_compat.c>
+        Order allow,deny
+        Allow from all
+      </IfModule>
+      </Directory>
+{%  endif %}
 {{ lookup('template','files/common-images-overrides.yml') | from_yaml | to_nice_yaml(indent=2) }}
```
evrardjp: Can you remove the lookup template, and point to the developer built images, when developer mode is enabled?

evrardjp:
why dnm?

evrardjp:
typo: Kesytone.

playbooks/roles/deploy-osh/templates/keystone.yaml.j2
Itxaka: That is out of scope for this patch, should be in a different one

playbooks/roles/deploy-osh/templates/keystone.yaml.j2
```diff
@@ -1,2 +1,40 @@
...
+      Options Indexes FollowSymLinks
+      AllowOverride All
+      <IfModule !mod_access_compat.c>
+        Require all granted
+      </IfModule>
+      <IfModule mod_access_compat.c>
+        Order allow,deny
+        Allow from all
+      </IfModule>
+      </Directory>
```
jsuchome: We also need
```
    a2enmod:
      - mod_version
```

since there's a `IfVersion` check in wsgi-keystone.conf and the module is apparently not included by default

Itxaka: Good catch. I think we may need to include that in all of our apache related overrides as it seems to be used all around.

Itxaka: FYI `version` and `mod_version` both seems to work

playbooks/roles/deploy-osh/templates/keystone.yaml.j2
```diff
@@ -1,2 +1,42 @@
...
+pod:
+  security_context:
+    keystone:
+      pod:
+        runAsUser: 0
```
jsuchome: I think we also need
```
      container:
        keystone_api:
          readOnlyRootFilesystem: false
```

So that apache is able to write into `/etc/apache2/sysconfig.d/`

playbooks/roles/deploy-osh/templates/keystone.yaml.j2
jsuchome: Yes, I think it will be needed in all similar PR's

playbooks/roles/deploy-osh/templates/keystone.yaml.j2
jsuchome: It is set to false by default with https://review.openstack.org/#/c/644627/

evrardjp: Valid point, could you update?

jsuchome:
With `readOnlyRootFilesystem: false` and 
https://github.com/SUSE-Cloud/socok8s/pull/178 it seems to work

playbooks/roles/deploy-osh/templates/keystone.yaml.j2
Itxaka: updated!

playbooks/roles/deploy-osh/templates/keystone.yaml.j2
```diff
@@ -1,2 +1,45 @@
...
+      container:
+        keystone_api:
+          readOnlyRootFilesystem: false
+images:
+  tags:
+    bootstrap: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_osh_image_version }}"
```
evrardjp: In this patch, we will only focus on keystone, so I expect we should not update heat image version, even in developer mode. Could we create a new var for overriding the keystone images, and we could slowly phase all the overrides to use that image? 

Would that work with keystone <rocky> and the other images staying the same?

Itxaka: well, that image is part of the keystone chart so it makes sense to bind them together no? After all, if we dont do this, we would get a keystone deployment that contains suse images + ubuntu images mixed in there.

I did understand that as part of the overrides we want to make the whole chart work under suse.

evrardjp: ok so you mean you've tested only with keystone + heat images together?

playbooks/roles/deploy-osh/templates/keystone.yaml.j2
```diff
@@ -1,2 +1,45 @@
 ---
-{{ lookup('template','files/common-images-overrides.yml') | from_yaml | to_nice_yaml(indent=2) }}
+{% if developer_mode %}
```
evrardjp: Could you add a comment like:
```TODO(itxaka): Remove the developer mode when 647403 is merged```

playbooks/roles/deploy-osh/templates/keystone.yaml.j2
evrardjp: In this case, we should, when developer mode, still point to a different value than ocata (present in the var suse_osh_image_version).

I am not really fond of splitting this into 2 PRs, as they are just one single thing. But I am not in favor of updating suse_osh_image_version across the board for non-keystone.

Itxaka: > ok so you mean you've tested only with keystone + heat images together?

yep, only tested with all suse images, no idea if it will work with a mix

> n this case, we should, when developer mode, still point to a different value than ocata (present in the var suse_osh_image_version).

Shouldnt that be done via the change in https://github.com/SUSE-Cloud/socok8s/pull/188 so we cna point to the new var across the board?

I can just use an extra var here `suse_osh_image_tag_kesytone` but it seems a bit superfluous to do it override by override where we could do it as a whole and have it synced on all overrides

playbooks/roles/deploy-osh/templates/keystone.yaml.j2
Itxaka: done!

Itxaka:
NOTICE:
You need to remove your `suse_osh_registry_location` var from extravars so it points to the default `docker.io` value

evrardjp:
This should go in!

PR #184, opened 2019-03-25
Command to just update airship osh site deployment only

kantarun:
In an existing airship deployment, a user may only want to apply
changes in existing osh (overcloud) site e.g. changing service
configuration, scale out or scale in number of service pods.

Using tags to execute only those ansible steps which are needed for
updating an osh site in a exsiting airship deployment.
Tag used is 'update_airship_osh_site'

Removing unused clean_airship_not_images in openstack deployment
mechanism as its not used (was left in error).

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -79,6 +79,9 @@ case "$deployment_action" in
     "deploy_airship")
         deploy_airship
         ;;
+    "update_airship_osh")
+        deploy_airship update_airship_osh_site
+        ;;
     "setup_everything")
         deploy_ses
         deploy_caasp
--- a/playbooks/generic-deploy_airship.yml
+++ b/playbooks/generic-deploy_airship.yml
@@ -5,12 +5,17 @@
   roles:
     - role: airship-setup-deployer
       when: redeploy_osh_only is not defined or not redeploy_osh_only
+      tags:
+        - install
 
 - hosts: all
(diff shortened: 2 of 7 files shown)


run.sh
```diff
@@ -79,6 +79,9 @@ case "$deployment_action" in
     "deploy_airship")
         deploy_airship
         ;;
+    "update_airship_osh")
```
evrardjp: shouldn't you do two steps instead?

I mean if there are real reasons you want to expose this feature, it might be worth exposing to the end users in a simple way:
setup_hosts -> setup_airship_undercloud -> setup_airship_openstack?

kantarun: Currently deploy_airship is idempotent so it can be executed again. In a way, logic is already separated into steps as internally it calls 2 separate roles for setting up ucp (undercloud) and osh (overcloud)

Reason for adding this is to quickly test and apply changes in osh site. The users mentioned in this can be devs.

If we would have been executing plays directly (e.g. ansible-playbook -i <inventory> <high_level play_name>) , then this would have been a separate play name instead of target for run.sh script.
I was thinking having different target based on most common use cases. From user perspective, deploy_airship and then updating existing osh/overcloud site seems more common pattern. May be add-compute another target as that can be another use-case.

PR #185, opened 2019-03-25
Cleanup jobs if they are not deleted.

rajdharwadkar:


script_library/clean-airship.sh
```diff
@@ -98,6 +98,16 @@ if [[ ${clean_action} == *"clean_images"* ]]; then
     fi
 fi
 
+#If jobs are not deleted.
+if [[ ${clean_action} == *"clean_ucp"* ]]; then
+    kubectl get jobs -n ucp | awk '{print$1}'| xargs kubectl delete job -n ucp
```
kantarun: Any reason we need separate if clause for this. Can this be moved before line #91

script_library/clean-airship.sh
```diff
@@ -98,6 +98,16 @@ if [[ ${clean_action} == *"clean_images"* ]]; then
...
+fi
+
+#If jobs are not deleted.
+if [[ ${clean_action} == *"clean_openstack"* ]]; then
+    kubectl get jobs -n openstack | awk '{print$1}'| xargs kubectl delete job -n openstack
```
kantarun: Any reason we need separate if clause for this. Can this be moved before line 83
https://github.com/SUSE-Cloud/socok8s/blob/17cb308a0ab1844f3800580c1a3f9bd5cb73292c/script_library/clean-airship.sh#L83

PR #188, opened 2019-03-26
We're finally switching from Ocata to Rocky release

jsuchome:
This will wait until Leap based Rocky images are available

### DIFF
--- a/playbooks/roles/deploy-osh/defaults/main.yml
+++ b/playbooks/roles/deploy-osh/defaults/main.yml
@@ -79,7 +79,7 @@ suse_osh_deploy_always_redeploy: True
 # included in all the places).
 # For a more surgical per component approach, use service overrides, like
 # suse_osh_deploy_cinder_yaml_overrides for cinder for example.
-suse_osh_image_version: "ocata"
+suse_osh_image_version: "rocky"
 
 # Default overrides
 suse_osh_deploy_cinder_yaml_overrides: {}


evrardjp:
This will not be enough, because the image name will not be simply `rocky` but instead it will be `rocky-opensuse_15`.

We need to adapt the templates accordingly.

Itxaka:
so `latest-rocky-suse_15` or should we use a different var for this? `suse_osh_image_tag: "latest-{{ suse_osh_image_version }}-suse_15"`?

BTW, this is also being used by airship under `site/soc/software/config/versions.yaml` but they are overriding. Bad thing, as they should not overlap.

Changing it to a different var just for the images should work for us I guess without affecting the airship part, which still needs fixing.

playbooks/roles/deploy-osh/defaults/main.yml
```diff
@@ -79,7 +79,7 @@ suse_osh_deploy_always_redeploy: True
 # included in all the places).
 # For a more surgical per component approach, use service overrides, like
 # suse_osh_deploy_cinder_yaml_overrides for cinder for example.
-suse_osh_image_version: "ocata"
+suse_osh_image_version: "rocky"
```
evrardjp: I guess we should create multiple variables, as it will be different if it's memcached or an openstack component.

Maybe:

```
suse_nonopenstack_image_version: "latest-opensuse_15"
```

?

Of we need to adapt all the references were this is used...

Shouldn't this be done organically when doing the overrides in the helm charts instead, though?

PR #189, opened 2019-03-26
Remove all references of non-workdir folders


evrardjp:
This ensures all the files which were outside the work directory
(ANSIBLE_RUNNER_DIR) are now inside it.

This allows a self-contained build, reproducible in CI.
It also changes the user documentation acccordingly.

### DIFF
--- a/playbooks/generic-clean_userfiles.yml
+++ b/playbooks/generic-clean_userfiles.yml
@@ -1,0 +1,34 @@
+---
+#
+# (c) Copyright 2019 SUSE LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+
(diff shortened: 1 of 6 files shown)


playbooks/generic-clean_userfiles.yml
```diff
@@ -0,0 +1,34 @@
...
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+
+- hosts: localhost
+  gather_facts: no
+  connection: local
+  vars_prompt:
+    - name: "delete_anyway"
```
jhesketh: (nit) It's not really "delete_anyway", perhaps "delete_user_workspace"?

playbooks/generic-clean_userfiles.yml
```diff
@@ -0,0 +1,34 @@
...
+      prompt: "Delete all files in workspace?"
+      default: "no"
+  tasks:
+    - name: Load workspace var
+      include_vars: "{{ playbook_dir }}/../vars/common-vars.yml"
+      when: delete_anyway | bool
```
jhesketh: (nit) inconsistent double quotes with below. Not sure which is preferred.

playbooks/generic-clean_userfiles.yml
```diff
@@ -0,0 +1,34 @@
...
+    - name: Deleting files in workspace
+      file:
+        path: "{{ socok8s_workspace }}"
+        state: absent
+      when:
+        - "delete_anyway | bool"
```
jhesketh: (nit) could be done as a block rather than having "delete_anyway" on both tasks.

PR #191, opened 2019-03-27
show the component name on the helm deploy

Itxaka:
currently we dont show the name of the component being deployed
so add a var to pass the name so its clear to the user to what
component we are applying the tasks to.

### DIFF
--- a/playbooks/roles/deploy-osh/tasks/component-install.yml
+++ b/playbooks/roles/deploy-osh/tasks/component-install.yml
@@ -1,11 +1,11 @@
 ---
-- name: Generate SUSE defaults
+- name: Generate SUSE defaults for {{ _component_runcommand.split("-") | last | splitext | first }}
   loop: "{{ _component_overrides }}"
   template:
     src: "{{ item.suse_file }}.j2"
     dest: "/tmp/socok8s-susedefaults-{{ item.suse_file }}"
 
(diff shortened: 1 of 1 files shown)


jsuchome:
Nice idea, but it's always the same name as the one that is part of `suse_file` ... couldn't we somehow use just one variable for both?

Itxaka:
> Nice idea, but it's always the same name as the one that is part of suse_file ... couldn't we somehow use just one variable for both?

I did not want to obfuscate this anymore as its potentially confusing as to add more abstraction in there. I could always change it to: `suse_file: {{ _component_name }}.yaml`

*disclaimer: not even sure if that would work

jsuchome:
@Itxaka this is more DRY , but also more ugly :-( Any time to make it little bit nicer?

Itxaka:
> @Itxaka this is more DRY , but also more ugly :-( Any time to make it little bit nicer?

Either its simpler or we get an ugly component name. You can choose one :P

Itxaka:
> But fyi the ansible way would be | splitext | first
> No need to update the PR though.

what have you done, now I need to change it. I cant live with myself knowing there is a much nicer way of writing it and leaving it like that.

Itxaka:
last change, I promise

evrardjp:
Not tested.

PR #192, opened 2019-03-27
show patch subject on task

Itxaka:
just a fancy task name which included the patch subject

### DIFF
--- a/playbooks/roles/dev-patcher/tasks/patch-repos.yml
+++ b/playbooks/roles/dev-patcher/tasks/patch-repos.yml
@@ -13,7 +13,7 @@
   vars:
     pattern_to_replace: ")]}'"
 
-- name: Apply {{ patchnumber }}
+- name: Apply {{ patchnumber }} - {{ jsoncontent[0]['subject'] }}
   shell: |
     set -o errexit
     git fetch {{ jsoncontent[0]['revisions'][current_revision]['fetch']['anonymous http']['url'] }} {{ jsoncontent[0]['revisions'][current_revision]['fetch']['anonymous http']['ref'] }}


playbooks/roles/dev-patcher/tasks/patch-repos.yml
```diff
@@ -13,7 +13,7 @@
   vars:
     pattern_to_replace: ")]}'"
 
-- name: Apply {{ patchnumber }}
+- name: Apply {{ patchnumber }} - {{ jsoncontent[0].subject }}
```
evrardjp: ['subject'].

evrardjp: (Clarification: It's not an attribute of that dict, it's a key. It could work though, but I am not so sure how it will work in the future... We don't know what ansible will prepare us)

Itxaka: It works and it should work forever, or ansible is shooting their own foot with this, but I do agree that using the dict notation makes it future proff in case ansible goes crazy and decides to break every playbook in the world :P

evrardjp: FYI:

Interface changes: It's been the case in the past :)

I am asking this because, AFAIK, there is no documentation saying you could do this. In fact, since 2.1 I believe there is a willingness to use more native objects, and stop doing these abstractions "for userfriendliness".

Itxaka: There is docs saying that you can use this but advise to be careful with the built in functions[0]

I do find it strange that both methods are available and recommended. Should deprecate the dot notation one IMHO as dict notation just makes more sense coming from python

[0] https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#creating-valid-variable-names

PR #195, opened 2019-03-27
Use the upstream Rocky images for Glance

jsuchome:
Initially only in development mode, now that the images are uploaded, directly using it at all times.

### DIFF
--- a/playbooks/roles/deploy-osh/templates/glance.yaml.j2
+++ b/playbooks/roles/deploy-osh/templates/glance.yaml.j2
@@ -9,4 +9,14 @@ conf:
     glance_store:
       rbd_store_user: {{ ses_cluster_configuration['glance']['rbd_store_user'] }}
       rbd_store_pool: {{ ses_cluster_configuration['glance']['rbd_store_pool'] }}
-{{ lookup('template','files/common-images-overrides.yml') | from_yaml | to_nice_yaml(indent=2) }}
+images:
+  tags:
+    bootstrap: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+    db_init: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+    db_drop: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+    ks_user: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+    ks_service: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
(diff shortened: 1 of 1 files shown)


playbooks/roles/deploy-osh/templates/glance.yaml.j2
```diff
@@ -9,4 +9,16 @@ conf:
...
+  tags:
+    bootstrap: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+    db_init: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+    db_drop: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+    ks_user: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+    ks_service: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+    ks_endpoints: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+    glance_db_sync: "{{ suse_osh_registry_location }}/openstackhelm/glance:{{ suse_openstack_image_version }}"
+    glance_api: "{{ suse_osh_registry_location }}/openstackhelm/glance:{{ suse_openstack_image_version }}"
+    glance_registry: "{{ suse_osh_registry_location }}/openstackhelm/glance:{{ suse_openstack_image_version }}"
```
evrardjp: glance_storage_init still has to be done. I am adding this comment in the review so that I remember :) 

The rest of the images seem to be all converted (except official images, like on the previous changes for Keystone, etc.)

playbooks/roles/deploy-osh/templates/glance.yaml.j2
```diff
@@ -9,4 +9,16 @@ conf:
     glance_store:
       rbd_store_user: {{ ses_cluster_configuration['glance']['rbd_store_user'] }}
       rbd_store_pool: {{ ses_cluster_configuration['glance']['rbd_store_pool'] }}
-{{ lookup('template','files/common-images-overrides.yml') | from_yaml | to_nice_yaml(indent=2) }}
+{% if developer_mode %}
```
evrardjp: Do we still need developer_mode ?
I thought there was no patch to use for this?

playbooks/roles/deploy-osh/templates/glance.yaml.j2
jsuchome: `glance_storage_init` uses docker.io/port/ceph-config-helper:v1.10.3 , do we need to replace that?

playbooks/roles/deploy-osh/templates/glance.yaml.j2
jsuchome: Right, no patch. I'll remove the developer_mode check

PR #197, opened 2019-03-27
Changed run_rally_tests to run_tests and set default value to true

matthew-fuller:
This change renames the variable from 'run_rally_tests' to the more accurate 'run_tests', since this flag determines whether all defined helm tests run (not just rally). Also sets the default value to true, since this will likely be the desired case in production.


vars/common-vars.yml
```diff
@@ -57,5 +57,5 @@ ucp_keystone_admin_password: 'password123'
 openstack_keystone_admin_password: 'password123'
 ucp_shipyard_keystone_password: 'password123'
 
-# Flag to run pod rally tests (default false)
-run_rally_tests: false
+# Flag to run or disable all helm tests (default true)
+run_tests: true
```
evrardjp: This is not really an option we should expose in common vars. Common vars is not a graveyard for vars... It's the place were we put variables that are common to multiple roles.

Here I would say, with the fact it's not used by multiple roles (where is this variable used btw?) it should probably go into the role defaults.

PR #198, opened 2019-03-27
Add a connection timeout to the SES node reboot

jesusaurus:
During the SES deploy, I have been seeing this reboot hang for a long
time before failing the entire playbook. Adding a short connection
timeout triggers a reconnection attempt that resolves the task timeout.

### DIFF
--- a/playbooks/openstack-ses_aio_hostconfig.yml
+++ b/playbooks/openstack-ses_aio_hostconfig.yml
@@ -64,6 +64,7 @@
     - name: Reboot host
       reboot:
         reboot_timeout: 900
+        connect_timeout: 90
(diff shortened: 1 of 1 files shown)


evrardjp:
Better having it than not having it.

evrardjp:
Note: Not merging it until others have seen it.

jesusaurus:
I wish that we could use a backoff mechanism; but since we can't and this is for the connection between localhost and the undercloud hosts (probably the slowest connection), I agree 30 seconds might be a little too aggressive.

evrardjp:
We let it a day, I guess that's fair enough.

PR #200, opened 2019-03-28
Modifed to have timeout value as configurable for openstack and ucp.

rajdharwadkar:


rajdharwadkar:
> just to be consistent with other var names, can we call this osh_timeout or openstack_helm_timeout?

I went with namespaces that we have but I can change it to osh.

jhesketh:
Change lgtm, but I agree that we should possibly be more explicit with names.  There are lots of parts that could be timeouts, this is specifically the helm charts though so we should make that obvious imo.

PR #202, opened 2019-03-28
Cinder: Use openSUSE images at all times


evrardjp:
Now that images are published we can use them by default.


evrardjp:
I am still testing this :)

evrardjp:
Worked for me, and 2 votes, let's merge.

PR #203, opened 2019-03-28
Use Rocky images for Heat

jsuchome:


playbooks/roles/deploy-osh/templates/heat.yaml.j2
```diff
@@ -1,2 +1,17 @@
...
+    ks_service: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+    ks_endpoints: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+    heat_db_sync: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+    heat_api: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+    heat_cfn: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+    heat_cloudwatch: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+    heat_engine: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+    heat_engine_cleaner: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+    heat_purge_deleted: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+    rabbit_init: docker.io/rabbitmq:3.7-management
```
Itxaka: I mirth have missed this in another patch conversation but are we gonna be verbose here and put *all* needed images or we just gonna *only* put the override ones (the file seems to point at that) and left the others fallback into the default ones until we fix them?

There is already a good way of getting the images for every chart so Im not sure about this. Plus on the, already merged, keystone patch, I did not mention anything other than what we are overriding, so we might need to adapt that to mention *all* images.

jsuchome: I can remove that `rabbit_init` one. I would agree that we only should have the overrides here.

However it was already done differently in some other PR's

evrardjp: oh yes, rabbitmq is not required in here.
As far as I know we've listed all the images that we'll suse-ify.

evrardjp:
If it works for 2 ppl, let's continue.

PR #208, opened 2019-03-29
Rabbitmq Add images overrides


evrardjp:
We still don't want to fork "official images", but all the rest
should be overriden. Therefore, neither prometheus_rabbitmq_exporter
[1] nor rabbitmq itself is updated here

[1]: prometheus_rabbitmq_exporter: docker.io/kbudde/rabbitmq-exporter


evrardjp:
I am testing this locally.

PR #210, opened 2019-03-29
Aading back airship components image override

kantarun:
With recent changes, airship components image override was not added
in ucp deployment via airship mechanism.

Adding image override for deckhand, armada and shipyard components


evrardjp:
Not tested, but lgtm. Haven't checked if some parts are missing.

PR #212, opened 2019-04-01
Changed var name to run_tests, moved location to extravars, set default value to true

matthew-fuller:
This change allows users to customize the behavior of helm tests during a site deployment. Adding the key run_tests to env/extravars and setting it to false will disable all helm tests for all charts during the deployment (they can still be run individually using the helm CLI), and adding the key test_timeout will allow users to specify a custom timeout value that tiller will wait before terminating a test (default is 300s). Documentation on the use of these variables is provided in source/user/customize-testing-behavior.rst

### DIFF
--- a/site/soc/software/charts/osh/openstack-compute-kit/neutron.yaml
+++ b/site/soc/software/charts/osh/openstack-compute-kit/neutron.yaml
@@ -11,7 +11,7 @@ metadata:
       component: neutron
     actions:
       - method: merge
-        path: .values.pod
+        path: .
       - method: replace
         path: .values.labels.agent.l3
       - method: replace
@@ -36,7 +36,8 @@ data:
     labels:
       release_group: airship-neutron
   test:
-    enabled: {{ run_rally_tests }}
+    enabled: {{ run_tests | default('true') }}
+    timeout: {{ test_timeout | default(300) }}
   values:
     pod:
       replicas:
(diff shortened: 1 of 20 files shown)


examples/workdir/env/extravars
```diff
@@ -19,6 +19,10 @@ redeploy_osh_only: false
 #suse_osh_registry_location: "docker.io"
 #suse_osh_image_version: "pike"
 
+# All helm tests defined for each chart are run by default during deployment.
```
evrardjp: Instead of providing those defaults in a file, we should use the documentation.

matthew-fuller: Do you mean move the explanation of the var out of this file and into the documentation? Or do you mean pull everything out of this file and simply add documentation to explain that users can add a var called 'run_tests' and set it to false if they wish to disable them?

evrardjp: The latter. I think we should simply give, as example, the bare minimum list of required extra variables. All the optional ones should be documented.

matthew-fuller: Done. I pulled everything out of the extravars example file and added a file in source/user documenting how to disable tests or customize the default test timeout value.

site/soc/software/charts/osh/openstack-compute-kit/neutron.yaml
```diff
@@ -36,7 +36,8 @@ data:
     labels:
       release_group: airship-neutron
   test:
-    enabled: {{ run_rally_tests }}
```
jgu17: do we also need to add merge at root level or for the new test element?

PR #213, opened 2019-04-02
SCRD-8345 Use random passwod in soc airship deployment

jgu17:
Change hard coded default passwords to generated random passwords for
all deployed airship and openstack services.

### DIFF
--- a/playbooks/roles/airship-deploy-ucp/defaults/main.yml
+++ b/playbooks/roles/airship-deploy-ucp/defaults/main.yml
@@ -1,4 +1,4 @@
 ---
 scale_profile: "{{ lookup('env','scale_profile') | default('minimal') }}"
-airship_pegleg_image: "{{ suse_airship_registry_location }}/airshipit/pegleg:{{suse_airship_components_image_tag}}"
+airship_pegleg_image: "{{ suse_airship_registry_location }}/airshipit/pegleg:{{ suse_airship_components_image_tag }}"
 suse_osh_image_version: "pike"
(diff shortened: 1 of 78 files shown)


playbooks/roles/airship-deploy-ucp/defaults/main.yml
```diff
@@ -1,4 +1,4 @@
 ---
 scale_profile: "{{ lookup('env','scale_profile') | default('minimal') }}"
-airship_pegleg_image: "{{ suse_airship_registry_location }}/airshipit/pegleg:{{suse_airship_components_image_tag}}"
+airship_pegleg_image: "{{ suse_airship_registry_location }}/airshipit/pegleg:{{ suse_airship_components_image_tag }}"
```
evrardjp: This should be in a different commit.

vars/common-vars.yml
```diff
@@ -53,16 +53,14 @@ suse_airship_components_image_tag: "{%- if suse_airship_build_local_images or su
...
 
+# where the generated secrets are stored
+secrets_location: "{{ socok8s_workspace }}/secrets"
+
 #Set timeout for ucp deployment
 ucp_deploy_timeout: 1800
 
 #Set timeout for osh deployment
-openstack_helm_deploy_timeout: 3000
+openstack_helm_deploy_timeout: 3000
```
evrardjp: This should be in a different commit.

evrardjp:
Note: I would prefer if we removed the non-password relevant changes, but I won't block this patch for it.

evrardjp:
Also, SCRD-8345 is an internal reference, and this is a public project, we should avoid that.

PR #215, opened 2019-04-02
update horizon overrides

Itxaka:
as the multi-os spec has landed upstream, the overrides have to be
changed to be in sync.

Mainly the change is that the overrides for apache2 need to be
under conf -> software -> apache2 now

### DIFF
--- a/playbooks/roles/deploy-osh/templates/horizon.yaml.j2
+++ b/playbooks/roles/deploy-osh/templates/horizon.yaml.j2
@@ -1,14 +1,14 @@
 ---
-{% if developer_mode %}
 conf:
-  apache2:
-    binary: apache2ctl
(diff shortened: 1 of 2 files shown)


playbooks/roles/deploy-osh/templates/horizon.yaml.j2
```diff
@@ -1,14 +1,15 @@
 ---
 {% if developer_mode %}
```
evrardjp: you can remove if developer_mode nowadays, and remove the upstream patch from the patch list.

Itxaka: yay!

evrardjp:
Tested and working in my env. (Sorry for the late answer).

PR #218, opened 2019-04-03
Consolidate ANSIBLE_RUNNER_DIR setup

jesusaurus:
There is logic in both install_ansible and run_ansible to check if
ANSIBLE_RUNNER_DIR is set and to some create initial subdirectories.
Since install_ansible will be run first, that logic will set the
variable and create the directory, preventing the logic in run_ansible
from ever running.

Consolidate code from run_ansible into install_ansible so that we have a
single code location for setting a default ANSIBLE_RUNNER_DIR value and
creating the directory if needed.

### DIFF
--- a/script_library/run-ansible.sh
+++ b/script_library/run-ansible.sh
@@ -4,21 +4,6 @@
 
 function run_ansible(){
     set -x
-    # ansible-runner default locations
-    if [[ -z ${ANSIBLE_RUNNER_DIR+x} ]]; then
-        echo "ANSIBLE_RUNNER_DIR env var is not set, defaulting to '~/suse-socok8s-deploy'"
-        export ANSIBLE_RUNNER_DIR="${HOME}/suse-socok8s-deploy"
-    fi
-
-    extravarsfile=${ANSIBLE_RUNNER_DIR}/env/extravars
-    inventorydir=${ANSIBLE_RUNNER_DIR}/inventory/
-
-    # This creates a structure that's similar to ansible-runner tool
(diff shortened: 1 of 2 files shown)


script_library/run-ansible.sh
```diff
@@ -4,21 +4,6 @@
 
 function run_ansible(){
     set -x
-    # ansible-runner default locations
```
evrardjp: I think we should keep this, as the bootstrap and the run_ansible functions can be dissociated. When not installing, ANSIBLE_RUNNER_DIR won't be set.

jesusaurus: I don't understand the use case you mean. Is there a case where we use 'script_library/run-ansible.sh' directly instead of being called by 'run.sh'? I thought it was only used by 'run.sh' which always checks ANSIBLE_RUNNER_DIR for a usable ansible binary.

evrardjp: 1) yes for debugging purposes, one might want to source run-ansible.
2) install_ansible function is only sourced when install_ansible needs to happen. When things don't need to happen (ansible already installed), ANSIBLE_RUNNER_DIR won't be set.

evrardjp: If you want, I ahve a different approach in https://github.com/evrardjp/socok8s/commits/scrd-4754-part2 , and it should solve your issue.

evrardjp:
I am not sure this is still required, could you check if the current behaviour is better for you ?

jesusaurus:
I was able to test your change yesterday, and it does fix my issue, I'll close this PR

PR #219, opened 2019-04-03
Use nova image for novncproxy_assets

jhesketh:
Upstream uses kolla/ubuntu-source-nova-novncproxy:ocata which as best as
I can tell is an image with openstack-nova-novncproxy which in turn is
just a packaged version of the nova cli entrypoint nova-novncproxy.

Thus the standard openstackhelm/nova image should be sufficient. Using
that image also reduces the number of images required to be pulled in.

This deploys fine, but I have not tested the actual novnc in horizon.

[0] https://github.com/openstack/kolla/blob/master/docker/nova/nova-novncproxy/Dockerfile.j2


Itxaka:
tested and works, novnc console appears correct with our image :)

evrardjp:
> I have not been able to properly test the spice image due to upstream issues but when hacking it around, spice worked like the nonvnc

Could you clarify what's the problem with upstream issues, and the required hacks?

Itxaka:
> Could you clarify what's the problem with upstream issues, and the required hacks?

Spice deployment will not be launched due to permission issues (pods are using `hostnetwrok: true` but service account appears to not have permissions for launching privileged containers.

removing the hostnetwork var makes the pod launch perfectly

evrardjp:
> > Could you clarify what's the problem with upstream issues, and the required hacks?
> 
> Spice deployment will not be launched due to permission issues (pods are using `hostnetwrok: true` but service account appears to not have permissions for launching privileged containers.
> 
> removing the hostnetwork var makes the pod launch perfectly

It's good I hope we'll remember that.

PR #221, opened 2019-04-03
changed the deployer hostname

rajdharwadkar:


Itxaka:
Not very happy with the commit explanation (doesnt say why the change) but should be ok

PR #224, opened 2019-04-04
caasp: add extra configs for apparmor profiles

Itxaka:
Some of our containers may need extra configs for apparmor.
For example, neutron-dhcp launches dnsmasq, which needs access
to the neutron config dir and the /var/lib/neutron/dhcp for leases.

We need a way of extending apparmor profiles with custom directives
for the proper execution of our openstack containers.

This patch allows us to add extra directives for apparmor profiles

### DIFF
--- a/playbooks/roles/setup-caasp-workers/handlers/main.yml
+++ b/playbooks/roles/setup-caasp-workers/handlers/main.yml
@@ -1,0 +1,10 @@
+---
+# ansible-lint: we need to reload the profiel and there is no ansible modules for that yet
+- name: Reload apparmor profile
+  shell: "apparmor_parser -r /etc/apparmor.d/{{ item.item }}"
+  when: item.changed
+  loop: "{{ apparmor_copy.results | flatten(levels=1) }}"
+  loop_control:
+    label: "{{ item.item }}"
+  tags:
+    - skip_ansible_lint
--- a/playbooks/roles/setup-caasp-workers/tasks/apparmor-profiles.yml
+++ b/playbooks/roles/setup-caasp-workers/tasks/apparmor-profiles.yml
@@ -1,0 +1,8 @@
+---
+- name: Copy openstack apparmor profile into the workers
+  copy:
+    src: "{{ item }}"
+    dest: "/etc/apparmor.d/local/{{ item }}"
+  loop: "{{ apparmor_profiles }}"
+  register: apparmor_copy
+  notify: Reload apparmor profile
--- a/playbooks/roles/setup-caasp-workers/defaults/main.yml
+++ b/playbooks/roles/setup-caasp-workers/defaults/main.yml
@@ -1,2 +1,4 @@
 ---
 caasp_reboot_timeout: 300
+apparmor_profiles:
+  - usr.sbin.dnsmasq
--- a/playbooks/roles/setup-caasp-workers/files/usr.sbin.dnsmasq
+++ b/playbooks/roles/setup-caasp-workers/files/usr.sbin.dnsmasq
@@ -1,0 +1,5 @@
+# Site-specific additions and overrides for 'usr.sbin.dnsmasq'
+# neutron containers need the following access to launch dnsmasq
+  /etc/neutron/dnsmasq.conf r,
+  /var/lib/neutron/dhcp/ rw,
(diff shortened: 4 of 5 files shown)


playbooks/roles/setup-caasp-workers/tasks/apparmor-profiles.yml
```diff
@@ -0,0 +1,22 @@
+---
+- name: Apparmor config
+  block:
+    # need to use shell because mount is not able to remount without passing a lot of params that we may not know
+    - name: Set the root disk into rw
+      shell: "mount -o remount, rw /"
```
rhafer: Why is this needed at all? Isn't /etc an overlay mounted rw by default?

playbooks/roles/setup-caasp-workers/tasks/apparmor-profiles.yml
```diff
@@ -0,0 +1,12 @@
+---
+- name: Copy openstack apparmor profile into the workers
+  copy:
+    src: "{{ item }}"
+    dest: "/etc/apparmor.d/local/{{ item }}"
+  with_items: "{{ apparmor_profiles }}"
```
evrardjp: Can you use `loop` instead of `with_items` ?

playbooks/roles/setup-caasp-workers/tasks/apparmor-profiles.yml
```diff
@@ -0,0 +1,12 @@
...
+  register: apparmor_copy
```
evrardjp: Can you use

`notify:`

and handlers instead?

playbooks/roles/setup-caasp-workers/tasks/apparmor-profiles.yml
Itxaka: done

playbooks/roles/setup-caasp-workers/tasks/apparmor-profiles.yml
Itxaka: nope, dynamic handlers do not work on ansible, nor does handlers with parameters, which kind of sucks

evrardjp: What do you mean by dynamic handlers do not work on ansible?

Itxaka: ok I did some hack around to manage those dynamic handlers, but IMO this is ugly AF.

Itxaka: > What do you mean by dynamic handlers do not work on ansible?

something like:

```
- name: restart service {{ item }}
  service:
    name: {{ item }}
    state: restarted

```

so then you can just notify with `restart service WHATEVER` and get it done dynamically.

My hack kind of works but its really ugly, I think handlers should work more like this.

evrardjp: Can be done with a chain of handlers and a set fact, but ok.

playbooks/roles/setup-caasp-workers/handlers/main.yml
```diff
@@ -0,0 +1,10 @@
+---
+# ansible-lint: we need to reload the profiel and there is no ansible modules for that yet
+- name: Reload apparmor profile
+  shell: "apparmor_parser -r /etc/apparmor.d/{{ item.item }}"
+  when: item.changed
+  loop: "{{ apparmor_copy.results | flatten(levels=1) }}"
+  loop_control:
+    label: "{{ item.item }}"
```
evrardjp: nit: Not sure why you would want to set the label to this, it doesn't improve readability.

playbooks/roles/setup-caasp-workers/tasks/apparmor-profiles.yml
evrardjp: That was a nit: anyway.

playbooks/roles/setup-caasp-workers/handlers/main.yml
evrardjp: I misread I thought this was loop_var.

PR #226, opened 2019-04-04
Adapt WORKDIR

evrardjp:

### DIFF
--- a/playbooks/openstack-caasp_instance.yml
+++ b/playbooks/openstack-caasp_instance.yml
@@ -2,9 +2,7 @@
 - hosts: localhost
   gather_facts: no
   connection: local
-  roles:
-    - ansible-config_template
-  post_tasks:
+  tasks:
     - name: Load generic vars
       include_vars: "{{ playbook_dir }}/../vars/common-vars.yml"
     - name: Load openstack vars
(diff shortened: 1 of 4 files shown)


evrardjp:
Requires this: https://github.com/SUSE-Cloud/automation/pull/3317

When this will be done, step 2 of the parallelisation will be done.

playbooks/openstack-caasp_instance.yml
```diff
@@ -163,27 +152,60 @@
...
-                hosts: "{{ workerinventory }}"
+                hosts:
+              {% for node in workers %}
+                  {{ node.Name }}:
+                    ansible_host: {{ node.Networks.split(',')[-1].replace(' ','') }}
+              {% endfor %}
+
               airship-openstack-control-workers:
-                hosts: "{{ controlinventory }}"
+                hosts: &firsttwoworkers
```
Itxaka: is this correct or a typo? The others have `*` instead (which Im ignorant on what is doing to be honest)

playbooks/openstack-caasp_instance.yml
```diff
@@ -163,27 +152,60 @@
...
-          config_template:
-            src: "{{ socok8s_workspace }}/inventory/skeleton-inventory.yml"
+          copy:
+            content: "{{ caasp_overrides }}"
             dest: "{{ socok8s_workspace }}/inventory/caasp.yml"
-            config_type: yaml
-            config_overrides: "{{ caasp_overrides }}"
           vars:
-            caasp_overrides:
+            caasp_overrides: |
```
Itxaka: wouldnt this be better stored as a template instead?

evrardjp: That will be done in the step 3, when things will move to a role for DRY.

playbooks/openstack-caasp_instance.yml
evrardjp: `&` is for storing a value, `*` is to load it.

Itxaka: :+1:

playbooks/openstack-caasp_instance.yml
Itxaka: :+1:

evrardjp:
Answered.

playbooks/openstack-caasp_instance.yml
```diff
@@ -193,20 +215,10 @@
...
-            state: started
-            delay: 10
-            timeout: 180
-          loop: "{{ groups['caasp-admin'] + groups['caasp-masters'] + groups['caasp-workers'] }}"
-
         - name: Get pubkey, and add it to known hosts
           command: "ssh-keyscan -H {{ hostvars[item]['ansible_host'] | quote }}"
           register: _caaspsshkeys
-          retries: 10
+          retries: 25
```
jhesketh: Why have we gotten rid of the wait for ssh?

evrardjp: because it has no purpose?

The only reason was to wait_for ssh to be up before scanning keys. But since then we have a `until:` condition that ensures the same things.

PR #230, opened 2019-04-05
Remove static names for CI runs

evrardjp:
Ensure the servernames don't collapse into each other
when running multiple jobs in parallel.


evrardjp:
Will run a CI test with this.

rhafer:
Would you mind squasing that last two commits? Otherwise this looks good I think.

evrardjp:
I think these are different, but can do.

evrardjp:
@rhafer WOW, we almost did a boo-boo here.

If we keep this last commit (which is convenient, I agree), we'll not be able to split the creation and the deletion in different jobs, should we want to split the creation of a long running caasp/ses cluster and its deletion.

evrardjp:
We've discussed this in our meeting today.

Long story short:
1) We could keep this, and wire the lookup('env','BUILD_NUMBER') differently for long running caasp cluster
2) Long running caasp cluster is not a current reality.
3) Using a unique prefix for everything is clearer, and we should derive the ANSIBLE_RUNNER_DIR from that prefix.

evrardjp:
I will be creating a different variable, like we suggested in the meeting, "environment name", to be unique, and for which the CI automatically wires the BUILD_NUMBER.

All the playbooks won't be reading the folder of the ANSIBLE_RUNNER_DIR to define the servername, but instead rely on this "environment name".

When splitting will happen down the line, we can just set "Environment name" for the right wiring to happen.

evrardjp:
Abandonning this, will recreate a new PR with the new changes.

evrardjp:
Superseded by #241 .

PR #231, opened 2019-04-05
use suse images for ceph_config_helper under cinder

Itxaka:
tested and working


evrardjp:
Will test. Looks nice if it works out of the box. I am worried our testing might not cover things properly.

PR #232, opened 2019-04-05
Remove cruft in caasp stack creation


evrardjp:
To run in CI, the heat stack name needs to be dynamically generated.
The current code offers this, but relies on a temp file, which is
ugly and error prone.

Instead this code relies on idempotency using the random filter.

### DIFF
--- a/playbooks/openstack-caasp_instance.yml
+++ b/playbooks/openstack-caasp_instance.yml
@@ -2,34 +2,28 @@
 - hosts: localhost
   gather_facts: no
   connection: local
+  vars:
+    stackname_suffix: caasp
   tasks:
     - name: Load generic vars
-      include_vars: "{{ playbook_dir }}/../vars/common-vars.yml"
-    - name: Load openstack vars
-      include_vars: "{{ playbook_dir }}/../vars/deploy-on-openstack.yml"
+      include_vars: "{{ item }}"
+      loop:
+        - "{{ playbook_dir }}/../vars/common-vars.yml"
+        - "{{ playbook_dir }}/../vars/deploy-on-openstack.yml"
 
-    - name: Define stackname to create
+    # TODO(evrardjp): Merge this logic with the handle-nodes-on-openstack logic
+    # when the whole stack is ready to move to that role.
+    - name: Define stackname based on runtime
       set_fact:
-        stackname: "{{ deploy_on_openstack_prefix }}-{{ lookup('env', 'BUILD_NUMBER') | default(65000 | random, true) }}"
+        stackname: "{{ deploy_on_openstack_keypairname }}-{{ 10000 | random(seed=socok8s_workspace) }}-{{ stackname_suffix }}"
 
-    - name: Check if a stack already exists
-      stat:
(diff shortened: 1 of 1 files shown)


playbooks/openstack-caasp_instance.yml
```diff
@@ -2,34 +2,28 @@
...
+    # when the whole stack is ready to move to that role.
+    - name: Define stackname based on runtime
       set_fact:
-        stackname: "{{ deploy_on_openstack_prefix }}-{{ lookup('env', 'BUILD_NUMBER') | default(65000 | random, true) }}"
-
-    - name: Check if a stack already exists
-      stat:
-        path: "{{ deploy_on_openstack_caasp_stacknamefile }}"
-      register: _stackfile
+        stackname: "{{ deploy_on_openstack_keypairname }}-{{ 10000 | random(seed=socok8s_workspace) }}-{{ stackname_suffix }}"
```
rhafer: Hm, why not stay with BUILD_NUMBER (IIRC that's set by jenkins?) by default? I might be missing something but I think that would make associating the stacks with ci jobs just straight forward. (E.g. for debugging purposes)

evrardjp: 1) to rely on the same kind of code everywhere.
2) In CI the ANSIBLE_RUNNER_DIR is now containing BUILD_NUMBER (or `build ID`, I can't remember). That env var is read in ansible as `socok8s_workdir`
3) This allows devs like you and I to simply set the ANSIBLE_RUNNER_DIR to a folder of your choice and run this. So if you want to run two things in parallel, just define another ANSIBLE_RUNNER_DIR location.

evrardjp: See also #230 .

rhafer: > * to rely on the same kind of code everywhere.
> 
> * In CI the ANSIBLE_RUNNER_DIR is now containing BUILD_NUMBER (or `build ID`, I can't remember). That env var is read in ansible as `socok8s_workdir`
> 
> * This allows devs like you and I to simply set the ANSIBLE_RUNNER_DIR to a folder of your choice and run this. So if you want to run two things in parallel, just define another ANSIBLE_RUNNER_DIR location.

Sure. That's all nice and shiny. I just wonder why we need to have random id attached to the stacknames for CI jobs (not for manually started deployments), when those ci jobs already come with a perfectly fine unique identifier that could be used in the job name, that's all.

In the end I don't care much. I just thought having a direct mapping between the job id and the stack name is somewhat convenient.

evrardjp: we want to reuse the ID of the jobs, that's correct.
We also want to run the same code in CI as in devs environments.

ANSIBLE_RUNNER_DIR is basically that, as it is derived from that value.

evrardjp: I am all ears on an alternative approach.

evrardjp: That will need updating based on #230.

PR #234, opened 2019-04-05
Remove cloud variable


evrardjp:
There is no need to expose `cloud`, as we are expecting the
developer to set OS_CLOUD in run.sh, and all those plays are
working on localhost.


evrardjp:
Only merge this when the first two commits are merged in PR #230 .

evrardjp:
This can be done in the reimplementation of #230.

evrardjp:
This was tested and working in my environment.

toabctl:
@evrardjp is this still needed?

evrardjp:
> @evrardjp is this still needed?

Needed no, useful yes. Will re-create a new one, because I am not into conflict resolution mode.

PR #235, opened 2019-04-05
Remove PREFIX variable


evrardjp:
Some vars are derived from PREFIX, some are not. This simplifies
by making everything explicit. It removes PREFIX from the code
requirements.


evrardjp:
This requires:
1) A change in SUSE-Cloud/automation
2) The change in the heat stack #232 .
3) A rebase when #230 will merge.

toabctl:
As discussed on mumble, I would prefer to use the `PREFIX` variable everywhere to define the needed resources (like network, servers, volumes, ...)

evrardjp:
We'll also need to define keypair by prefix.

I believe this PR should be revised with the input given in our meeting https://github.com/SUSE-Cloud/socok8s/pull/230#issuecomment-480739616 .

I will go back to the drawing board, giving the variable a different name than "PREFIX" (like suggested in the meeting: environment_name or something)

evrardjp:
Abandonning this.

PR #236, opened 2019-04-05
use suse images for ceph_config_helper under nova

Itxaka:
tested and working


evrardjp:
How can I test this?

evrardjp:
> curious -- is this a new pod in Rocky?

I don't think so.

evrardjp:
I don't know what this does, so I don't know how to test it, therefore I can't vote.

Itxaka:
> How can I test this?

Mainly you clean the whole deployment and redeploy with this, and check the output of the containers. They must have the same output (i.e. do the same actions) as the default ones.

Crude, but it works.

> curious -- is this a new pod in Rocky?

Nope, its a container part of the nova chart which creates the ceph keyring as a secret inside k8s, and makes sure the user, keys, pools and so on are created for the storage backend (rdb in our case)

PR #237, opened 2019-04-05
Add missing routing for heat and novncproxy in /etc/host

jgu17:
Add missing routing for heat and novncproxy in /etc/host.


jgu17:
LGTM

toabctl:
Can we try to have meaningful commit messages? It's really hard to follow if there is no description **why** something is useful...

PR #238, opened 2019-04-05
Registry server setup needs to run on all worker nodes and deployer.

rajdharwadkar:
As inventory file has changed and looks like we missed this to change.


rajdharwadkar:
> the airship-workers have been renamed to "caasp-workers" in the inventory file

Yeah looks like I had old inventory file and made changes accordingly. fixed it.

PR #239, opened 2019-04-05
Networking documentation

sgrasley:
This adds network documentation for configuring multiple networks
in caasp and consuming them in airship allowing network function
seperation.


evrardjp:
I know it's draft, but I thought it would be good to clarify the purpose of each section of the documentation.

PR #240, opened 2019-04-05
dev-patcher: set git user.name and user.email if needed

jesusaurus:
If either user.name or user.email is not set, then the upcoming
cherry-pick in patch-repos.yml will fail. Previously, this was a manual
configuration step. This change sets them to sane service-user values
in the event that they have not already been configured.

### DIFF
--- a/playbooks/roles/dev-patcher/tasks/main.yml
+++ b/playbooks/roles/dev-patcher/tasks/main.yml
@@ -27,6 +27,25 @@
   register: _gitclone
   until: _gitclone is succeeded
 
+- name: Read git config settings
+  git_config:
+    list_all: yes
+  register: git_config
+
+- name: Set user.name if needed
+  git_config:
+    name: user.name
+    scope: global
+    value: "Dev Patcher"
+  when: "'user.name' not in git_config['config_values'].keys()"
+
+- name: Set user.email if needed
+  git_config:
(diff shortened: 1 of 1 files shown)


jhesketh:
Looks like ansible-lint is unhappy. Instead of using || in bash, could check if the user.name is set in ansible's `when` or something.

evrardjp:
> Looks like ansible-lint is unhappy. Instead of using || in bash, could check if the user.name is set in ansible's `when` or something.

Yes indeed, I suppose we could use git_config module to first list all things, and define conditionally using `git_config` with a `when` condition.

playbooks/roles/dev-patcher/tasks/main.yml
```diff
@@ -27,6 +27,25 @@
...
+    scope: global
+    value: "Dev Patcher"
+  when: "'user.name' not in git_config['config_values'].keys()"
+
+- name: Set user.email if needed
+  git_config:
+    name: user.email
+    scope: global
+    value: "dev-patcher@{{ ansible_hostname }}"
+  when: "'user.name' not in git_config['config_values'].keys()"
```
evrardjp: user.email ?

jesusaurus: whoops

PR #241, opened 2019-04-08
Use unique servernames


evrardjp:
Currently servernames are non-unique, due to the fact they rely
on PREFIX, which is not unique (only workspace is currently unique).

This is a problem if you are running more than a single environment,
because all the nodes of a same type (for example deployer/ses) will
share the same name if they are on the same network.

This is solved by making sure "PREFIX" becomes a unique variable,
representing each environment.

Because the term "PREFIX" was already used for convenience in terms
of non unique networking, I removed all the reference of derivation
of network from the PREFIX variable.

### DIFF
--- a/doc/source/deploy-guide/prepare-localhost.rst
+++ b/doc/source/deploy-guide/prepare-localhost.rst
@@ -164,33 +164,32 @@ OpenStack CLI’s ``openstack keypair create``) for accessing the
 instances created. Remember the name of this keypair (which appears as
 ``foctodoodle-key`` in the example below)
 
-Set this for all the following scripts:
+Set this for **all** the following scripts in a deployment:
 
 .. code-block:: console
 
-   export OS_CLOUD=engcloud # Assuming you followed the example for clouds.yaml
-   # Your username plus whatever else you would like, will be used for naming
-   # objects you create in the cloud
-   export PREFIX=foctodoodle
+   export OS_CLOUD=engcloud
+   # 'engcloud' is the name in the `clouds.yaml`,
    # Set the name of the keypair you created
    export KEYNAME=foctodoodle-key
-   # Set the name of the subnet you created
-   export INTERNAL_SUBNET=${PREFIX}-subnet
+   # Set the name of the network you will use in the deployment
+   export INTERNAL_NETWORK=foctodoodle-net
+   # Set the name of the subnet you will use in the deployment
+   export INTERNAL_SUBNET=foctodoodle-subnet
+   # Set the name of the floating ip network you will use in the deployment
+   export EXTERNAL_NETWORK=floating
 
-Now create a network, a subnet, a router and a connection to the
-floating network (you only have to do it once):
+If you haven't created the internal networks/subnets and appropriate routers
+already, do it now (you only have to do it once):
 
 .. code-block:: console
 
-   openstack network create ${PREFIX}-net
-   openstack subnet create --network ${PREFIX}-net --subnet-range 192.168.100.0/24 ${PREFIX}-subnet
-   openstack router create ${PREFIX}-router
-   openstack router set --external-gateway floating ${PREFIX}-router
-   openstack router add subnet ${PREFIX}-router ${PREFIX}-subnet
+   openstack network create ${INTERNAL_NETWORK}
+   openstack subnet create --network ${INTERNAL_NETWORK} --subnet-range 192.168.100.0/24 ${INTERNAL_SUBNET}
+   openstack router create ${INTERNAL_NETWORK}-router
+   openstack router set --external-gateway floating ${INTERNAL_NETWORK}-router
+   openstack router add subnet ${INTERNAL_NETWORK}-router ${INTERNAL_SUBNET}
 
-Prior to executing scripts, be aware that you may need to do some
-cleanup prior to retrying scripts or playbooks should they fail. In some
-steps a teardown script is provided to clean up any created resources.
 Reconfirming that you’ve done all the previous steps to set up now will
 save you some time later.
 
--- a/playbooks/openstack-osh_instance.yml
+++ b/playbooks/openstack-osh_instance.yml
@@ -11,7 +11,7 @@
   roles:
     - role: handle-nodes-on-openstack
       vars:
-        servername: "{{ deploy_on_openstack_oshnode_server_name }}"
+        servername_suffix: "deployer"
         flavor: "{{ deploy_on_openstack_oshnode_flavor }}"
         image: "{{ deploy_on_openstack_oshnode_image }}"
         network: "{{ deploy_on_openstack_internal_network }}"
--- a/playbooks/openstack-osh_instance.yml
+++ b/playbooks/openstack-osh_instance.yml
@@ -2,109 +2,21 @@
 - hosts: localhost
   gather_facts: no
   connection: local
-  tasks:
+  pre_tasks:
     - name: Load generic vars
-      include_vars: "{{ playbook_dir }}/../vars/common-vars.yml"
-    - name: Load openstack vars
-      include_vars: "{{ playbook_dir }}/../vars/deploy-on-openstack.yml"
-
-    - name: Handle creation
-      when:
-        - not ((osh_node_delete | default('false'))| bool)
-      block:
-        - name: Create volume
-          os_volume:
-            cloud: "{{ deploy_on_openstack_cloudname }}"
-            display_name: "{{ deploy_on_openstack_oshnode_volume_name }}"
-            size: 80
-            state: "present"
-        - name: Create server
-          os_server:
-            cloud: "{{ deploy_on_openstack_cloudname }}"
-            flavor: "{{ deploy_on_openstack_oshnode_flavor }}"
-            image: "{{ deploy_on_openstack_oshnode_image }}"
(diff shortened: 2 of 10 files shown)


playbooks/openstack-osh_instance.yml
```diff
@@ -2,109 +2,21 @@
...
+      loop:
+        - "{{ playbook_dir }}/../vars/common-vars.yml"
+        - "{{ playbook_dir }}/../vars/deploy-on-openstack.yml"
+  roles:
+    - role: handle-nodes-on-openstack
+      vars:
+        servername_suffix: "deployer"
+        flavor: "{{ deploy_on_openstack_oshnode_flavor }}"
+        image: "{{ deploy_on_openstack_oshnode_image }}"
+        network: "{{ deploy_on_openstack_internal_network }}"
```
evrardjp: internalnetwork -- tiny mistake here.

playbooks/openstack-ses_aio_instance.yml
```diff
@@ -2,107 +2,20 @@
...
+      loop:
+        - "{{ playbook_dir }}/../vars/common-vars.yml"
+        - "{{ playbook_dir }}/../vars/deploy-on-openstack.yml"
+  roles:
+    - role: handle-nodes-on-openstack
+      vars:
+        servername_suffix: "ses"
+        flavor: "{{ deploy_on_openstack_sesnode_flavor }}"
+        image: "{{ deploy_on_openstack_sesnode_image }}"
+        network: "{{ deploy_on_openstack_internal_network }}"
```
evrardjp: internalnetwork

playbooks/roles/handle-nodes-on-openstack/tasks/create_on_openstack.yml
```diff
@@ -0,0 +1,59 @@
...
+    state: "present"
+
+- name: Create server
+  os_server:
+    cloud: "{{ deploy_on_openstack_cloudname }}"
+    flavor: "{{ flavor }}"
+    image: "{{ image }}"
+    key_name: "{{ deploy_on_openstack_keypairname }}"
+    name: "{{ servername }}"
+    network: "{{ internalnetwork }}"
```
evrardjp: I need to edit this or the playbooks

vars/deploy-on-openstack.yml
```diff
@@ -1,11 +1,13 @@
...
 deploy_on_openstack_cloudname: "{{ lookup('env','OS_CLOUD') | default('engcloud-cloud-ci', true) }}"
-deploy_on_openstack_prefix: "{{ lookup('env','PREFIX') }}" #PREFIX is set at all times (ensured in script library)
-deploy_on_openstack_external_network: "floating"
-deploy_on_openstack_internal_network: "{{ deploy_on_openstack_prefix }}-net"
-deploy_on_openstack_internal_subnet: "{{ deploy_on_openstack_prefix }}-subnet"
+# PREFIX must be unique per environment.
+# If you want to deploy two environments on the same network,
+# make sure PREFIX is defined for your environment before running run.sh
+deploy_on_openstack_prefix: "{{ lookup('env','PREFIX') | default('socok8s', true) }}"
+deploy_on_openstack_external_network: "lookup('env', 'EXTERNAL_NETWORK') | default('floating', true) }}"
```
evrardjp: missing {{

jhesketh:
Seems reasonable to me (minus the existing comments)

playbooks/roles/handle-nodes-on-openstack/tasks/create_on_openstack.yml
```diff
@@ -0,0 +1,59 @@
+---
+- name: Create volume
+  os_volume:
+    cloud: "{{ deploy_on_openstack_cloudname }}"
+    display_name: "{{ servername }}-vol"
+    size: 80
+    state: "present"
+
+- name: Create server
+  os_server:
```
toabctl: While changing this - what about switching to a heat stack? That would make naming, grouping and cleanup easier imo

evrardjp: I love this idea, but I would prefer doing this in a later patchset.

evrardjp:
Ready

vars/deploy-on-openstack.yml
```diff
@@ -1,15 +1,16 @@
...
-deploy_on_openstack_external_network: "floating"
-deploy_on_openstack_internal_network: "{{ deploy_on_openstack_prefix }}-net"
-deploy_on_openstack_internal_subnet: "{{ deploy_on_openstack_prefix }}-subnet"
+# PREFIX must be unique per environment.
+# If you want to deploy two environments on the same network,
+# make sure PREFIX is defined for your environment before running run.sh
+deploy_on_openstack_prefix: "{{ lookup('env','PREFIX') | default('socok8s', true) }}"
+deploy_on_openstack_external_network: "{{ lookup('env', 'EXTERNAL_NETWORK') | default('floating', true) }}"
+deploy_on_openstack_internal_network: "{{ lookup('env','INTERNAL_NETWORK') | default('ccpci-net', true) }}"
+deploy_on_openstack_internal_subnet: "{{ lookup('env', 'INTERNAL_SUBNET') | default('ccpci-subnet', true) }}"
```
jhesketh: I wonder if we shouldn't give defaults to these at the risk of developers deploying over each other. Perhaps we should raise an error if the env is unset?

evrardjp: I kept the current behaviour, but I totally think we should remove those defaults in a later commit.

rhafer:
@evrardjp Looks good to me in general. I'd like to give it a "quick" test before approving tough. And there is one thing I don't quite understand:

> Because the term "PREFIX" was already used for convenience in terms of non unique networking, I removed all the reference of derivation of network from the PREFIX variable.

What's the problem with deriving the network name from PREFIX as well? Do you want to allow multiple (parallel) deployments using the same network?

toabctl:
> What's the problem with deriving the network name from PREFIX as well? Do you want to allow multiple (parallel) deployments using the same network?

I don't see a problem. I proposed already https://github.com/SUSE-Cloud/socok8s/pull/166 to automatically setup the network based on the PREFIX so you can have multiple deployments in parallel...

evrardjp:
> > What's the problem with deriving the network name from PREFIX as well? Do you want to allow multiple (parallel) deployments using the same network?
> 
> I don't see a problem. I proposed already #166 to automatically setup the network based on the PREFIX so you can have multiple deployments in parallel...

It's not a problem, it's just inefficient. I agree we should profide a facility to create those networks if they aren't created yet.

rhafer:
> It's not a problem, it's just inefficient.

Hm, that's just about creating and plugging a network, a subnet and router. I doubt it brings much inefficiency. In fact I think it might cause problems if we allow multiple deployments on the same network.

I guess it's fine to add that in a different PR. But I wonder if we should at least keep the PREFIX -> network-name relationship.

evrardjp:
> > It's not a problem, it's just inefficient.
> 
> Hm, that's just about creating and plugging a network, a subnet and router. I doubt it brings much inefficiency. In fact I think it might cause problems if we allow multiple deployments on the same network.
> 
> I guess it's fine to add that in a different PR. But I wonder if we should at least keep the PREFIX -> network-name relationship.

I am done fighting for this, went back to drawing board once already. If we want an alternative approach, fine by me.

rhafer:
> I am done fighting for this, went back to drawing board once already. If we want an alternative approach, fine by me.

I don't quite get this reaction. I didn't question the general approach of this PR anywhere. I am just trying to understand a (pretty small) aspect of it and why we're doing it. I didn't even downvote. Not sure what's wrong with that.

doc/source/deploy-guide/prepare-localhost.rst
```diff
@@ -164,33 +164,32 @@ OpenStack CLI’s ``openstack keypair create``) for accessing the
...
-   # objects you create in the cloud
-   export PREFIX=foctodoodle
+   export OS_CLOUD=engcloud
+   # 'engcloud' is the name in the `clouds.yaml`,
    # Set the name of the keypair you created
    export KEYNAME=foctodoodle-key
-   # Set the name of the subnet you created
-   export INTERNAL_SUBNET=${PREFIX}-subnet
+   # Set the name of the network you will use in the deployment
+   export INTERNAL_NETWORK=foctodoodle-net
```
toabctl: I don't see the advantage of have now multiple variables you need to remember. Now you need to make sure that your exports of `INTERNAL_NETWORK` and `INTERNAL_SUBNET` match. You can point now to a subnet that is not in the network.

evrardjp:
> > I am done fighting for this, went back to drawing board once already. If we want an alternative approach, fine by me.
> 
> I don't quite get this reaction. I didn't question the general approach of this PR anywhere. I am just trying to understand a (pretty small) aspect of it and why we're doing it. I didn't even downvote. Not sure what's wrong with that.

Sorry, long day for me.
I just meant we can iterate if necessary.
I think @toabctl have an approach to make those networks created with prefix, and if you think it's the way forward, then let's close this, and adapt accordingly.

rhafer:
> I just meant we can iterate if necessary.

That's more or less what I wrote in my earlier comment.

So meanwhile my test deploying this worked. I guess we can merge this and continue the with #166

PR #242, opened 2019-04-08
Removed merged patches

jsuchome:
And added some comments to existing ones


jsuchome:
@Itxaka I just noticed that you're removing some patches in your PR, but this has little more

jsuchome:
Right, this could be done by a robot :-)

PR #249, opened 2019-04-09
Refresh velum-bootstrap blob

rhafer:
This updates the velum-bootstrap tar ball to the latest level from the
the release-3.0 branch. Which should fix issue with the kubeconfig
refresh that we have been experiencing lateley.

It also updates files/tools/generate-environment to work with the latest
changes in the above code (minionID got renamed to minionId).


Itxaka:
minor nit, should we put the commit sha in the filename so we know which one are we using?

rhafer:
> minor nit, should we put the commit sha in the filename so we know which one are we using?

Yeah, I was wondering how to track it was well. Not so sure about putting it into the filename as we'd need to update the playbook with every update then.

Would you be fine with having it in the commit message?

jsuchome:
I think that would be enough

Itxaka:
> Would you be fine with having it in the commit message?

ok for me as long as we have it somewhere that we can refer to, I dont mind :)

toabctl:
for my understanding - are theses files copied from somewhere? if so, could you add the source link to the commit msg please?

rhafer:
> for my understanding - are theses files copied from somewhere? if so, could you add the source link to the commit msg please?

Sure. It's a private repo though (otherwise we would still be consuming is as a submodule). So it's not of much use for people outside the SUSE org.

evrardjp:
> > for my understanding - are theses files copied from somewhere? if so, could you add the source link to the commit msg please?
> 
> Sure. It's a private repo though (otherwise we would still be consuming it as a submodule). So it's not of much use for people outside the SUSE org.

Well technically it can be used by anyone. But we won't promote this as "the official way(TM)".

rhafer:
> Well technically it can be used by anyone. But we won't promote this as "the official way(TM)".

I was referring to the link to the repo not being really useful for anybody outside the org.

PR #251, opened 2019-04-09
Changes made to improve Performance for add compute.

rajdharwadkar:
To skip certain tasks which is not required for add compute.
Fix for SCRD-8788.


run.sh
```diff
@@ -58,7 +58,7 @@ case "$deployment_action" in
         deploy_osh
         ;;
     "add_compute")
-        add_compute
+        add_compute add_computenode
```
kantarun: Is there a need to pass add_computenode ? I mean are there different actions possible for add_compute function ?

script_library/deployment-actions-kvm.sh
```diff
@@ -56,7 +56,12 @@ function deploy_airship(){
 }
 function add_compute(){
     echo "Now Add compute"
-    run_ansible ${socok8s_absolute_dir}/playbooks/add_compute.yml
+    tagged_info=''
+    tags='all'
+    if [[ "${1:-default}" != 'default' ]]; then
+        tagged_info=" --tags $1"
+    fi
+    run_ansible ${socok8s_absolute_dir}/playbooks/add_compute.yml ${tagged_info}
```
kantarun: Do you need that additional logic ? Just directly pass whatever tag ( add_computenode) is used to identify add_compute node steps ? Not sure what other tags can be passed to add_compute function ?

rajdharwadkar:
> Okay for me.
> 
> Still don't like idea of having top level playbook for specific action and still need to pass tags for that action (only one type of flow is possible). Defeats the purpose of having top-level playbook.

Yes but currently our top level playbook name is to deploy. And I dont think remove can be part of deploy playbook.

script_library/deployment-actions-kvm.sh
```diff
@@ -56,7 +56,8 @@ function deploy_airship(){
 }
 function add_compute(){
     echo "Now Add compute"
-    run_ansible ${socok8s_absolute_dir}/playbooks/add_compute.yml
+    tagged_info=" --tags add_computenode"
+    run_ansible ${socok8s_absolute_dir}/playbooks/add_compute.yml ${tagged_info}
```
jgu17: It is kind of odd to run the specific playbook for add compute and with a specific tag for add compute.  I hope we can come back soon to see if there is possibility of a more elegant solution.

script_library/deployment-actions-kvm.sh
```diff
@@ -56,7 +56,8 @@ function deploy_airship(){
 }
 function add_compute(){
     echo "Now Add compute"
-    run_ansible ${socok8s_absolute_dir}/playbooks/add_compute.yml
+    tagged_info=" --tags add_computenode"
```
jgu17: for consistency in the consistency, can we use use "add_compute" or "add_compute_node" so we have underscore between words in the names?

playbooks/roles/airship-deploy-osh/tasks/main.yml
```diff
@@ -156,6 +159,7 @@
   tags:
     - install
     - update_airship_osh_site
+    - add_computenode
 
 - name: Wait 10 minutes before checking the action status
```
jgu17: what's your observation of the time needed for add compute redeploy osh,? If it is less than 10 minutes, can we put in logic that we don't wait for 10 minutes when add compute? or a shorter time?

rajdharwadkar: > what's your observation of the time needed for add compute redeploy osh,? If it is less than 10 minutes, can we put in logic that we don't wait for 10 minutes when add compute? or a shorter time?

I am skipping wait for 10 mins ?  because I am directly doing update software .  you can see tags are for some of the tasks.

script_library/deployment-actions-kvm.sh
rajdharwadkar: > It is kind of odd to run the specific playbook for add compute and with a specific tag for add compute. I hope we can come back soon to see if there is possibility of a more elegant solution.

Yes I agree but not sure where I can add tag because all configure compute and controller is in one role.

jgu17: have you considered to use the deploy playbook with the add_compute_node tag?

playbooks/roles/airship-deploy-osh/tasks/main.yml
jgu17: you are right...

script_library/deployment-actions-kvm.sh
rajdharwadkar: > have you considered to use the deploy playbook with the add_compute_node tag?

I see add/remove compute is different feature so I dont think we should use deploy playbook for that.  maybe rename deploy playbook to generic name ?

playbooks/generic-deploy_airship.yml
```diff
@@ -36,6 +36,13 @@
       tags:
         - install
 
+- hosts: airship-openstack-compute-workers
+  gather_facts: yes
+  any_errors_fatal: true
+  roles:
+    - role: airship-configure-worker
+      when: redeploy_osh_only is defined and (redeploy_osh_only | bool)
```
jgu17: is redeploy_osh_only needed? I thought we have a different tag for running this purpose and we are deprecating this env variable?

rajdharwadkar: > is redeploy_osh_only needed? I thought we have a different tag for running this purpose and we are deprecating this env variable?

if redeploy_osh is undefined then  will go to all the tasks to check in deploy osh ? and I am running this only compute-workers where as other check on  caasp workers.

script_library/deployment-actions-kvm.sh
```diff
@@ -56,7 +56,8 @@ function deploy_airship(){
 }
 function add_compute(){
     echo "Now Add compute"
-    run_ansible ${socok8s_absolute_dir}/playbooks/add_compute.yml
+    tagged_info=" --tags add_compute_node"
```
kantarun: Not sure what's the benefit of having a variable and using it in next line as-is

PR #253, opened 2019-04-10
Remove rocky specific changes to neutron from our side

jsuchome:
With https://review.openstack.org/#/c/651250/ they are available
upstream, we only need to pass openstack release version to the
helm deploy script.


jsuchome:
We actually want to merge this only after that upstream patch is merged

jsuchome:
Upstream patch was merged, I've tested this one again and it works

PR #254, opened 2019-04-10
Fix values locations for keystone and placement

jsuchome:
Following the changes in
  https://review.openstack.org/#/c/647403/
  https://review.openstack.org/#/c/642067/


jsuchome:
I know they will be upstreamed eventually, but until that time we need this for socok8s

PR #255, opened 2019-04-10
Updating airship distro name to align as per spec

kantarun:
Spec: https://review.openstack.org/#/c/643106/

Also distro value is aligned with OSH image convention.


jsuchome:
What images are we talking here? 

OpenStack images used by OSH look like `neutron:rocky-opensuse_15` ... am I confusing it with something else?

jgu17:
this LGTM but not aligned with the spec mentioned in the commit msg?

jgu17:
> What images are we talking here?
> OpenStack images used by OSH look like neutron:rocky-opensuse_15 ... am I confusing it with something else?

These are the airship ucp service images.

PR #256, opened 2019-04-10
Use openSUSE Leap 15 Rocky Horizon image

jgu17:
Use openSUSE Leap 15 Rocky Horizon image, and required Horizon and
helm toolkit chart version.

### DIFF
--- a/site/soc/software/config/versions.yaml
+++ b/site/soc/software/config/versions.yaml
@@ -8,6 +8,12 @@ data:
         location: /armada/airship-components/openstack-helm-infra
         subpath: ingress
     osh:
+      # TODO remove after Treasuremap catches up on uplifting version
+      horizon:
+        location: https://git.openstack.org/openstack/openstack-helm
+        reference: 5c9dd146b8ffec478dc00d56dcf5b6533045cdbd
+        subpath: horizon
+        type: git
       #TODO remove when Airship treasuremap uplift version moves forward again
       # needs commit "make db-{drop,init) py3 compatible"
       glance-htk:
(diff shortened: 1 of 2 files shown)


site/soc/software/config/versions.yaml
```diff
@@ -8,6 +8,13 @@ data:
         location: /armada/airship-components/openstack-helm-infra
         subpath: ingress
     osh:
+      # TODO remove after Treasuremap catches up on uplifting version
+      horizon:
+        location: https://git.openstack.org/openstack/openstack-helm
+        #reference: 38fb90d52824f9d23c99aab91e23097cd9b657d9
```
kantarun: Should we remove that commented version ?

jgu17: yes :-)

jgu17:
> LGTM.
> Do we need to carry these overrides here ..is not there is proposal to add values_overrides directory and then having suse specific overrides there. Need to check may be its not merged yet.

not sure what you referring to. maybe post a link?  This is the same overrides osh deployment uses now.

jsuchome:
Yes, OSH is currently using these overrides.
Eventually we want to move them upstream, see https://review.openstack.org/#/c/651491/ but it may take some time.

jsuchome:
The change looks good, but why does the PR have 3 commits? Can you squash them into one and correctly rebase against master?

site/soc/software/config/versions.yaml
jgu17: removed.

PR #262, opened 2019-04-16
Added socok8s.sh for args bash-completion and man pages

hemna:
This patch adds the new socok8s.sh wrapper script for run.sh

The socok8s.sh script is generated from an argbash template file
(socok8s.m4) and Makefile.  socok8s.sh has a nice help output and
explainations of the inputs and the evironment vars as well as the available
commands to run.  This also adds bash-completion support and man page.

socok8s.sh is an argbash generated script that knows how to parse
arguemnts on the cmdln, output help and then calls a nearly unmodified
run.sh
By default ./socok8s.sh with no command will run 'setup_everything'
socok8s.sh also shows the used environment vars in the help output.

More info about argbash is here:
https://github.com/matejak/argbash

More detail:
This patch adds a new script call socok8s.sh, which is a wrapper
for the run.sh.  socok8s.sh is generated by the new Makefile

This patch also adds a Makefile, which is used to generate the socok8s.sh
script, it's man page and the bash-completion script that you can source
for your shell environment.

The building of the socok8s.sh is via the new socok8s.m4 file.
This is the argbash input file that describes the arguments that the
socok8s.sh script supports, it's help, description and version.  If you want
to add a new argument that the script supports, then you should:
* modify socok8s.m4
* run make all

The socok8s.m4 defines the 3 arguments
* A boolean option (on/off) for quiet mode
* A boolean option (on/off) for preflight only mode
* command = which command to execute (deploy_ses, deploy_caasp, etc)
* environment vars
* DEPLOYMENT_MECHANISM - if not set, defaults to 'openstack'
* SOCOK8S_DEVELOPER_MODE - Turn on developer mode?
* USE_ARA - if not set, defalts to False
* PREFIX - The prefix name for all objects created in the OpenStack Cloud
* KEYNAME - OpenStack security keypair name you created.
* INTERNAL_NETWORK -
* INTERNAL_SUBNET - The name of the subnet you created in OpenStack
* DELETE_ANYWAY - Defaulted to NO

```
└─[$] <git:(argbash*)> ./socok8s.sh -h
This script is used to bootstrap the socok8s dev env. To build the documentation, run tox -edocs
Usage: ./socok8s.sh [-q|--(no-)quiet] [-p|--(no-)pre] [-v|--version] [-h|--help] [<command>]
	<command>: The command you want to run. (default: 'setup_everything')
	-q, --quiet, --no-quiet: suppress noisy output (off by default)
	-p, --pre, --no-pre: Run preflight only (off by default)
	-v, --version: Prints version
	-h, --help: Prints help

Environment variables that are supported:
	SOCOK8S_DEVELOPER_MODE: Enable SOCOK8S script developer mode. (default: 'False')
	DEPLOYMENT_MECHANISM: The deployment type you want. (default: 'openstack')
	USE_ARA: Use ARA?. (default: 'False')
	OS_CLOUD: The cloud name defined in your ~/.config/openstack/clouds.yaml.)
	PREFIX: The prefix name for all objects created in the OpenStack cloud. (default: 'socok8s')
	KEYNAME: openstack security keypair name to add to your server.)
	INTERNAL_NETWORK: INTERNAL_NETWORK to match your network. It will be used as network and server names.)
	INTERNAL_SUBNET: The name of the subnet you created in OpenStack.)
	DELETE_ANYWAY: Force delete all userspace created stuffs in OpenStack?. (default: 'NO')
```


jsuchome:
This looks like a very nice idea, but do we have to download stuff from the internet?

If argbash is not part of our distributions (and sadly it does not seem to be), is there any alternative?

jesusaurus:
I like the idea of a manpage and helpful output, but isn't run.sh already an entry-point wrapper script? Have we reached the point of needing a wrapper script for our wrapper script?

hemna:
> I like the idea of a manpage and helpful output, but isn't run.sh already an entry-point wrapper script? Have we reached the point of needing a wrapper script for our wrapper script?

Well I was trying to be the least intrusive as possible with these changes.  If you look at the socok8s.sh, it's quite extensive in code that's generated, and I'd like to keep the argparsing an manpages generation, etc detached from the meat of the actual processing code.  Also, the advantage is the socok8s.sh can be regenerated with new params as needed without modifying run.sh.   I think this mechanism is a good comprimise of adding man page, bash-completion, argparsing without heavily embedding all of that with the real code (run.sh).

toabctl:
Could/Should we use eg. python and argparse to replace run.sh itself instead of adding another wrapper?

Itxaka:
> Could/Should we use eg. python and argparse to replace run.sh itself instead of adding another wrapper?

+10000000000000000

evrardjp:
> Could/Should we use eg. python and argparse to replace run.sh itself instead of adding another wrapper?

Agreed! I didn't see that comment before proposing click :)

hemna:
I didn't want to disturb all the work that's been done in run.sh, but I'd much rather have this done in python as well.

dirkmueller:
@hemna rebase? or no longer needed?

PR #264, opened 2019-04-16
Cleanup the check for running helm server

jsuchome:
The block/rescue part is not needed, if we always use the stored unit
file that is provided with the socok8s.

Once the unit file becomes part of some (helm) package, local version should be removed.

### DIFF
--- a/playbooks/roles/airship-setup-deployer/tasks/helm-run.yml
+++ b/playbooks/roles/airship-setup-deployer/tasks/helm-run.yml
@@ -1,63 +1,45 @@
 ---
 # Inspired by https://github.com/openstack/openstack-helm-infra/blob/master/roles/build-helm-packages/tasks/setup-helm-serve.yaml
-- block:
-    - name: checking if local helm server is running
-      wait_for:
-        delay: 5
-        timeout: 10
-        host: 127.0.0.1
-        port: 8879
-        state: started
-        msg: "Failure to connect on helm server -- Attempting automatic recovery"
-      register: _server_running
+- name: Copy systemd unit into place for helm server if it is not present
+  template:
+    src: helm-serve.service.j2
+    dest: /etc/systemd/system/helm-serve.service
+    mode: 0640
 
-    - name: Check port 8879 is running a helm repository
-      uri:
-        url: http://127.0.0.1:8879
-        return_content: yes
-      register: this
-      failed_when: "'Helm Repository' not in this.content"
+- name: Ensure helm serve service is running
+  systemd:
+    state: started
+    daemon_reload: yes
+    name: helm-serve
+    enabled: yes
 
-  rescue:
-    - name: Fixing previous failure - moving systemd unit into place for helm server
-      template:
-        src: helm-serve.service.j2
-        dest: /etc/systemd/system/helm-serve.service
-        mode: 0640
+- name: Check if helm server is up
+  wait_for:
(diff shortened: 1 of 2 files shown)


playbooks/roles/deploy-osh/tasks/helm-run.yml
```diff
@@ -1,63 +1,51 @@
...
+  register: helm_serve_service
 
-    - name: Check port 8879 is running a helm repository
-      uri:
-        url: http://127.0.0.1:8879
-        return_content: yes
-      register: this
-      failed_when: "'Helm Repository' not in this.content"
+- name: Copy systemd unit into place for helm server if it is not present
+  when: not helm_serve_service.stat.exists
```
jhesketh: We probably don't need to check if it exists beforehand. This command should be safe to run idempotent and will ensure we get the latest template.

playbooks/roles/deploy-osh/tasks/helm-run.yml
```diff
@@ -1,63 +1,51 @@
...
-        delay: 5
-        host: 127.0.0.1
-        port: 8879
-        state: started
+- name: Check port 8879 is running a helm repository
+  uri:
+    url: http://127.0.0.1:8879
+    return_content: yes
+  register: this
+  failed_when: "'Helm Repository' not in this.content"
```
jhesketh: Is there a race between the helm server listening on 8879 and serving the correct content? Should we have a retry/timeout here?

Also, previously when this check failed the service was restarted and appeared to work. This might have been in place as a workaround for something we're unaware of. We should perhaps leave a comment here or look at a similar "restart on failure".

playbooks/roles/deploy-osh/tasks/helm-run.yml
jsuchome: I'm not sure. I think the original point was, let's have systemd unit file localy as a backup, but in future we want it to have packaged.
So this way it only uses the local copy for the cases it wasn't packaged (yet)

playbooks/roles/deploy-osh/tasks/helm-run.yml
jsuchome: But the only way the original rescue worked was to provide the unit file and start the service again. Now that I make sure the service file is present, I do not see another reason for repair - if it still does not work, user should know

playbooks/roles/airship-setup-deployer/tasks/helm-run.yml
```diff
@@ -1,63 +1,45 @@
...
-    - name: checking if local helm server is running
-      wait_for:
-        delay: 5
-        timeout: 10
-        host: 127.0.0.1
-        port: 8879
-        state: started
-        msg: "Failure to connect on helm server -- Attempting automatic recovery"
-      register: _server_running
+- name: Copy systemd unit into place for helm server if it is not present
```
Itxaka: can we somehow do an include here to include the helm-run.yml instead of duplicating the code?

```
- name: Include osh helm tasks
  include_role:
    name: deploy_osh
    tasks_from: helm-run
```

*Not tested, but theoretically it should work?

jsuchome: In general yes, but I'm not sure if we should start with this place. More of the code is duplicated unfortunately

PR #265, opened 2019-04-17
Do not duplicate the code for deployer setup

jsuchome:


playbooks/roles/airship-setup-deployer/tasks/kubectl-install.yml
```diff
@@ -1,96 +0,0 @@
...
-    - name: Find if runner has a kubeconfig file
-      stat:
-        path: "{{ kubeconfig_file_path }}"
-      delegate_to: localhost
-      register: new_config
-
-    - name: Find if an existing .kube/config file
-      stat:
-        path: "{{ ansible_user_dir }}/.kube/config"
-      delegate_to: localhost
```
jhesketh: playbooks/roles/deploy-osh/tasks/kubectl-install.yml does not have this delegation.

playbooks/roles/airship-setup-deployer/tasks/main.yml
```diff
@@ -43,17 +43,23 @@
     append: yes
 
 - name: Ensure kubectl is installed
-  include: kubectl-install.yml
+  include_role:
```
jgu17: I like the idea here. But the airship roles are the ones used for the product deployment. I'd recommend go the other direction to have deploy_osh to include the airship role instead.

playbooks/roles/airship-setup-deployer/tasks/main.yml
```diff
@@ -43,17 +43,23 @@
...
+    name: deploy_osh
+    tasks_from: kubectl-install
   tags:
     - install
 
 - name: Ensure helm and tiller are up to date
-  include: helm-install.yml
+  include_role:
```
jgu17: same comment as for line 46

playbooks/roles/airship-setup-deployer/tasks/main.yml
```diff
@@ -43,17 +43,23 @@
...
+    name: deploy_osh
+    tasks_from: helm-install
   tags:
     - install
 
 - name: Ensure helm runs on localhost
-  include: helm-run.yml
+  include_role:
```
jgu17: same comments for line 46

playbooks/roles/airship-setup-deployer/tasks/main.yml
```diff
@@ -86,17 +92,9 @@
...
-    - /etc/openstack
-
-- name: Copy clouds.yml
-  become: yes
+- name: Save default clouds.yaml file
   template:
-    src: clouds.yml.j2
-    dest: /etc/openstack/clouds.yaml
+    src: clouds.yaml.j2
+    dest: /root/.config/openstack/clouds.yaml
```
jgu17: why moving to /root?

playbooks/roles/airship-setup-deployer/tasks/main.yml
jsuchome: In general I would not care, but there's slight difference in one of the file where we've already added some change in osh role but not in airship

playbooks/roles/airship-setup-deployer/tasks/main.yml
jsuchome: I'm actually not sure now why did I do that change originally (there's a link to the original commit).  I remember @dirkmueller suggested it and it sounded reasonable

playbooks/roles/airship-setup-deployer/tasks/main.yml
jgu17: I don't care which version to keep either. If the osh version has more features than the airship version, we can move the osh version over to the airship role folder.

playbooks/roles/airship-setup-deployer/tasks/main.yml
jgu17: does that mean now we require only run as root user?

jsuchome: That would require it, yes. Actually you might be right that keeping it under /etc/ looks better

jsuchome:
@jgu17 I removed the cloud config file change. Still keeping OSH files as the only one, as they include at least this change https://github.com/SUSE-Cloud/socok8s/pull/259

Anyway, we could base it on airship files if you think it's better

playbooks/roles/airship-setup-deployer/tasks/kubectl-install.yml
jsuchome: Right, it was changed with https://github.com/SUSE-Cloud/socok8s/pull/259

playbooks/roles/deploy-osh/tasks/helm-run.yml
```diff
@@ -1,12 +1,14 @@
 ---
 # Inspired by https://github.com/openstack/openstack-helm-infra/blob/master/roles/build-helm-packages/tasks/setup-helm-serve.yaml
 - name: Copy systemd unit into place for helm server if it is not present
+  become: yes
```
Itxaka: is this really needed? We are already using the root user for deployment so it should not be necessary to do this unless Im missing something?

jsuchome: I am not sure, I was also surprised.
I was just syncing the code with current master and those changes were introduced with https://github.com/SUSE-Cloud/socok8s/pull/273 without specific explanation

Itxaka:
duplicating my comment here from the chat:

why not create a simple role for client apps that include helm and kubectl + maybe other optional apps like kubebox[0], kube-ops-view[1] or kubectx[2]


[0] https://github.com/astefanutti/kubebox
[1] https://github.com/hjacobs/kube-ops-view
[2] https://github.com/ahmetb/kubectx

playbooks/roles/deploy-osh/tasks/helm-run.yml
Itxaka: yikes

jsuchome:
@Itxaka sure, but that's out of scope of this PR. Actually I think Jira ticket for such work should be created

playbooks/roles/deploy-osh/tasks/helm-run.yml
jgu17: when we run the ansible playbooks on deployer, we should be able to run not as a root user, which is a security recommendation/practice we've been following.

jgu17:
> @jgu17 I removed the cloud config file change. Still keeping OSH files as the only one, as they include at least this change #259
> Anyway, we could base it on airship files if you think it's better

@jsuchome I still prefer to base the file in the airship-setup-deployer role for two reasons: 1) the airship-setup-deployer is more specific/modular in what it does than deploy_osh, and it was created with the purpose for deploy-osh nd deploy-airship toles to share and reuse code in mind; (2) deploy_osh has a lot of things in it fpr helm developers and have deploy_airship to reference that seems not as clean.

jsuchome:
@jgu17 ok, I will rework it so it is the other way

evrardjp:
I would just put things into a single role which can be reused for both parts.

jgu17:
> I would just put things into a single role which can be reused for both parts.

Agreed with JP here. airship_setup_deployer is set for this purpose (we can rename it to reflect that is a shared role by airship and osh).

jsuchome:
Closing in favor of https://github.com/SUSE-Cloud/socok8s/pull/281

PR #266, opened 2019-04-17
Overwrite ingress images with upstream SUSE

jhesketh:


jsuchome:
I guess this works. As it is shared for 2 deployments, it probably makes sense to have in `common-images-overrides.yml` instead of `templates/ingress-*`

PR #268, opened 2019-04-18
Integrate suse image for keystone and ingress

jgu17:
Integrated neutron suse rocky image for ingress controller and keystone
suse rocky image for keystone.


site/soc/software/config/versions.yaml
```diff
@@ -31,11 +36,40 @@ data:
...
         subpath: helm-toolkit
         type: git
+    ucp:
+      keystone:
+        location: https://git.openstack.org/openstack/openstack-helm
+        reference: 229db2f1552b2bfb0beece4ecd4a1b11eac690c0
+        subpath: keystone
+        type: git
+      keystone-htk:
+        location: https://git.openstack.org/openstack/openstack-helm-infr
```
jsuchome: Typo: `openstack-helm-infr`

playbooks/roles/airship-setup-deployer/tasks/main.yml
```diff
@@ -53,6 +53,7 @@
     - install
 
 - name: Ensure helm runs on localhost
+  become: yes
```
Itxaka: why

site/soc/software/charts/osh/openstack-keystone/keystone.yaml
```diff
@@ -30,4 +30,57 @@ data:
...
       replicas:
         api: 1
+      security_context:
+        keystone:
+          pod:
+            runAsUser: 0
+          container:
+            keystone_api:
+              readOnlyRootFilesystem: false
+    conf:
```
Itxaka: this is still kind of terrible. Why not take advantage of the ansible templating capabilities and use this file as a template and *include* the other file with the overrides? 

currently if we change something on our templates and you are not aware, airship deployment will fail. If you link both of this files by including the osh overrides in here, then if we change something it will be fixed automatically, making it easier to maintain, cleaner and much clear where this overrides are coming from.

I just dont understand why are we doing manual work on each template when it should be left for the machines, which are really good at menial tasks :)

site/soc/software/config/versions.yaml
```diff
@@ -31,11 +36,40 @@ data:
         reference: 8f290d3b8d2a8d787f107d05b3be0199d33d7026
         subpath: helm-toolkit
         type: git
+      keystone:
+        location: https://git.openstack.org/openstack/openstack-helm
+        reference: 229db2f1552b2bfb0beece4ecd4a1b11eac690c0
```
Itxaka: why this reference? is this better than master? Is it a known working freeze? Some comments here would really be appreciated for those of us that are not up to date with the airship meetings :)

site/soc/software/charts/osh/openstack-keystone/keystone.yaml
jsuchome: You know that we want to push such change to upstream and then we would not need to carry it with us...

site/soc/software/config/versions.yaml
```diff
@@ -232,8 +266,21 @@ data:
...
-      ingress: {}
-      keystone: {}
+      ingress:
+        ingress_module_init: "{{ suse_osh_registry_location }}/openstackhelm/neutron:{{ suse_openstack_image_version }}"
+        ingress_routed_vip: "{{ suse_osh_registry_location }}/openstackhelm/neutron:{{ suse_openstack_image_version }}"
+      keystone:
+        bootstrap: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+        db_init: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+        keystone_db_sync: "{{ suse_osh_registry_location }}/openstackhelm/keystone:{{ suse_openstack_image_version }}"
+        db_drop: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
```
Itxaka: kind of crazy to have this duplicated data here. Isnt there a better way without having to duplicate all those entries? Like have a generic images common to all included everywhere and then add extra images needed by each component?

site/soc/software/charts/osh/openstack-keystone/keystone.yaml
Itxaka: that fix would still work once we remove it from osh as airship would not need it anymore thus it would not be automatically included here. Its about making it easier to maintain...

playbooks/roles/airship-setup-deployer/tasks/main.yml
jgu17: because someof us run the script on deployer, not as root user.

site/soc/software/charts/osh/openstack-keystone/keystone.yaml
jgu17: @jsuchome do you have the link to the upstream change to include these? yes, definitely this should go upstream.

site/soc/software/config/versions.yaml
jgu17: master is a moving target and is not always validated within airship upstream. This is the airship approach, with an uplifted version, and we override some of the components if we need a different version of some services.

site/soc/software/config/versions.yaml
jgu17: sounds like a good enhancement task even for upstream, but if should not happen, in socok8s. I would say a different task from this though.

jgu17: with that being said, we already use common vars for the registry and os version, this is pretty much fixed  and don't expect frequent changes in the list here.

site/soc/software/charts/osh/openstack-keystone/keystone.yaml
jgu17: I looked at a couple of patches using the value_overrides folder. Good idea, I am curious if the upstream osh team has talked to upstream airship team how the new pattern will be consumed into Airship?

jgu17: we likely need to separate the values_overrides into a different commit pending how the upstream airship plans to consume the new pattern.

jsuchome: Good point, but I do not have an idea how Airship is deploying helm. What do you think, is it possible to use the arguments pointing to values_overrides directories?

jgu17: That needs an upstream discussion the best approach to handle release/distro flavor overrides in Airship w/o having to provide full distro overrides like this PR. I have some ideas but want to talk to the att engineers at ptg next week. For tech preview, I think we should merge this PR for now.

site/soc/software/config/versions.yaml
Itxaka: tbf this was clarified to me by Matthew as it seems that arsihip pulls its own helm repos instead of using the ones deployed by the osh playbooks, so this makes sense :)

site/soc/software/charts/osh/openstack-keystone/keystone.yaml
Itxaka: :+1:

PR #269, opened 2019-04-19
Generating certs for TLS and  using them for public endpoints

kantarun:
Ingress certs are generated for each new deployment and are
different from airship default certs.

Certs are generated via cfssl (CloudFare Utility). Airship promenade
uses same tool to generate SSL certs which are bundled in airship
treasuremap sample sites.

No TLS for public endpoints in ucp.
Added TLS for keystone, cinder, nova, neutron, hoirzon public endpoints
in openstack namespace.

Need to change ingress domain suffix from 'svc.cluster.local' to
'cluster.local' so that tls secret reference is added in k8s ingress
definition for public endpoints. This is different from upstream
treasuremap airskiff site reference. Ingress k8s definition is generated
by helm-toolkit _ingress.tpl template and should have domain different
from svc.{cluster_domain_suffix} value to add tls section.


playbooks/roles/airship-deploy-ucp/tasks/main.yml
```diff
@@ -55,6 +57,17 @@
     - install
     - update_airship_osh_site
 
+- name: Create site network manifests
```
jgu17: why are we separating the network manifest copy from the rest?

playbooks/roles/airship-deploy-ucp/defaults/main.yml
```diff
@@ -1,4 +1,9 @@
 ---
+ucp_namespace_name: "{{ ucp_namespace | default('ucp') }}"
```
jgu17: namespace variables are already defined in common-vars.yaml.

playbooks/roles/airship-deploy-ucp/defaults/main.yml
```diff
@@ -1,4 +1,9 @@
...
+ceph_namespace_name: "{{ ceph_namespace | default('ceph') }}"
```
jgu17: do yo need ceph name space name here?

examples/workdir/env/extravars
```diff
@@ -19,6 +19,12 @@ redeploy_osh_only: false
 #suse_osh_registry_location: "docker.io"
 #suse_osh_image_version: "pike"
 
+# For TLS CA cert: common name used by default
+#suse_tls_cert_CA_common_name: "ca.cluster.local"
+
+# For TLS cert: common name used by default
+#suse_tls_cert_common_name: "server.cluster.local"
```
jgu17: we should allow user defined domain suffix. but that can be in a different commit.

playbooks/roles/airship-deploy-ucp/tasks/main.yml
kantarun: As now network file has jinja template code so need to be copied as template which is done later.

playbooks/roles/airship-deploy-ucp/defaults/main.yml
kantarun: Just copied from various places. We copy site file so not sure we need it or not ?
Ideally these should be defined in common-vars and referred wherever needed. But somehow agreement is to have them defined in individual role defaults and then it follow boilerplate code pattern and copied to keep it consistent usage.

playbooks/roles/airship-deploy-ucp/defaults/main.yml
kantarun: Oddly, only ucp namespace is defined. hmm, may be should define other 2 there as well. Will do that

rhafer:
@kantarun Is this still needed or can we close it for now? You could just reopen if you find time to work on it again.

Itxaka:
Lets close this for now, it can be restored later

PR #270, opened 2019-04-19
Add new test_plugin for openstack preflight

hemna:
This patch adds a new test_plugins directory
    and a test plugin called openstack-preflight.py
    that adds some test for making sure images, networks
    security groups are available against the configured
    openstack deployment.

    the test plugin adds the following tests
    os_check_image - check to see if an image is available on OpenStack
    os_check_network - check to see if a network is defined on OpenStack
    os_check_subnet - check to see if an subnet is defined on OpenStack
    os_check_security_group - check to see if a security group is defined on OpenStack

    Example usage:

    - name: test to make sure my image is available
      fail:
       msg: "Failed to find {{my_image}} on OpenStack
      when: not my_image is os_check_image

    This also adds ansible native tests for caasp, ses, osh
    deployments to ensure their required images and security groups
    are in place.  The failure messages provide detail about what's
    missing.

### DIFF
--- a/playbooks/openstack-osh_instance.yml
+++ b/playbooks/openstack-osh_instance.yml
@@ -8,6 +8,7 @@
       loop:
         - "{{ playbook_dir }}/../vars/common-vars.yml"
         - "{{ playbook_dir }}/../vars/deploy-on-openstack.yml"
+
   roles:
     - role: handle-nodes-on-openstack
       vars:
--- a/playbooks/openstack-preflight_osh.yml
+++ b/playbooks/openstack-preflight_osh.yml
@@ -1,0 +1,37 @@
+#
+# (c) Copyright 2019 SUSE LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+# Playbook to verify that required resources are in the OpenStack
+# Cloud
+
+---
+- hosts: localhost
+  gather_facts: no
+  connection: local
+  tasks:
+    - name: Load generic vars
+      include_vars: "{{ playbook_dir }}/../vars/common-vars.yml"
+    - name: Load openstack vars
+      include_vars: "{{ playbook_dir }}/../vars/deploy-on-openstack.yml"
+
+    - name: _openstack_preflight_osh | Verify OSH Image exists
+      fail:
+        msg: "Could not find Image '{{deploy_on_openstack_oshnode_image}}'"
+      when: not deploy_on_openstack_oshnode_image is os_check_image
+
+    - name: _openstack_preflight_osh | Verify security group exists
+      fail:
+        msg: "Could not find Security Group '{{deploy_on_openstack_oshnode_securitygroup}}'"
+      when: not deploy_on_openstack_oshnode_securitygroup is os_check_security_group
--- a/playbooks/openstack-ses_aio_instance.yml
+++ b/playbooks/openstack-ses_aio_instance.yml
@@ -8,6 +8,7 @@
       loop:
         - "{{ playbook_dir }}/../vars/common-vars.yml"
         - "{{ playbook_dir }}/../vars/deploy-on-openstack.yml"
+
   roles:
     - role: handle-nodes-on-openstack
       vars:
--- a/playbooks/test_plugins/openstack-preflight.py
+++ b/playbooks/test_plugins/openstack-preflight.py
@@ -1,0 +1,95 @@
+# (c) Copyright 2019 SUSE LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+# Load a "packages" manifest from a directory; cache these
+# by default.
+
+# Given such a dictionary, locate the latest version of a
+# named package.
+"""
+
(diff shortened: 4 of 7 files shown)


script_library/openstack-preflight.py
```diff
@@ -0,0 +1,158 @@
...
+                    help="Run the caasp checks?",
+                    default=False, action="store_true")
+parser.add_argument("-o", "--osh",
+                    help="Run the osh checks?",
+                    default=False, action="store_true")
+parser.add_argument("-a", "--all",
+                    help="Run all the checks?",
+                    default=False, action="store_true")
+
+CLOUD = os.getenv('OS_TEST_CLOUD', 'devstack')
```
jesusaurus: Could we use OS_CLOUD here instead of creating a new variable?

hemna: done

hemna:
Hi JP,
  I added this test because I was struggling to find out the required images, networking options, etc that are placed in the vars/deploy-on-openstack.yml file.  If you try and run any of the deploy_* commands with run.sh you will run into a ton of missing items in my own openstack deployment, that I then have to track down and spend a ton of time doing.   Ansible simply spits out errors with large stack traces that are impossible to read to find the real error, a missing image named X.  This wastes a ton of time, and this simple python script tests and spits out errors that are human readable and easy to rectify in a short period of time.   We desperately need some checks in place like this to make this usable to a wider audience.

evrardjp:
> Hi JP,
> I added this test because I was struggling to find out the required images, networking options, etc that are placed in the vars/deploy-on-openstack.yml file. If you try and run any of the deploy_* commands with run.sh you will run into a ton of missing items in my own openstack deployment, that I then have to track down and spend a ton of time doing. Ansible simply spits out errors with large stack traces that are impossible to read to find the real error, a missing image named X. This wastes a ton of time, and this simple python script tests and spits out errors that are human readable and easy to rectify in a short period of time. We desperately need some checks in place like this to make this usable to a wider audience.

This should be all documented already in the deploy guide. It means we have a gap.
Are you trying to run on your own openstack cloud?

hemna:
> > Hi JP,
> > I added this test because I was struggling to find out the required images, networking options, etc that are placed in the vars/deploy-on-openstack.yml file. If you try and run any of the deploy_* commands with run.sh you will run into a ton of missing items in my own openstack deployment, that I then have to track down and spend a ton of time doing. Ansible simply spits out errors with large stack traces that are impossible to read to find the real error, a missing image named X. This wastes a ton of time, and this simple python script tests and spits out errors that are human readable and easy to rectify in a short period of time. We desperately need some checks in place like this to make this usable to a wider audience.
> 
> This should be all documented already in the deploy guide. It means we have a gap.
> Are you trying to run on your own openstack cloud?
> 
> Next to that, ARA would help on the debugging if you don't want to browse your scrollback. It's also documented in the deploy guide.

It's not documented entirely.   Yes I'm trying to deploy against my own openstack deployment.  I think we should have the checks in place to help out.   Would you prefer I make an ansible plugin to allow us to check these entries on the openstack deployment?   The current failures w/o this script in place is very difficult to find the actual missing requirements.   Hence the need for all the checks I added in the script.

playbooks/openstack-caasp_instance.yml
```diff
@@ -10,6 +10,9 @@
     - name: Load openstack vars
       include_vars: "{{ playbook_dir }}/../vars/deploy-on-openstack.yml"
 
+    - name: Preflight OpenStack Caasp resources
+      include: _openstack_preflight_caasp.yml
```
toabctl: This gets now called whenever we call the deploy caasp step. Even if the stack got already successfully deployed. I don't think that is needed and wasted time

playbooks/openstack-osh_instance.yml
```diff
@@ -8,6 +8,10 @@
       loop:
         - "{{ playbook_dir }}/../vars/common-vars.yml"
         - "{{ playbook_dir }}/../vars/deploy-on-openstack.yml"
+
+    - name: Preflight OpenStack oshnode resources
+      include: _openstack_preflight_osh.yml
```
toabctl: same here (gets always called)

playbooks/openstack-ses_aio_instance.yml
```diff
@@ -8,6 +8,10 @@
       loop:
         - "{{ playbook_dir }}/../vars/common-vars.yml"
         - "{{ playbook_dir }}/../vars/deploy-on-openstack.yml"
+
+    - name: Preflight OpenStack SES resources
+      include: _openstack_preflight_ses.yml
```
toabctl: same here

playbooks/test_plugins/openstack-preflight.py
```diff
@@ -0,0 +1,113 @@
...
+    found = False
+    for net in cloud.network.subnets():
+        if net.name == net_name:
+            found = True
+            break
+
+    return found
+
+
+#def ses_checks(cloud, os_vars):
```
toabctl: Can you remove the commented code?

hemna:
> In general I would prefer to have these kind of checks in an extra step (so it is not called every time). Then we can document that a user can run the check when needed.

Currently everytime you execute run.sh deploy_caasp all of the preflight checks are run as well.  Shall we remove those as well?

playbooks/test_plugins/openstack-preflight.py
hemna: done

playbooks/openstack-ses_aio_instance.yml
hemna: refactored

playbooks/openstack-osh_instance.yml
hemna: refactored

playbooks/openstack-caasp_instance.yml
hemna: refactored

playbooks/openstack-preflight_osh.yml
```diff
@@ -0,0 +1,28 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+# Playbook to verify that required resources are in the OpenStack
+# Cloud
+
+---
```
rhafer: This doesn't seem to be a valid ansible play.  It's missing the toplevel attributes. (See also the test failure https://ci.suse.de/blue/organizations/jenkins/cloud-socok8s%2Fsocok8s/detail/PR-270/5/pipeline/34/#step-43-log-132)

Same is true for `openstack-preflight_ses.yml`

hemna: fixed both of those.

dirkmueller:
@hemna rebase?

PR #272, opened 2019-04-22
Fix for upstream repo shuffling

kantarun:
Moved cloning of repo to project specific directory (openstack vs
airship).
And now patches are also applied following same structure.

### DIFF
--- a/vars/manifest.yml
+++ b/vars/manifest.yml
@@ -3,27 +3,36 @@ upstream_repos:
   openstack-helm-infra:
     src: https://git.openstack.org/openstack/openstack-helm-infra
     version: master
+    dest_subdir: openstack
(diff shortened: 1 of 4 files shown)


vars/manifest.yml
```diff
@@ -3,27 +3,36 @@ upstream_repos:
...
     src: https://github.com/openstack/openstack-helm-images
     version: master
+    dest_subdir: openstack
   loci:
     src: https://github.com/openstack/loci
     version: master
-  airship-armada:
+    dest_subdir: openstack
+  armada:
     src: https://github.com/openstack/airship-armada
```
jgu17: this old link is now redirected to https://opendev.org/airship/. should we change to the new url?

PR #275, opened 2019-04-23
docs: some documentation structure improvements

toabctl:
This is useful when reading the documentation. The links (on top and
bottom) can be used to navigate to the previous/next section.


doc/source/deploy-guide/prepare-localhost.rst
```diff
@@ -112,6 +112,18 @@ configuration file loading ara plugins:
 For more details on ARA's web interface, please read
 https://ara.readthedocs.io/en/stable/webserver.html .
 
+Enable pipelining (recommended)
```
evrardjp: If you revert the order, you need to adapt mitogen accordingly, as the content of mitogen doesn't go into [ssh_connection] section.

PR #277, opened 2019-04-23
Add Nova metadata shared secret manifest

jgu17:
Added soc site  Nova metadata shared secret manifest as it is now required by Nova global chart

### DIFF
--- a/site/soc/secrets/passphrases/osh_nova_metadata_proxy_shared_secret.yaml
+++ b/site/soc/secrets/passphrases/osh_nova_metadata_proxy_shared_secret.yaml
@@ -1,0 +1,11 @@
+---
+schema: deckhand/Passphrase/v1
+metadata:
+  schema: metadata/Document/v1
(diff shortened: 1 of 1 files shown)


site/soc/secrets/passphrases/osh_nova_metadata_proxy_shared_secret.yaml
```diff
@@ -0,0 +1,11 @@
+---
```
toabctl: for my understanding - does the master version **not** need this?

jgu17: yes the current master version as of today needs this too.

toabctl: Then splitting the commit would be nice. Then we could revert the "switch to master" commit (if we want to go back to master) but keep the fix

PR #278, opened 2019-04-24
Install kubectl from package

jsuchome:
~Not tested yet~ Seems to work

### DIFF
--- a/playbooks/roles/airship-setup-deployer/tasks/main.yml
+++ b/playbooks/roles/airship-setup-deployer/tasks/main.yml
@@ -17,6 +17,17 @@
   tags:
     - install
 
+- name: Install client packages
+  package:
+    name: "{{ suse_osh_client_packages }}"
+    state: present
+  register: install_client_packages
+  until: install_client_packages is success
+  retries: 5
+  delay: 2
+  tags:
+    - install
+
 - name: Pip install python libraries
   pip:
     name: "{{ suse_airship_deploy_python_libs }}"
@@ -42,8 +53,8 @@
     groups: docker
     append: yes
 
-- name: Ensure kubectl is installed
-  include: kubectl-install.yml
+- name: Ensure kubectl is properly configured
+  include: kubectl-setup.yml
   tags:
     - install
 
--- a/playbooks/roles/deploy-osh/tasks/main.yml
+++ b/playbooks/roles/deploy-osh/tasks/main.yml
@@ -51,8 +51,19 @@
   tags:
     - install
 
-- name: Ensure kubectl is installed
-  include: kubectl-install.yml
+- name: Install client packages
+  package:
+    name: "{{ suse_osh_client_packages }}"
+    state: present
+  register: install_client_packages
+  until: install_client_packages is success
+  retries: 5
+  delay: 2
+  tags:
+    - install
+
+- name: Ensure kubectl is properly configured
+  include: kubectl-setup.yml
   tags:
     - install
 
--- a/playbooks/roles/airship-setup-deployer/defaults/main.yml
+++ b/playbooks/roles/airship-setup-deployer/defaults/main.yml
@@ -16,6 +16,9 @@ suse_airship_deploy_packages:
   - python2-pip
   - python-devel
 
+suse_osh_client_packages:
+  - kubectl
+
 suse_airship_deploy_python_libs:
   - "cmd2<=0.8.7"
   - python-openstackclient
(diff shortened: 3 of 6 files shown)


jsuchome:
`kubectl-install.yml` actually does some more work, like fetching kubeconfig, that needs to be moved elsewhere ... or just renamed

jsuchome:
BTW this is in conflict with https://github.com/SUSE-Cloud/socok8s/pull/265 , I do not care which goes first

playbooks/roles/airship-setup-deployer/tasks/main.yml
```diff
@@ -17,6 +17,17 @@
   tags:
     - install
 
+- name: Install client packages
+  package:
+    name: "{{ suse_osh_client_packages }}"
```
toabctl: Does that work given that `{{ suse_osh_client_packages }}` is a list?

playbooks/roles/airship-setup-deployer/tasks/main.yml
```diff
@@ -17,6 +17,17 @@
...
+    state: present
+  register: install_client_packages
```
toabctl: is that var used anywher?

playbooks/roles/airship-setup-deployer/tasks/main.yml
```diff
@@ -17,6 +17,17 @@
...
+  until: install_client_packages is success
+  retries: 5
```
toabctl: why retry and delay? Do we expect that zypper might be used in parallel?

playbooks/roles/airship-setup-deployer/tasks/main.yml
jsuchome: it is not. I think this is there so that lint does not complain

playbooks/roles/airship-setup-deployer/tasks/main.yml
jsuchome: The whole block is basically copy an paste of the `suse_airship_deploy_packages` installation, so I'd say yes.

playbooks/roles/airship-setup-deployer/tasks/main.yml
jsuchome: Again, copy & paste. But I assume that might have been the original concern, some zypper process already running....

playbooks/roles/deploy-osh/tasks/main.yml
```diff
@@ -51,8 +51,19 @@
   tags:
     - install
 
-- name: Ensure kubectl is installed
```
jgu17: out of curiosity -- does it make sense that we remove all these deployer package and client install tasks from deploy_osh and include the airship-setup-deployer role in the osh deployment playbook?

playbooks/roles/airship-setup-deployer/tasks/main.yml
jgu17: wondering if it is because only one item in the list?

playbooks/roles/deploy-osh/tasks/main.yml
jsuchome: Of course, that's why I wrote that there's a conflict with https://github.com/SUSE-Cloud/socok8s/pull/265. 

The end result is to have only deployment tasks at one place only

playbooks/roles/airship-setup-deployer/tasks/main.yml
jsuchome: I actually plan this list will include helm package once we have it packaged

evrardjp: package action plugin flattens out lists, so it's good if you have one or many packages.

playbooks/roles/airship-setup-deployer/tasks/main.yml
evrardjp: You can blame me for having this kind of things added in ansible-lint.
Recent versions of ansible-lint removed this, because it was considered "too nitpicky". Anyway, this is useful to ensure no network issues arises during installation of packages. As it is a network operation, retries is generally a good thing.

toabctl: Is that something that can be done globally maybe? So all `package` plugins calls do the retry/delay thing?

rhafer: AFAIU that would require changes to the package module in ansible. Or maybe some form for wrapper task for installing packages (used via `include_task` ?).
I think, if we'd want to go down that route we should add it to the backlog handle it in a separate PR.

PR #279, opened 2019-04-24
Add env var SOCOK8S_ENVNAME as unique name

toabctl:
For beeing able to deploy multiple environments in parallel, introduce
the unique environment variable SOCOK8S_ENVNAME.
This variable will be used for deriving other resource names like:

- the workspace name (previously ANSIBLE_RUNNER_DIR) where the state
  will be stored
- OpenStack resources (networks/stacks/servers/volumes/...)

Adding this variable makes it easier to collect resources which belong
to a given deployment or workspace.

This commit is just the first step which adds the variable and uses it
for defining the workspace name.

Note: Every deployment must now set SOCOK8S_ENVNAME before running any
steps from run.sh

Note: ANSIBLE_RUNNER_DIR has no longer any effect. The workspace
directory is now constructed via ${SOCOK8S_ENVNAME}-workspace .

### DIFF
--- a/script_library/pre-flight-checks.sh
+++ b/script_library/pre-flight-checks.sh
@@ -1,5 +1,18 @@
 #!/usr/bin/env bash
 
+check_common_env_vars_set (){
+    # basic checks that are needed for everything else!
+    if [ -z ${SOCOK8S_ENVNAME+x} ]; then
+        echo "No SOCOK8S_ENVNAME given. export SOCOK8S_ENVNAME=... for setting a env name" && exit 1
+    fi
+    # NOTE(toabctl): SOCOK8S_WORKSPACE_BASEDIR is always set in run.sh
+    if [ -z ${SOCOK8S_WORKSPACE_BASEDIR+x} ]; then
+        echo "No SOCOK8S_WORKSPACE_BASEDIR given. export SOCOK8S_WORKSPACE_BASEDIR=... for setting a directory" && exit 1
+    fi
+
+    echo "Using ${SOCOK8S_WORKSPACE_BASEDIR}/${SOCOK8S_ENVNAME}-workspace as workspace directory"
+}
+
 check_openstack_env_vars_set (){
     if [ -z ${OS_CLOUD+x} ]; then
         echo "No OS_CLOUD given. export OS_CLOUD=... corresponding to your clouds.yaml" && exit 1
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -100,3 +100,11 @@ htmlhelp_basename = '%sdoc' % project
 
 # Example configuration for intersphinx: refer to the Python standard library.
 #intersphinx_mapping = {'http://docs.python.org/': None}
+
+
+# A string of reStructuredText that will be included at the beginning of
+# every source file that is read.
+rst_prolog = """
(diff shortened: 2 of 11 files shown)


jsuchome:
So why can't set the default value to `suse-socok8s-deploy` ?

toabctl:
> So why can't set the default value of SOCOK8S_ENVNAME to `suse-socok8s-deploy`

we can. but I thought not having `suse` and having `workspace` in the name makes it much cleaner what this is about

script_library/pre-flight-checks.sh
```diff
@@ -1,5 +1,18 @@
 #!/usr/bin/env bash
 
+check_common_env_vars_set (){
+    # basic checks that are needed for everything else!
+    if [ -z ${SOCOK8S_ENVNAME+x} ]; then
+        echo "No SOCOK8S_ENVNAME given. export SOCOK8S_ENVNAME=... for setting a env name" && exit 1
+    fi
+    # NOTE(toabctl): SOCOK8S_WORKSPACE_BASEDIR is always set in run.sh
+    if [ -z ${SOCOK8S_WORKSPACE_BASEDIR+x} ]; then
+        echo "No SOCOK8S_WORKSPACE_BASDIR given. export SOCOK8S_WORKSPACE_BASEDIR=... for setting a directory" && exit 1
```
evrardjp: nit: _BASEDIR

script_library/pre-flight-checks.sh
```diff
@@ -1,5 +1,18 @@
 #!/usr/bin/env bash
 
+check_common_env_vars_set (){
+    # basic checks that are needed for everything else!
+    if [ -z ${SOCOK8S_ENVNAME+x} ]; then
+        echo "No SOCOK8S_ENVNAME given. export SOCOK8S_ENVNAME=... for setting a env name" && exit 1
+    fi
+    # NOTE(toabctl): SOCOK8S_WORKSPACE_BASEDIR is always set in run.sh
+    if [ -z ${SOCOK8S_WORKSPACE_BASEDIR+x} ]; then
```
evrardjp: If it's always set, why do we check for it being set?

script_library/pre-flight-checks.sh
toabctl: fixed

script_library/pre-flight-checks.sh
toabctl: hm. good question. I can remove the check but it feels safer to me when I leave it in. Should I remove it?

jsuchome:
Yep, finally we can start using the mailing list ;)

script_library/pre-flight-checks.sh
evrardjp: Yeah.

evrardjp:
> Looks good to me. I just wonder if we should remove the remainder of the `PREFIX` env var in `vars/deploy-on-openstack.yml` and replace it with `SOCOK8S_ENVNAME` as well.
> But I guess that can be done in a subsequent PR as well.

I agree there, in subsequent PR.

script_library/pre-flight-checks.sh
evrardjp: I would.

PR #283, opened 2019-04-25
Change default mechanism for run.sh

sgrasley:
In preparation of CCP tech preview. we would like to change the default DEPLOYMENT_MECHANISM to be KVM. The default for tech preview is supposed to be bring your own CAASP and SES. This will set the deployment mechanism to skip openstack environment creation.

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -11,7 +11,7 @@ socok8s_absolute_dir="$( cd "$(dirname "$0")" ; pwd -P )"
 
 # USE an env var to setup where to deploy to
 # by default, ccp will deploy on openstack for inception style fun (and CI).
-DEPLOYMENT_MECHANISM=${DEPLOYMENT_MECHANISM:-"KVM"}
+DEPLOYMENT_MECHANISM=${DEPLOYMENT_MECHANISM:-"kvm"}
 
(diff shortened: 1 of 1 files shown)


sgrasley:
DEPLOYMENT_MECHANISM currently directs run.sh to create an openstack environment where we can deploy ccp.  Changing to KVM just skips the openstack env. deployment.  You can still deploy airship on openstack env.  I re-worded original message to clarify.

toabctl:
Thanks for the hints! @sgrasley can you now squash the 2 commits please? Then I'm happy :)

evrardjp:
I would prefer if those commits would be squashed.

For the name, it's indeed a terrible name, but it made sense when ppl deployed on kvm, and wanted to continue on making modules to deploy the whole thing on kvm.

Right now, nothing kvm is actually done, so I suppose it could be renamed as 'bring your own' (BYO).

sgrasley:
Will squash on merge if that is ok.

toabctl:
> Will squash on merge if that is ok.

@sgrasley I would prefer if we squash **before** we merge. Because only in that case we review what we get at the end (and I care about the commit msg and logical separation between commits).

PR #284, opened 2019-04-25
Integrate suse rocky image for nova

rajdharwadkar:


site/soc/software/charts/osh/openstack-compute-kit/nova.yaml
```diff
@@ -80,15 +80,36 @@ data:
     conf:
       ceph:
         enabled: true
-        monitors: {{ ','.join(suse_airship_deploy_ceph_mons) }}
+        #monitors: {{ ','.join(suse_airship_deploy_ceph_mons) }}
```
toabctl: Why do we need to comment this?

site/soc/software/charts/osh/openstack-compute-kit/nova.yaml
```diff
@@ -80,15 +80,36 @@ data:
...
+        #admin_keyring: {{ ceph_admin_keyring_key }}
```
toabctl: same question here

site/soc/software/charts/osh/openstack-compute-kit/nova.yaml
```diff
@@ -80,15 +80,36 @@ data:
...
         admin_keyring: {{ ceph_admin_keyring_b64key | b64decode }}
         cinder:
-          user: {{ ses_cluster_configuration.cinder.rbd_store_user }}
-          keyring: {{ ses_cluster_configuration.cinder.key }}
+          user: {{ ses_cluster_configuration['cinder']['rbd_store_user'] }}
```
toabctl: How is that change related to the commit msg which is about the nova rocky image?

site/soc/software/charts/osh/openstack-compute-kit/nova.yaml
jgu17: I think we need the monitors, @rajdharwadkar ?

site/soc/software/charts/osh/openstack-compute-kit/nova.yaml
```diff
@@ -80,15 +80,36 @@ data:
...
+          keyring: {{ suse_ses_cinder_pool_key | default( ceph_admin_keyring_b64key | b64decode) }}
```
jgu17: can you explain why we are changing to use suse_ses_cinder_pool_key which seems to be deploy_osh var?

site/soc/software/charts/osh/openstack-compute-kit/nova.yaml
```diff
@@ -80,15 +80,36 @@ data:
...
+          AllowOverride All
+          <IfModule !mod_access_compat.c>
+            Require all granted
+          </IfModule>
+          <IfModule mod_access_compat.c>
+            Order allow,deny
+            Allow from all
+          </IfModule>
+        </Directory>
+      #nova:
```
jgu17: so what is the libvirt type if not spec'ed here?

site/soc/software/config/versions.yaml
```diff
@@ -41,6 +41,11 @@ data:
         reference: 8f290d3b8d2a8d787f107d05b3be0199d33d7026
         subpath: helm-toolkit
         type: git
+      nova-htk:
+        location: https://git.openstack.org/openstack/openstack-helm-infra
+        reference: 8f290d3b8d2a8d787f107d05b3be0199d33d7026
```
jgu17: please update this commit id here to 5e1ecd9840397bf9e8829ce0d98fcb721db1b74e the same as the value from PR 294. The upstream has uplifted the htk version used in Airship.

site/soc/software/charts/osh/openstack-compute-kit/nova.yaml
```diff
@@ -83,9 +83,29 @@ data:
         monitors: {{ ','.join(suse_airship_deploy_ceph_mons) }}
         admin_keyring: {{ ceph_admin_keyring_b64key | b64decode }}
         cinder:
-          user: {{ ses_cluster_configuration.cinder.rbd_store_user }}
+          user: {{ ses_cluster_configuration['cinder']['rbd_store_user'] }}
```
jgu17: was the old style broken or new style is preferred?

rajdharwadkar: I saw that osh team is using that style and thought it should be consistent with that code.  But I see many places to be fixed in our code. so will keep our style.

PR #285, opened 2019-04-26
caasp: Update CaaSP nodes before enrolling

rhafer:
This reduces some complexity as will rely on fewer parts of the velum
automation. It fixes the current deployment issues we are seeing in the
update-caasp playbook.
It should also speed up the deployment a bit since we can run e.g. the
reboot for all nodes in parallel.


toabctl:
@rhafer travis is complaining...

rhafer:
> @rhafer travis is complaining...

Yeah. Let's see if the latest change is any better.

evrardjp:
It is better indeed :)

PR #287, opened 2019-04-26
Add a Jenkinsfile for testing

toabctl:
The Jenkinsfile will be used for PR testing together with the
github-branch-source[1] plugin.
That plugin does periodically scan the SUSE-Cloud organization and
will create new Jenkins jobs whenever a PR (which must be from
somebody with admin or write access and must contain a Jenkinsfile)
against SUSE-Cloud/socok8s is created.

That way we can also test changes to the Jenkinsfile beside changes to
socok8s itself.

[1] https://go.cloudbees.com/docs/plugins/github-branch-source/

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,55 @@
+pipeline {
+
+    options {
+        timestamps()
+        timeout(time: 45, unit: 'MINUTES', activity: true)
+    }
+
+    agent {
+        node {
+            label "cloud-ccp-ci"
+        }
+    }
+    environment {
+        /* use lowercase SOCOK8S_ENVNAME. CaaSP Velum doesn't like it otherwise */
+        SOCOK8S_ENVNAME = "cloud-socok8s-${env.BRANCH_NAME.toLowerCase()}-${env.BUILD_NUMBER}"
+        OS_CLOUD = "engcloud-cloud-ci"
+        KEYNAME = "engcloud-cloud-ci"
+        DELETE_ANYWAY = "YES"
+        SOCOK8S_DEVELOPER_MODE = "True"
+        DEPLOYMENT_MECHANISM = "openstack"
+        ANSIBLE_STDOUT_CALLBACK = "yaml"
+    }
+
+    stages {
+        stage('Show environment information') {
+            steps {
+                sh 'printenv'
+            }
+        }
+
+        stage('Deploy everything') {
+            steps {
+                sh "./run.sh"
+            }
+        }
+    }
+
+    post {
+        failure {
+            script {
+                if (env.hold_instance_for_debug == 'true') {
+                    echo "You can reach this node by connecting to its floating IP as root user, with the default password of your image."
+                    timeout(time: 3, unit: 'HOURS') {
+                        input(message: "Waiting for input before deleting  env ${SOCOK8S_ENVNAME}.")
+                    }
+                }
+            }
+        }
+        always {
+            script {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,45 @@
...
+        timeout(time: 30, unit: 'MINUTES', activity: true)
+    }
+
+    agent {
+        node {
+            label "cloud-ccp-ci"
+        }
+    }
+    environment {
+        SOCOK8S_ENVNAME = "cloud-socok8s-${env.BUILD_NUMBER}"
```
rhafer: Is this really going to be unique? Doesn't the BUILD_NUMBER start from 1 from each PR? In that case it might make sense to add `env.BRANCH_NAME` to the `SOCOK8S_ENVNAME`.

evrardjp:
I love this idea! Can we have multiple jenkins files to represent different jobs? One for PR, one for periodics...

Jenkinsfile
evrardjp: Indeed we used BUILD_NUMBER in periodics but that was always the same job, here the jobs will all be different I guess...

toabctl:
> I love this idea! Can we have multiple jenkins files to represent different jobs? One for PR, one for periodics...

I don't think we need different `Jenkinsfile` files. We can just create periodic jobs which use the `Jenkinsfile` from the master branch. That should be very easy.

Jenkinsfile
toabctl: Fixed. Added `env.BRANCH_NAME`.

toabctl:
Rebased and added a `DELETE_ANYWAY=YES` envvar to cleanup everything

Jenkinsfile
```diff
@@ -0,0 +1,46 @@
...
+            steps {
+                sh "./run.sh"
+            }
+        }
+    }
+
+    post {
+        failure {
+            script {
+                echo "It failed. We want to keep the env here?!"
```
evrardjp: I would change script by this script:
```
      script {
        if (env.hold_instance_for_debug == 'true') {
          echo "You can reach this node by connecting to its floating IP as root user, with the default password of your image."
          timeout(time: 3, unit: 'HOURS') {
               input(message: "Waiting for input before deleting the ccp env.")
          }
        }
```

Advantage: When ppl want to hold an instance for debugging purposes, they can just set a variable in environment just a little above, in the same PR. Super easy, and everything done in advance for the dev.

toabctl: Done

Jenkinsfile
```diff
@@ -0,0 +1,53 @@
...
+            label "cloud-ccp-ci"
+        }
+    }
+    environment {
+        SOCOK8S_ENVNAME = "cloud-socok8s-${env.BRANCH_NAME}-${env.BUILD_NUMBER}"
+        OS_CLOUD = "engcloud-cloud-ci"
+        KEYNAME = "engcloud-cloud-ci"
+        DELETE_ANYWAY = "YES"
+        SOCOK8S_DEVELOPER_MODE = "True"
+        DEPLOYMENT_MECHANISM = "openstack"
```
Itxaka: can we set `ANSIBLE_STDOUT_CALLBACK=yaml` here so the output from the job is more readable?

toabctl: Done. Thanks!

toabctl:
Agreed that we don't want to fix the CI run itself for now but have a CI

PR #289, opened 2019-04-26
Move caasp stack deletion to extra file

toabctl:
Instead of using a "when" statement, just run the deploy or cleanup
playbook during the deployment or teardown step.

### DIFF
--- a/playbooks/openstack-caasp_instance.yml
+++ b/playbooks/openstack-caasp_instance.yml
@@ -15,8 +15,6 @@
         stackname: "{{ socok8s_envname }}-{{ stackname_suffix }}"
 
     - name: Handle creation if not exists
-      when:
-        - not ((caasp_stack_delete | default('false'))| bool)
       block:
         - name: Create stack
           os_stack:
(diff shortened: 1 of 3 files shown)


jsuchome:
What about renaming the original playbook to  openstack-create-caasp?
Or at least have both of them named in the same style, like "prepare-caasp"/"cleanup-caasp" or "create-caasp-instance"/"delete-caasp-instance" ?

evrardjp:
> What about renaming the original playbook to openstack-create-caasp?
> Or at least have both of them named in the same style, like "prepare-caasp"/"cleanup-caasp" or "create-caasp-instance"/"delete-caasp-instance" ?

I guess we could do that for consistency. openstack-create_caasp and openstack-delete_caasp make sense to me.

toabctl:
@

> What about renaming the original playbook to openstack-create-caasp?
> Or at least have both of them named in the same style, like "prepare-caasp"/"cleanup-caasp" or "create-caasp-instance"/"delete-caasp-instance" ?

@jsuchome can that be done in a follow up commit (so I can rename also the network deploy playbook) ?

toabctl:
> What about renaming the original playbook to openstack-create-caasp?
> Or at least have both of them named in the same style, like "prepare-caasp"/"cleanup-caasp" or "create-caasp-instance"/"delete-caasp-instance" ?

@jsuchome done in https://github.com/SUSE-Cloud/socok8s/pull/295

PR #292, opened 2019-04-26
Added missing test keys to ucp/mariadb.yaml chart

matthew-fuller:
This change is needed so that setting run_tests: false and test_timeout to a custom value will be recognized by the ucp/mariadb chart. Somehow this chart got overlooked when this feature was implemented with an earlier change.


rajdharwadkar:
LGTM

matthew-fuller:
Commit message has been amended to include the description above.

PR #294, opened 2019-04-27
Version uplift

jgu17:
Updated soc site versions manifest after treasuremap global versions has been uplifted recently. There is newer sanctioned version for keystone and helmtool kit.


jgu17:
> Can you update to commit msg for 0a52c08 and describe why this is needed please?

Added some more details. Generally we follow the upstream uplift version changes.  Nothing more can be added :-).

PR #296, opened 2019-04-29
Don't add non-SES related repos on the SES node

dirkmueller:


toabctl:
@dirkmueller did you test that?

dirkmueller:
@toabctl no, still trying to understand how to test this

jsuchome:
This seems to make sense

dirkmueller:
potentially it works now, it is beyond ses installation at this point in time (but not yet finished)

PR #297, opened 2019-04-29
Save clouds.yaml under /etc so it is avalable for all users

jsuchome:
That's the same location airship deployer uses

### DIFF
--- a/playbooks/roles/deploy-osh/tasks/main.yml
+++ b/playbooks/roles/deploy-osh/tasks/main.yml
@@ -143,22 +143,22 @@
   tags:
     - run
 
-- name: Create .config/openstack directory
+- name: Ensure /etc/openstack directory exists
+  become: yes
   file:
-    path: /root/.config/openstack
+    path: /etc/openstack
     state: directory
     owner: root
     group: root
     mode: '0755'
-  tags:
-    - run
 
 - name: Save default clouds.yaml file
+  become: yes
   template:
-    src: clouds.yaml.j2
-    dest: /root/.config/openstack/clouds.yaml
     owner: root
     mode: '0640'
+    src: clouds.yml.j2
+    dest: /etc/openstack/clouds.yaml
 
 - name: Setup Basic ingress and namespace
   import_tasks: component-install.yml


toabctl:
I think this is a `openstacksdk` thing. It searches first in `$PWD`, then `$HOME/.config/openstack` and then in `/etc/openstack/` (see https://docs.openstack.org/openstacksdk/latest/user/guides/connect_from_config.html).
So do we run different ansible playbooks with different users? Or what is the problem here?

jsuchome:
We usually run everything as a root so it does not bother us, but airship guys want to make it possible for other users AFAIU: https://github.com/SUSE-Cloud/socok8s/pull/265#discussion_r278183738

jsuchome:
Also related https://github.com/SUSE-Cloud/socok8s/pull/293

playbooks/roles/deploy-osh/tasks/main.yml
```diff
@@ -151,22 +151,19 @@
...
+  with_items:
+    - /etc/openstack
 
 - name: Save default clouds.yaml file
+  become: yes
   template:
-    src: clouds.yaml.j2
-    dest: /root/.config/openstack/clouds.yaml
-    owner: root
-    mode: '0640'
```
dirkmueller: why did you remove the root/mode setting? I think it is better to be explicit?

playbooks/roles/deploy-osh/tasks/main.yml
```diff
@@ -151,22 +151,19 @@
...
-    path: /root/.config/openstack
+    path: "{{ item }}"
     state: directory
-    owner: root
-    group: root
-    mode: '0755'
-  tags:
-    - run
+  with_items:
+    - /etc/openstack
```
dirkmueller: only one item? why not use the previous path: directly?

playbooks/roles/deploy-osh/tasks/main.yml
jgu17: I think this is for the user that runs the playbook to have access to the clouds.yaml that is created? That user can be root, or "stack" for example. maybe eventually (in product version) we need a group that can access the conf files that have secrets in it.

playbooks/roles/deploy-osh/tasks/main.yml
jsuchome: Could be. I just copied the code from airship task

playbooks/roles/deploy-osh/tasks/main.yml
jsuchome: Yeah, it's just copied from the airship part

dirkmueller: as you're running this as root (see become: yes), we should be careful with the mode and owner settings. please just add the owner: and mode: lines that you added back (I would assume it actually needs to be 0644 for non-root to be able to read the clouds.yaml). 

please note that you can split secrets into a separate secret yaml, that could be stored under the ~root user as before so that generic settings other than the password are available to everyone.

playbooks/roles/deploy-osh/tasks/main.yml
```diff
@@ -151,22 +151,22 @@
...
-  tags:
-    - run
 
 - name: Save default clouds.yaml file
+  become: yes
   template:
-    src: clouds.yaml.j2
-    dest: /root/.config/openstack/clouds.yaml
     owner: root
     mode: '0640'
```
rhafer: With these permission other users wont be able to access that file. OTOH if that if clouds.yml contains credentials we don't want to make those `0644`.

jsuchome:
Hm, maybe it does not make much sense after all. I just wanted to sync it with airhip setup code...

Itxaka:
is this still wanted or we can close it @jsuchome ?

jsuchome:
let's close it

PR #299, opened 2019-04-29
Use Heat stack for creating the VIP port

toabctl:
That makes the output parsing to get the actual VIP simpler (we can
just use the Heat output).
Also this is in line with all the other heat stacks we use. So
cleaning up an env is just cleaning up all the heat stacks that start
with the SOCOK8S_ENVNAME .


toabctl:
Not sure if this is the best idea (because the heat stack file is very small). But I like it to have everything in heat stacks.
An alternative would be, to add the port to the network stack I guess.

toabctl:
Tested. seems to work

PR #301, opened 2019-04-29
Fix extravars file creation when only deploying CaaSP

toabctl:
When deploying CaaSP before SES, the env/extravars file was not
created and the "lineinfile" ansible statement failed because the file
was not there.
That is now fixed because using "blockinfile" creates the
file. "blockinfile" is also used in
playbooks/roles/ses-osh/tasks/main.yml which was responsible for
creating env/extravars . Now it doesn't matter which (SES or CaasP) is
deployed first.

Also move the code to write the VIP to env/extravars to the networking
stack creation. The VIP is created there so it makes sense to also set
the VIP from there.

### DIFF
--- a/playbooks/openstack-create_network.yml
+++ b/playbooks/openstack-create_network.yml
@@ -24,3 +24,17 @@
         - name: show network stack output
           debug:
             var: network_stack_create_output
+
+    - name: Get the VIP from the network stack
+      command: openstack stack output show {{ deploy_on_openstack_network_stackname }} vip -c output_value -f value
+      changed_when: false
+      register: _network_stack_vip_output
+
+    - name: Add vip with cidr in extravars
+      delegate_to: localhost
+      blockinfile:
+        path: "{{ socok8s_extravars }}"
+        create: yes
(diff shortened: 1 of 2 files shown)


playbooks/openstack-create_network.yml
```diff
@@ -24,3 +24,16 @@
...
             var: network_stack_create_output
+
+    - name: Get the VIP from the network stack
+      command: openstack stack output show {{ deploy_on_openstack_network_stackname }} vip -c output_value -f value
+      changed_when: false
+      register: _network_stack_vip_output
+
+    - name: Add vip with cidr in extravars
+      delegate_to: localhost
+      blockinfile:
```
rhafer: I think you need to use a different marker here than what is used in the ses-osh role. Other wise the two `blockinfile` calls with conflict.
See: https://docs.ansible.com/ansible/latest/modules/blockinfile_module.html
"When more then one block should be handled in one file you must change the marker per task"

toabctl: Fixed

PR #302, opened 2019-04-29
Pinning cloned repos to specific commit for QA testing

kantarun:
As QA team is now on-board for testing, we need to stick to
specific commits to provide them a consistent snapshot of
locally cloned upstream repositories. Airship internally
(via treasuremap) references specific commit-id for various components
(helm toolkit, openstack services, core components).

May be dev-patcher needs to be enhanced to allow patch set
number in a given review to have consistent snapshot.

All of the commit versions are referring to lastest commit in each
repo master branch to be in sync with current socok8s behavior.

We plan to create git tag on socok8s repos later once this change
is merged and we have a working deployment. Working deployment will
have most of openstack services integrated and are using suse images
with necessary overrides on airship side.

### DIFF
--- a/vars/manifest.yml
+++ b/vars/manifest.yml
@@ -2,37 +2,46 @@
 upstream_repos:
   openstack-helm-infra:
     src: https://git.openstack.org/openstack/openstack-helm-infra
-    version: master
+    # Commit on April 24, 2019
+    version: 0ed4f0de5ed7b0eb7bb6cb6f819ad48e3fd71a23
     dest_subdir: openstack
   openstack-helm:
     src: https://git.openstack.org/openstack/openstack-helm
-    version: master
+    # Commit on April 28,2019
+    version: 4b4745f1cda13d2d1c7b92b474135ebfb7ee689c
     dest_subdir: openstack
   openstack-helm-images:
     src: https://github.com/openstack/openstack-helm-images
-    version: master
+    # Commit on April 23, 2019
+    version: bd9bdf0d0e7e8a3fff18f74b98c3beabec2b2d92
     dest_subdir: openstack
   loci:
     src: https://github.com/openstack/loci
-    version: master
+    # Commit on April 24, 2019
+    version: b901ac0bed3f3dcc3f3f5d332e966a6608d1091b
     dest_subdir: openstack
   armada:
     src: https://opendev.org/airship/armada
-    version: master
+    # Commit on April 23, 2019
+    version: b881e176f5fb63fbfb003389ca2f8922db137ede
     dest_subdir: airship
   shipyard:
     src: https://opendev.org/airship/shipyard
-    version: master
+    # April 29, 2019
+    version: 902ee1b7afef80c8eb352bfcfb4ac4b6b227e87b
     dest_subdir: airship
   deckhand:
     src: https://opendev.org/airship/deckhand
-    version: master
+    # Commit on April 22, 2019
+    version: a611a326be52516cdf69c8b385d56169f53a6a12
     dest_subdir: airship
   pegleg:
     src: https://opendev.org/airship/pegleg
-    version: master
+    # Commit on April 24, 2019
+    version: 50ffabdaf511e4206fd7b7c0123f686b9cb9b2a3
     dest_subdir: airship
   treasuremap:
     src: https://opendev.org/airship/treasuremap
-    version: master
+    # Commit on April 28, 2019
+    version: 74e6baf65e60ab6971f35849a741e9e22da75923
     dest_subdir: airship


vars/manifest.yml
```diff
@@ -2,37 +2,46 @@
...
     src: https://git.openstack.org/openstack/openstack-helm-infra
-    version: master
+    # Commit on April 24, 2019
+    version: 0ed4f0de5ed7b0eb7bb6cb6f819ad48e3fd71a23
     dest_subdir: openstack
   openstack-helm:
     src: https://git.openstack.org/openstack/openstack-helm
-    version: master
+    # Commit on April 28,2019
+    version: 4b4745f1cda13d2d1c7b92b474135ebfb7ee689c
```
jgu17: do we need to pin openstack helm here? don't think it is used by product deployment.

vars/manifest.yml
```diff
@@ -2,37 +2,46 @@
...
     src: https://github.com/openstack/openstack-helm-images
-    version: master
+    # Commit on April 23, 2019
+    version: bd9bdf0d0e7e8a3fff18f74b98c3beabec2b2d92
     dest_subdir: openstack
   loci:
     src: https://github.com/openstack/loci
-    version: master
+    # Commit on April 24, 2019
+    version: b901ac0bed3f3dcc3f3f5d332e966a6608d1091b
```
jgu17: again don't think it is used by product deployment.

vars/manifest.yml
```diff
@@ -2,37 +2,46 @@
...
     src: https://git.openstack.org/openstack/openstack-helm
-    version: master
+    # Commit on April 28,2019
+    version: 4b4745f1cda13d2d1c7b92b474135ebfb7ee689c
     dest_subdir: openstack
   openstack-helm-images:
     src: https://github.com/openstack/openstack-helm-images
-    version: master
+    # Commit on April 23, 2019
+    version: bd9bdf0d0e7e8a3fff18f74b98c3beabec2b2d92
```
jgu17: why pin it since it is not used by product deployment

vars/manifest.yml
kantarun: Reason I did was for consistency sake as it can be confusing for non-team developer member to understand nuances of the reasoning. 
Don't we need to pin these repos for OSH deployment mechanism as well. I was thinking this may be the time.

vars/manifest.yml
kantarun: Reason I did was for consistency sake as it can be confusing for non-team developer member to understand nuances of the reasoning. 
Don't we need to pin these repos for OSH deployment mechanism as well. I was thinking this may be the time.

vars/manifest.yml
kantarun: Reason I did was for consistency sake as it can be confusing for non-team developer member to understand nuances of the reasoning. 
Don't we need to pin these repos for OSH deployment mechanism as well. I was thinking this may be the time.

jgu17:
need osh squad's opinion what's the best for them. If they like the osh pinned down as well, I don't mind your approach. Otherwise, we can do two commits, one is airship dev patch version pin, and the second is osh-infa dev patch version pin. we can add the osh-infra version pin before we tag release, and revert it afterwards for the osh squad.

vars/manifest.yml
dirkmueller: @kantarun can we maybe just stop checking it out then if it isn't used? then its clear and we're also not downloading stuff that we don't need..

vars/manifest.yml
```diff
@@ -2,37 +2,46 @@
 upstream_repos:
   openstack-helm-infra:
     src: https://git.openstack.org/openstack/openstack-helm-infra
-    version: master
+    # Commit on April 24, 2019
+    version: 0ed4f0de5ed7b0eb7bb6cb6f819ad48e3fd71a23
```
dirkmueller: I think we need to spend a little more thought on this and add one layer of indirection. what I can imagne is a list of repo/githashes under a specific tag (like e.g. ccp_preview_m1) and only if some magic flag is set (env) then those git hashes are used instead of master. 

we could set that flag by default globlaly somewhere, but imho there has to be a convenient way to set everything back to "latest" for continuous integration and development purposes. maybe we connect it to the already existing developer mode flag for now.

toabctl: Something like we have already for git project. `latest`, `stable-X` or a specific commit.

evrardjp: I would also do a developer_mode -> latest + fetch the upstream patches, if not, use the specific sha. I agree a different flag could be used too, to avoid mixing things up, and still keep the consistent interface. Because everything is cloned with ansible, we could have conditionals and ternaries in the "version" line.

vars/manifest.yml
```diff
@@ -2,37 +2,46 @@
...
     dest_subdir: openstack
   openstack-helm:
     src: https://git.openstack.org/openstack/openstack-helm
-    version: master
+    # Commit on April 28,2019
```
evrardjp: Please freeze all the repos at the same time. They are then considered to be working together as they have passed gates together. Different times means there could be problems.

PR #305, opened 2019-04-30
Use a current leap image rather than GM

dirkmueller:
The openSUSE Leap team provides a daily updated Leap 15.0 image,
so we don't need to run against a GM image. By using an updated
image, we don't need to spend 10 minutes on downloading and installing
8 months of maintenance patches.

Also cleanup repo list a bit.

### DIFF
--- a/vars/deploy-on-openstack.yml
+++ b/vars/deploy-on-openstack.yml
@@ -18,10 +18,7 @@ deploy_on_openstack_sesnode_userdata: |
 deploy_on_openstack_repos_to_configure:
   SLE12SP3-product: http://provo-clouddata.cloud.suse.de/repos/x86_64/SLES12-SP3-Pool/
   SLE12SP3-update: http://provo-clouddata.cloud.suse.de/repos/x86_64/SLES12-SP3-Updates/
-  SLE12SP3SDK-product: http://provo-clouddata.cloud.suse.de/repos/x86_64/SLE12-SP3-SDK-Pool/
-  SLE12SP3SDK-update: http://provo-clouddata.cloud.suse.de/repos/x86_64/SLE12-SP3-SDK-Updates/
   SUSE-CA: http://download.suse.de/ibs/SUSE:/CA/SLE_12_SP3/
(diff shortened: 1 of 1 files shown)


vars/deploy-on-openstack.yml
```diff
@@ -18,10 +18,7 @@ deploy_on_openstack_sesnode_userdata: |
 deploy_on_openstack_repos_to_configure:
   SLE12SP3-product: http://provo-clouddata.cloud.suse.de/repos/x86_64/SLES12-SP3-Pool/
   SLE12SP3-update: http://provo-clouddata.cloud.suse.de/repos/x86_64/SLES12-SP3-Updates/
-  SLE12SP3SDK-product: http://provo-clouddata.cloud.suse.de/repos/x86_64/SLE12-SP3-SDK-Pool/
-  SLE12SP3SDK-update: http://provo-clouddata.cloud.suse.de/repos/x86_64/SLE12-SP3-SDK-Updates/
   SUSE-CA: http://download.suse.de/ibs/SUSE:/CA/SLE_12_SP3/
-  OpenStack-Cloud8-product: http://provo-clouddata.cloud.suse.de/repos/x86_64/SUSE-OpenStack-Cloud-8-Pool/
```
rhafer: We can't delete these 3. They're still referenced by the ses playbooks. See `deploy_on_openstack_ses_repos_per_imagename` a few lines deeper.

The SES playbook (https://github.com/SUSE-Cloud/socok8s/blob/master/playbooks/openstack-ses_aio_hostconfig.yml#L5) installs a couple of packages from some of those repos. I am sure most of that is unneeded, but that should be cleaned up in a separate PR IMO.

dirkmueller: @rhafer this could be removed after https://github.com/SUSE-Cloud/socok8s/pull/296 which I was hoping would be merged by now

rhafer: Ok. I didn't see that one.

rhafer:
Needs to be merged after #296

PR #306, opened 2019-04-30
Use wait-for-pods from upstream

jsuchome:
This removes our local version of `wait-for-pods.sh`.

There might be a reason why we have it, but I did not find any documentation. @jgu17 you've introduced this change (`or .status == {}'` check)

```
     # Adding status == {} empty check for stuck job as seen in local setup
￼	    # Need to validate this empty status meaning in k8s docs
￼	    kubectl get jobs -o json --namespace=$1 | jq -r \
￼	        '.items[] | .spec.completions == .status.succeeded or .status == {}' | \
```

before. Do we still need it? If so, could you please propose the change upstream instead?


jsuchome:
Current upstream code for the reference: https://github.com/openstack/openstack-helm/blob/master/tools/deployment/common/wait-for-pods.sh#L31

rhafer:
@jsuchome Seems this needs a rebase. Other than that it looks good IMO (if we really can live with out that change)

jsuchome:
rebased

jgu17:
> This removes our local version of wait-for-pods.sh.
> There might be a reason why we have it, but I did not find any documentation. @jgu17 you've introduced this change (or .status == {}' check)
>      # Adding status == {} empty check for stuck job as seen in local setup
> ￼	    # Need to validate this empty status meaning in k8s docs
> ￼	    kubectl get jobs -o json --namespace=$1 | jq -r \
> ￼	        '.items[] | .spec.completions == .status.succeeded or .status == {}' | \
> 
> before. Do we still need it? If so, could you please propose the change upstream instead?

It does what the comments... what we saw at the time was the job status is empty when th epod got stuck. I have not run osh deployment for a long time. If you don't see the issue anymore (which can be caasp specific), I don't see the need to keep it/push it upstream IMO.

jsuchome:
But if the pod is "stuck", that's an error, right? So why would we check for such condition in the same check that checks if job is finished?

jgu17:
> But if the pod is "stuck", that's an error, right? So why would we check for such condition in the same check that checks if job is finished?

if you don't have that check, the script will be waiting forever instead of going to error, if my memory is correct.

jsuchome:
> if you don't have that check, the script will be waiting forever instead of going to error, if my memory > is correct.

That's wierd, right? The script has a timeout for a reason...

PR #307, opened 2019-04-30
Do not update ansible venv all the time

toabctl:
When calling run.sh, currently the ansible venv in the workspace
directory is updated all the time.
This is time consuming and may not work when running multiple run.sh
steps in parallel.
So stop doing that and just use the available venv.

### DIFF
--- a/script_library/bootstrap-ansible-if-necessary.sh
+++ b/script_library/bootstrap-ansible-if-necessary.sh
@@ -10,8 +10,11 @@ function install_ansible (){
     local socok8s_workspace=${SOCOK8S_WORKSPACE_BASEDIR}/${SOCOK8S_ENVNAME}-workspace
     if [[ ! -d ${socok8s_workspace}/.ansiblevenv/ ]]; then
         virtualenv ${socok8s_workspace}/.ansiblevenv/
+        source ${socok8s_workspace}/.ansiblevenv/bin/activate
+        pip install --upgrade -r $(dirname "$0")/script_library/requirements.txt
+        python -m ara.setup.env > ${socok8s_workspace}/.ansiblevenv/ara.rc
+    else
+        echo "Found virtualenv at ${socok8s_workspace}/.ansiblevenv . Using that"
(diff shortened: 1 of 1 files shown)


script_library/bootstrap-ansible-if-necessary.sh
```diff
@@ -10,8 +10,11 @@ function install_ansible (){
     local socok8s_workspace=${SOCOK8S_WORKSPACE_BASEDIR}/${SOCOK8S_ENVNAME}-workspace
     if [[ ! -d ${socok8s_workspace}/.ansiblevenv/ ]]; then
         virtualenv ${socok8s_workspace}/.ansiblevenv/
+        source ${socok8s_workspace}/.ansiblevenv/bin/activate
```
jsuchome: BTW, could you add explicit `pip install --upgrade pip`?

toabctl: why?

jsuchome: Well at least my pip is complaining about it and I have to do it manually

toabctl: can we do that in another commit/PR ?

dirkmueller: Yeah, in place upgrading pip is more fun.. let's not add it here

script_library/bootstrap-ansible-if-necessary.sh
```diff
@@ -10,8 +10,11 @@ function install_ansible (){
...
+        pip install --upgrade -r $(dirname "$0")/script_library/requirements.txt
+        python -m ara.setup.env > ${socok8s_workspace}/.ansiblevenv/ara.rc
+    else
+        echo "Found virtualenv at ${socok8s_workspace}/.ansiblevenv . Using that"
```
dirkmueller: NIT: 
```suggestion
        echo "Re-using existing virtualenv at ${socok8s_workspace}/.ansiblevenv"
```

PR #308, opened 2019-04-30
Add log collection for k8s resources

Itxaka:
Reuses the upstream ansible roles to collect logs from the k8s
resources for troubleshooting

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -1,6 +1,9 @@
 exclude_paths:
   - submodules
   - playbooks/roles/ansible-config_template
+  - playbooks/roles/describe-kubernetes-objects
+  - playbooks/roles/gather-pod-logs
+  - playbooks/roles/helm-release-status
   - playbooks/roles/ses
 # Allow warning 204, which means we are enabling longer than 120 chars lines.
 skip_list:
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -90,11 +90,16 @@ pipeline {
                     }
                 }
             }
-        }
-        always {
             script {
-                sh './run.sh teardown'
+                sh './run.sh gather_logs'
             }
+            zip archive: true, dir: 'logs/', zipFile: 'logs.zip'
+            archiveArtifacts artifacts: 'logs.zip'
+        }
+        cleanup {
+            script {
+                    sh './run.sh teardown'
+                }
         }
     }
 }
--- a/playbooks/generic-collect-logs.yml
+++ b/playbooks/generic-collect-logs.yml
@@ -1,0 +1,61 @@
+---
+#
+# (c) Copyright 2019 SUSE LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+
+- hosts: soc-deployer
+  gather_facts: false
+  vars:
+    # this is where on the target machine the logs will be copied
+    logs_dir: "/tmp/socok8s_gather_logs"
+    zuul:
+      executor:
+        # this is where on the ansible machine the logs will be copied
+        log_root: "../logs"
+  pre_tasks:
+    - name: Load common variables
+      include_vars: "{{ playbook_dir }}/../vars/common-vars.yml"
+    - name: Clear existing local logs dir
+      file:
+        path: "{{ zuul['executor']['log_root'] }}"
+        state: absent
+      delegate_to: localhost
+    - name: Create logs dir
+      file:
+        path: "{{ zuul['executor']['log_root'] }}"
+        state: directory
+      delegate_to: localhost
+    - name: Clear existing remote logs dir
+      file:
+        path: "{{ logs_dir }}"
+        state: absent
+    - name: Install jq
+      zypper:
+        name: jq
+        state: present
+        update_cache: yes
+      register: _pkginstall
+      until: _pkginstall is success
+      retries: 5
+  roles:
+    - role: helm-release-status
+      tags:
+        - helm-release-status
+    - role: describe-kubernetes-objects
+      tags:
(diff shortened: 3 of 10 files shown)


script_library/deployment-actions-openstack.sh
```diff
@@ -81,6 +81,7 @@ function clean_airship(){
     fi
 }
 function teardown(){
+    gather_logs
```
jsuchome: I guess we do not need it always, not for local setups usually. So maybe some variable?

Itxaka: good point

Itxaka:
> I like this, but I would remove the logs/ folder from this, and put the logs in /var/logs/ or /tmp/ or in the workspace folder.

Im not sure how the jenkins workspace works, but are you able to reach /tmp properly from the workspace? or do you mean the actual socok8s workspace?

> In fact, I would prefer if there was a LOG_LOCATION, which defaults to something in the workspace (not necessary the socok8s cloned folder), and ensure this folder is created.

This sounds good, let me add this

evrardjp:
> Im not sure how the jenkins workspace works, but are you able to reach /tmp properly from the workspace? or do you mean the actual socok8s workspace?

I would avoid using /tmp. We need to then think about running in parallel and stuff. So yes, the current socok8s workspace folder sounds nice.

Itxaka:
unfortunately I found an issue with this. If the job fails before deploy_osh jenkins will still try to run this, but becuase the machines are not up the step will fail and the other cleanup steps will not run so this may block the cleanup part.

Blocking this until I know how to ignore a failed step...

Itxaka:
btw I removed the log configuration because I could not get it to work proper, its still missing some proper env vars to get the actual logs from the workspace so I decided to keep it simple and maybe iterate after this is merged if someone is up to spend more time in this.

Jenkinsfile.integration
```diff
@@ -90,11 +90,16 @@ pipeline {
                     }
                 }
             }
-        }
-        always {
             script {
-                sh './run.sh teardown'
+                sh './run.sh gather_logs'
```
evrardjp: no logs if everything is fine?
I would gather logs before cleanup.

run.sh
```diff
@@ -120,6 +120,9 @@ case "$deployment_action" in
     "clean_airship")
         clean_airship
         ;;
+    "gather_logs")
```
evrardjp: Do we really need a user interface for this?

script_library/deployment-actions-openstack.sh
```diff
@@ -81,6 +81,10 @@ function clean_airship(){
     fi
 }
 function teardown(){
+    if [[ ${SOCOK8S_GATHER_LOGS:-"NO"} == "YES" ]]
```
evrardjp: if we are exposing a new feature, I don't see why we should hide this in teardown.

Jenkinsfile.integration
jsuchome: Why gather logs if everything is fine?

run.sh
Itxaka: yes. If Im deploying locally and it fails during deploy_osh I want to be able to gather all logs, because I might not be able to troubleshoot it in that moment, so its good to expose this

script_library/deployment-actions-openstack.sh
Itxaka: expose but not force the user to use it, let them choose if they want to ignore gathering logs or not. I dont see a problem here, after all gather_logs takes time and you might not be interested in spending any

Itxaka:
can we get this in? I dont want to spend time bikeshedding this, the objective is simple, add gather_logs, add them in failure case to jenkins, expose it to developers in case they want ot use it.

Lets not expend valuable time in discussing something so simple please.

playbooks/generic-collect-logs.yml
```diff
@@ -0,0 +1,61 @@
...
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+
+- hosts: soc-deployer
+  gather_facts: false
+  vars:
+    # this is where on the target machine the logs will be copied
+    logs_dir: "/tmp/socok8s_gather_logs"
+    zuul:
```
jhesketh: (nit) This is being ran in our local Jenkins right? Why are you structuring the vars around zuul? I get the upstream gathering might use this structure, but we could still just define the path.

.ansible-lint
```diff
@@ -1,6 +1,9 @@
 exclude_paths:
   - submodules
   - playbooks/roles/ansible-config_template
+  - playbooks/roles/describe-kubernetes-objects
```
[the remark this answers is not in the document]
Itxaka: well those are upstream playbooks and Im not sure they are using ansible lint, so checking those means sending upstream changes which may not be well received without the usual discussion, blueprints, bikeshedding, etc...

So we m,aye be able to contribute linting back upstream...but in the far future when we have time :P

playbooks/generic-collect-logs.yml
Itxaka: yeah so this is because we are using the upstream playbooks for the gathering, and those are used on zuul primary they all use the ansible vars `{{ zuul.executor.log_root }}` for the logs (see here[0]).

As we just want to leverage those upstream playbooks with as little changes as possible, its easier to just pass those vars from our include_role so the logs get copied in the proper dir.

Im not sure how would we be able to override those vars any other way btw, I just did a quick job, maybe there is a way of overriding those but not that I know of :)


[0] https://github.com/openstack/openstack-helm-infra/blob/master/roles/gather-pod-logs/tasks/main.yaml#L59

Jenkinsfile.integration
evrardjp: Because of comparisons or trends?
I don't see a reason why logs shouldn't be gathered in fact. Is storage really a problem?

run.sh
evrardjp: I am fine with that, I just love playing devil's advocate :D

script_library/deployment-actions-openstack.sh
evrardjp: I see. It's just to avoid time.
I would still prefer being explicit than implicit. If you want to gather logs, just run the command before teardown, else don't run the command. It reduces the code we have to maintain, and code paths we have to remember.

evrardjp:
All of my comments can be achieved in a later patchset with a refactor. They are only about reducing cognitive load. It shouldn't prevent this from merging.

Itxaka:
> All of my comments can be achieved in a later patchset with a refactor. They are only about reducing cognitive load. It shouldn't prevent this from merging.

Awesome, then moving on with this!

PR #310, opened 2019-04-30
Jenkinsfile: Run steps in parallel

toabctl:
Instead of deploying SES, then CaaSP, then ..., do steps in parallel
where possible.
This should speedup the CI runs a lot.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -28,9 +28,53 @@ pipeline {
             }
         }
 
-        stage('Deploy everything') {
+        stage('Setup network') {
             steps {
-                sh "./run.sh"
+                sh "./run.sh deploy_network"
+            }
+        }
+        stage('Setup hosts') {
+            parallel {
+                stage('Setup CaaSP') {
+                    steps {
+                        sh "./run.sh deploy_caasp"
+                    }
+                }
+                stage('Setup SES') {
+                    steps {
+                        sh "./run.sh deploy_ses"
+                    }
+                }
+                stage('Setup Deployer') {
+                    steps {
+                        sh "./run.sh deploy_ccp_deployer"
+                    }
+                }
+            }
+        }
+        stage('Enroll CaaSP workers') {
+            steps {
+                sh "./run.sh enroll_caasp_workers"
+            }
+        }
+        stage('Setup CaaSP workers and apply upstream patches') {
+            parallel {
+                stage('Setup CaaSP workes for OpenStack Helm') {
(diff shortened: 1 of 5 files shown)


Jenkinsfile
```diff
@@ -28,9 +28,45 @@ pipeline {
...
+                sh "./run.sh deploy_network"
+            }
+        }
+        stage('Setup hosts') {
+            parallel {
+                stage('Setup CaaSP') {
+                    steps {
+                        sh "./run.sh deploy_caasp"
+                        sh "./run.sh enroll_caasp_workers"
+                        sh "./run.sh setup_caasp_worker_for_openstack"
```
jsuchome: It's _setup_caasp_workers_for_openstack-

toabctl: fixed

Jenkinsfile
```diff
@@ -28,9 +28,37 @@ pipeline {
...
+                sh "./run.sh deploy_network"
+            }
+        }
+        stage('Setup hosts') {
+            parallel {
+                stage('Setup CaaSP') {
+                    steps {
+                        sh "./run.sh deploy_caasp"
+                        sh "./run.sh enroll_caasp_workers"
+                        sh "./run.sh setup_caasp_workers_for_openstack"
```
jsuchome: Actually I think `setup_caasp_workers_for_openstack` cannot run before `deploy_ccp_deployer`, see `osh-deployer-in-hosts.yml` task

toabctl: Fixed

rhafer:
Does this need anything from https://github.com/SUSE-Cloud/socok8s/pull/145 or is this an alternative approach?

toabctl:
> Does this need anything from #145 or is this an alternative approach?

I think this is a different thing. Not sure if we need something from #145

toabctl:
Seems to be always failing with:
```
msg: Timeout waiting for the floating IP to be attached.
```
See https://ci.suse.de/blue/organizations/jenkins/cloud-socok8s%2Fsocok8s/detail/PR-310/6/pipeline/34#step-43-log-354

jsuchome:
Would it be possible that the whole job fails once _any_ of parallel step fails?

rhafer:
https://ci.suse.de/blue/organizations/jenkins/cloud-socok8s%2Fsocok8s/detail/PR-310/18/pipeline/  succeeded. The subsequent failure of the later job was unrelated.

PR #311, opened 2019-04-30
Jenkinsfile: Always run the cleanup step

toabctl:
Instead of only running the cleanup when the post-condition is
"failure", always run the post script.
Otherwise we do not cleanup eg. aborted builds.


toabctl:
doesn't make sense. there is already an always section

PR #314, opened 2019-04-30
Create separate tmp directory outside workspace

dirkmueller:
The workspace is currently fiddled with temporary files like
the virtualenv (which is only created if not the right packages
are installed upfront) and the persistent config files.
    
Separate the venv which is scratch (cache) data from the config
dir so that the $HOME directory is not fulled up with that for
those who have quota on $HOME (like I have). use SOCOK8S_TMP_BASEDIR
for that.

### DIFF
--- a/playbooks/openstack-enroll_caasp_workers.yml
+++ b/playbooks/openstack-enroll_caasp_workers.yml
@@ -39,6 +39,7 @@
           - phantomjs
           - libxml2-devel
           - libxslt-devel
+          - which
 
     - name: Copying velum automation
       unarchive:
--- a/playbooks/openstack-osh_hostconfig.yml
+++ b/playbooks/openstack-osh_hostconfig.yml
@@ -27,6 +27,9 @@
         option: download.use_deltarpm
         value: "false"
 
+    - name: Workaround stupid JeOS image bugs (bsc#1138727)
+      shell: "netconfig update -f"
+
     - name: Configure host repositories
       zypper_repository:
         name: "{{ item }}"


run.sh
```diff
@@ -14,7 +14,9 @@ socok8s_absolute_dir="$( cd "$(dirname "$0")" ; pwd -P )"
 DEPLOYMENT_MECHANISM=${DEPLOYMENT_MECHANISM:-"kvm"}
 
 # The base directory where workspace(s) are created in
-SOCOK8S_WORKSPACE_BASEDIR=${SOCOK8S_WORKSPACE_BASEDIR:-~}
+SOCOK8S_WORKSPACE_BASEDIR_DEFAULT=${TMPDIR:-${XDG_CACHE_HOME:-$HOME/.cache}}
```
jgu17: The artifacts in the workspace includes generated passwords. They need to be kept around for the user to find the passwords to use for some services, for example, shipyard service account, or try to connect to MariaDB directly on cli, or looked up during the upgrade/redeploy so we don't generate new passwords. Default to user home is not ideal, maybe /opt/soc or /opt/airship?

evrardjp: Cache is indeed incorrect. ~/.config/socok8s would be another choice. @dirkmueller what can we use to find that home config folder? It's not in env vars IIRC (not a data folder)

dirkmueller: @evrardjp @jgu17 thanks for the explanations! I have fixed this.

doc/source/deploy-guide/prepare-localhost.rst
```diff
@@ -59,11 +59,10 @@ Install the following software on your `localhost`:
...
   * ansible>=2.7.0
-  * python-openstackclient
-  * python-requests
-  * python-jmespath
-  * python-openstacksdk
-  * python-netaddr
+  * python/python3-openstackclient
+  * python/python3-requests
+  * python/python3-openstacksdk
+  * python/python3--netaddr
```
toabctl: double `--`

dirkmueller: resolved

script_library/bootstrap-ansible-if-necessary.sh
```diff
@@ -6,16 +6,16 @@ set -e
...
-        source ${socok8s_workspace}/.ansiblevenv/bin/activate
-        pip install --upgrade pip
-        pip install --upgrade -r $(dirname "$0")/script_library/requirements.txt
-        python -m ara.setup.env > ${socok8s_workspace}/.ansiblevenv/ara.rc
+    # NOTE(toabctl): ${SOCOK8S_TMP_BASEDIR} and ${SOCOK8S_ENVNAME} are always set
+    local socok8s_venvspace=$SOCOK8S_TMP_BASEDIR/${SOCOK8S_ENVNAME}-cache
+    if [[ ! -d ${socok8s_venvspace}/.ansiblevenv/ ]]; then
+        virtualenv ${socok8s_venvspace}/.ansiblevenv/
+        source ${socok8s_venvspace}/.ansiblevenv/bin/activate
+        pip3 install --upgrade pip
```
Itxaka: should we choose this based on the default python used by ansible-playbook? seems like we are forcing python3 here while ansible could be installed in a python2 env or even virtualenv

script_library/bootstrap-ansible-if-necessary.sh
```diff
@@ -6,16 +6,16 @@ set -e
...
-    if [[ ! -d ${socok8s_workspace}/.ansiblevenv/ ]]; then
-        virtualenv ${socok8s_workspace}/.ansiblevenv/
-        source ${socok8s_workspace}/.ansiblevenv/bin/activate
-        pip install --upgrade pip
-        pip install --upgrade -r $(dirname "$0")/script_library/requirements.txt
-        python -m ara.setup.env > ${socok8s_workspace}/.ansiblevenv/ara.rc
+    # NOTE(toabctl): ${SOCOK8S_TMP_BASEDIR} and ${SOCOK8S_ENVNAME} are always set
+    local socok8s_venvspace=$SOCOK8S_TMP_BASEDIR/${SOCOK8S_ENVNAME}-cache
+    if [[ ! -d ${socok8s_venvspace}/.ansiblevenv/ ]]; then
+        virtualenv ${socok8s_venvspace}/.ansiblevenv/
```
Itxaka: virtualenv here could point to python2 virtualenv, if we want to force this to be a python3 venv (which seems to me based on line 14 use of pip3) then we should pass the `--python=python3` flag here

toabctl:
@dirkmueller needs a rebase.

dirkmueller:
don't have time for this anymore

PR #315, opened 2019-05-02
Use correct Dockerfile intended for Leap images

jsuchome:
Only merge this after https://review.opendev.org/#/c/651787/

### DIFF
--- a/playbooks/roles/suse-build-images/defaults/main.yml
+++ b/playbooks/roles/suse-build-images/defaults/main.yml
@@ -71,8 +71,8 @@ docker_images: []
 #     tags:
 #       libvirt: "{{ registry_location }}openstackhelm/libvirt:latest-opensuse_15"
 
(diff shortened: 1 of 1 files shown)


playbooks/roles/suse-build-images/defaults/main.yml
```diff
@@ -71,8 +71,8 @@ docker_images: []
 #     tags:
 #       libvirt: "{{ registry_location }}openstackhelm/libvirt:latest-opensuse_15"
 
-# This is the default dockerfile name in OSH images for leap15/SLE15 images.
-docker_dockerfile: "Dockerfile.suse_15"
+# This is the default dockerfile name in OSH images for OpenSUSE Leap15 images.
+docker_dockerfile: "Dockerfile.opensuse_leap_15"
```
evrardjp: I think for consistency it would be nice to be opensuse_leap15 or opensuse_15.

PR #317, opened 2019-05-03
Use a current leap image rather than GM

jsuchome:
The openSUSE Leap team provides a daily updated Leap 15.0 image,
so we don't need to run against a GM image. By using an updated
image, we don't need to spend 10 minutes on downloading and installing
8 months of maintenance patches.

Also cleanup repo list a bit.

(cherry picked from commit 17e4613777eb8ede64f883b4b3714bc4e81e1e8a)


jsuchome:
OK, `openSUSE-Leap-15.0-JeOS.x86_64-15.0.1-OpenStack-Cloud` image is gone from ECP ... :-(

jsuchome:
@dirkmueller do you know about the right Leap image with updates we can use?

dirkmueller:
@will take a look

will:
Oh geeze, do I have to? I've never really even used suse, it'll take me a long time to get up to speed  :(

PR #319, opened 2019-05-03
enroll_caasp_workers should be possible to run in parallel with SES setup

jsuchome:


jsuchome:
Error says `Parallel stages or branches can only be included in a top-level stage` so I assume my attempt is futile (unless I just wrote it wrong)

PR #320, opened 2019-05-03
change ref single quote to backtick 

csymons-suse:


csymons-suse:
too many commits for this simple change. deployments/requirements.rst won't render :ref:'provisioninfra'

PR #321, opened 2019-05-03
change singlequotes to backticks

csymons-suse:


csymons-suse:
System requirements :ref:'provisioninfra' won't render with singlequotes

PR #322, opened 2019-05-06
Update helm-serve.service template to support non-root user.

sgrasley:
Updated osh and airship templates to support non-root user on deployer.

### DIFF
--- a/playbooks/roles/airship-setup-deployer/templates/helm-serve.service.j2
+++ b/playbooks/roles/airship-setup-deployer/templates/helm-serve.service.j2
@@ -3,7 +3,7 @@ Description=Helm Server
 After=network.target
 
 [Service]
-User=root
+User={{ ansible_user }}
 Restart=always
 ExecStart=/usr/local/bin/helm serve
 
--- a/playbooks/roles/deploy-osh/templates/helm-serve.service.j2
+++ b/playbooks/roles/deploy-osh/templates/helm-serve.service.j2
@@ -3,7 +3,7 @@ Description=Helm Server
 After=network.target
 
 [Service]
-User=root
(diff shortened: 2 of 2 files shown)


jsuchome:
Could  you move the template under https://github.com/SUSE-Cloud/socok8s/tree/master/playbooks/roles/common-deployer ?

sgrasley:
> Could you move the template under https://github.com/SUSE-Cloud/socok8s/tree/master/playbooks/roles/common-deployer and make both tasks reference the same template?

That may be possible if we move the tasks to common-deployer too.  I will need to read up on that.

jgu17:
> Could  you move the template under https://github.com/SUSE-Cloud/socok8s/tree/master/playbooks/roles/common-deployer and make both tasks reference the same template?

could/should that be a separate commit?

playbooks/roles/airship-setup-deployer/templates/helm-serve.service.j2
```diff
@@ -3,7 +3,7 @@ Description=Helm Server
 After=network.target
 
 [Service]
-User=root
+User={{ ansible_user }}
```
evrardjp: Could you clarify how do you intent to use this?

This basically expects the ssh plugin to be setting ansible_user. So you need to gather facts and use said plugin.

There is also a need to understand if we're gonna use become for this or not, and where the user comes from (localhost or the deployer node). According to the commit message, I would say this is the deployer node, so I assume this ansible_user is set in the inventory for ssh connection plugin usage.

jsuchome:
> could/should that be a separate commit?

Well I think it could be done directly here. But if you want to postpone such work, it could wait I guess

PR #325, opened 2019-05-07
WIP: enable debug on jenkinsfile

Itxaka:


Jenkinsfile
```diff
@@ -20,6 +20,7 @@ pipeline {
         SOCOK8S_DEVELOPER_MODE = "True"
         DEPLOYMENT_MECHANISM = "openstack"
         ANSIBLE_STDOUT_CALLBACK = "yaml"
+        ANSIBLE_DEBUG = "True"
```
evrardjp: Do we really need this?  Things will become hardly readable.

I expect that ``ANSIBLE_VERBOSITY=2`` (or 1) would be enough for most cases

Itxaka: This was just to catch the problem with CI jenkins jobs failing to reboot

evrardjp: sounds good. We should have a look at pipelining too, and enable it by default in gates if it's not yet done.

Itxaka:
not happening anymore so closing

Itxaka:
hitting the same issue, so reopening to check if I can hit it with this PR

jsuchome:
Looks unnecessary now

PR #326, opened 2019-05-07
Fix docker service handler for non root users.

sgrasley:
Added become to service handler for non root user.


sgrasley:
The instructions we are working on requires that non root users are granted password less sudo capabilities  on the deployer so no other changes are required.

Itxaka:
Interesting. I wonder how having passwordless sudo users instead of root improves security, but thats for architects/security engineers to talk about. Thanks for the update!

PR #328, opened 2019-05-08
Replace Jenkinsfile with Jenkinsfile.integration

toabctl:
Jenkinsfile.integration  is a copy of the Jenkinsfile and adds an
aditional step to clone the git repository.
This is the next step to improve the PR testing and allow multiple
Jenkinsfiles (eg. one for integration, one for docs, one for linting,
...).

Git clone code is copied from
https://github.com/SUSE/caaspctl/blob/master/ci/jenkins/\
  pipelines/prs/caaspctl-integration.Jenkinsfile

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -23,6 +23,22 @@ pipeline {
     }
 
     stages {
+        stage('Git Clone') {
+            steps {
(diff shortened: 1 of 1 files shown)


toabctl:
Not needed.

PR #329, opened 2019-05-08
Move Jenkinsfile to Jenkinsfile.integration

toabctl:
This makes it easier for adding more jobs (like Jenkinsfile.docs,
Jenkinsfile.lint, ...).

Also the new path is used by [0]

[0] https://github.com/SUSE-Cloud/automation/pull/3401


toabctl:
> Would it be a problem to move to CI/ folder?

No. But we need to adjust the path in the job definition (here: https://github.com/SUSE-Cloud/automation/pull/3401), too. 
But we already have the tox and travis files in root. I don't see why we want to move the Jenkinsfiles to a different dir for now

evrardjp:
> > Would it be a problem to move to CI/ folder?
> 
> No. But we need to adjust the path in the job definition (here: [SUSE-Cloud/automation#3401](https://github.com/SUSE-Cloud/automation/pull/3401)), too.
> But we already have the tox and travis files in root. I don't see why we want to move the Jenkinsfiles to a different dir for now

It's all good, don't worry :)

PR #333, opened 2019-05-08
Add airship deploy stages to Jenkinsfile.integration

jesusaurus:
Add support for deploying using airship as well as using helm directly.

This depends on a change to add the new `deployment` parameter to the
jenkins job.

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -58,7 +58,11 @@ pipeline {
                 sh "./run.sh enroll_caasp_workers"
             }
         }
+
         stage('Setup CaaSP workers and apply upstream patches') {
+            when {
+                expression { params.deployment == "helm" }
+            }
             parallel {
                 stage('Setup CaaSP workes for OpenStack Helm') {
                     steps {
@@ -73,9 +77,21 @@ pipeline {
             }
         }
         stage('Deploy OpenStack Helm') {
+            when {
+                expression { params.deployment == "helm" }
+            }
             steps {
                 sh "./run.sh build_images"
                 sh "./run.sh deploy_osh"
+            }
+        }
+
+        stage('Deploy Airship') {
+            when {
+                expression { params.deployment == "airship" }
+            }
+            steps {
+                sh "./run.sh setup_airship"
             }
         }
     }


Jenkinsfile.integration
```diff
@@ -73,11 +77,23 @@ pipeline {
...
+            when {
+                expression { params.deployment == "helm" }
+            }
             steps {
                 sh "./run.sh build_images"
                 sh "./run.sh deploy_osh"
             }
         }
+
+        stage('Deploy Airship') {
```
dirkmueller: why are we deploying airship after helm? shouldn't airship actually drive the osh deployment?

jgu17: I have the same question. For a periodical job, is airship test sufficient from the point of view that it deploys osh charts and suse images. If we want to test osh only deployment also as a periodical job, it should be standalone. I don't like the idea of two periodical jobs mixed, e.g., in this case, when osh only deployment fails, the job will fail and airship deployment doesn't get checked.

jesusaurus: For each run of the job, only one of the two deployment types will be exercised. This is not running the Airship stage after the OSH stage, one of the two will be marked as skipped. The expected use is to run different builds of the job with different parameters in order to exercise the different stages.

jesusaurus:
This change is coupled with https://github.com/SUSE-Cloud/automation/pull/3395

Jenkinsfile.integration
dirkmueller: does setup_airship deploy the openstack services? or what is testing that airship can deploy openstack services?

evrardjp: yes. @jgu17 tell me if I am wrong.

Basically, the whole process sets everything up until the deploy airship part.
At that point, there is a shortcut for developers to deploy the helm charts directly (that's deploy_osh). This used to be the default, as there was no deploy_airship step a few weeks/months ago.
For production deploys using airship, the setup_airship step is used instead, which sets up all the airship components, sets up keystone and barbican for said components (what they call the undercloud). Finally the deployment of openstack or any other helm chart can be done using airship tooling.

PR #334, opened 2019-05-09
Switch to upstream overrides

jsuchome:
No need to provide override values by ourselves when they are already
present upstream.


jsuchome:
You can see this is currently for compute-kit only - I will extend it to rest of components if you do not have anything against this approach

jsuchome:
Actually, not all upstream patches have been merged, we'll need https://review.opendev.org/#/c/653676/

Itxaka:
> Do empty templates work, or should they get dummy entries to be valid? (Just curious on how helm/k8s handles this). I suppose you have tested and everything works okay, but I am wondering about the internals here.

I think we tested this once to see if the values were override or broken by an empty template and helm was inteligent enough to not break anything and deal properly with empty templates.

FAKE EDIT:

just tested, seems to work with no issues:
```
oot@ubuntu:~/openstack-helm-infra# helm upgrade --install mariadb mariadb --namespace=openstack --set volume.enabled=false --set pod.replicas.server=1 --values=/tmp/empty-values.yml
Release "mariadb" does not exist. Installing it now.
NAME:   mariadb
LAST DEPLOYED: Fri May 17 12:09:56 2019
NAMESPACE: openstack
STATUS: DEPLOYED
```

where /tmp/empty-values.yml is:
```
root@ubuntu:~/openstack-helm-infra# cat /tmp/empty-values.yml 
---
```

jsuchome:
The pr-merge job actually did not run the last step, I guess something is wrong with its configuration...

PR #335, opened 2019-05-09
Include documentation for updating socok8s, images, and charts

matthew-fuller:
This PR is being opened primarily to get feedback and suggestions for new topics for inclusion in the operations guide documentation.

### DIFF
--- a/doc/source/operations.rst
+++ b/doc/source/operations.rst
@@ -33,9 +33,6 @@ To remove a compute node, run the following command from the root of the socok8s
 
    Although multiple compute nodes can be added at the same time, they must be removed individually. Once the node has been successfully removed, the host details must be removed from "airship-openstack-compute-workers" group in the inventory.
 
-Adding or removing network nodes
---------------------------------
-
 Change control plane scale profile
 ----------------------------------
 SUSE Containerized OpenStack provides two built-in scale profiles: "minimal," which deploys a single pod for each service, and "ha," which is the default profile and deploys a minimum of 2 pods for each service, or 3 or more pods for services that will be heavily utilized or require a quorum. Changing scale profiles can be accomplished by adding a "scale_profile" key to ${WORKSPACE}/env/extravars and specifying a profile value:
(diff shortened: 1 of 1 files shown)


sgrasley:
We should probably cover the different  run.sh options.   deploy, clean etc.

doc/source/operations.rst
```diff
@@ -55,23 +55,60 @@ Once the appropriate profile has been selected, it can be applied by running the
 Updates
 =======
 
-Update Airship UCP Services
----------------------------
+SUSE Containerized OpenStack is delivered as an rpm package, so performing updates is generally accomplished by simply updating the rpm package to the latest version and redeploying the cloud using the steps outlined in the installation procedure. This is the typical update path and will incorporate all recent changes, as well as automatically updating component chart and image versions. However, there may be use cases where it is desirable to update services or components to specific versions.
```
jgu17: can we add specifics how the user can update individual services, using an example?

doc/source/operations.rst
```diff
@@ -55,23 +55,60 @@ Once the appropriate profile has been selected, it can be applied by running the
...
+SUSE Containerized OpenStack is delivered as an rpm package, so performing updates is generally accomplished by simply updating the rpm package to the latest version and redeploying the cloud using the steps outlined in the installation procedure. This is the typical update path and will incorporate all recent changes, as well as automatically updating component chart and image versions. However, there may be use cases where it is desirable to update services or components to specific versions. 
+
+Updating OpenStack Version
+--------------------------
+
+To make a global change to the OpenStack version used by all component images, create a key in ${WORKSPACE}/env/extravars called "suse_osh_image_version" and set it to the desired value. For example, to use the "ocata" version, add the following line to the extravars file:
+
+.. code-block:: yaml
+
+   suse_osh_image_version: "ocata"
```
jgu17: the default is already rocky for suse_opestack_image_version

doc/source/operations.rst
```diff
@@ -55,23 +55,60 @@ Once the appropriate profile has been selected, it can be applied by running the
...
 ------------------------------------------
 
-Update this repository
-----------------------
+The versions.yaml file can also be used for more advanced update configurations such as using a specific image or Helm chart source version. 
 
-The SUSE Containerized OpenStack repository can be updated by performing the git pull command from the socok8s directory:
+.. note::
+   
+   The following configuration options are not supported for the SUSE Containerized OpenStack Technical Preview. Please proceed at your own risk.
```
jgu17: we need words like "A change to a non default image registry and/or custom built image will lose any product supported by SUSE. "

doc/source/operations.rst
```diff
@@ -55,23 +55,60 @@ Once the appropriate profile has been selected, it can be applied by running the
...
+
+     heat:
+       location: https://git.openstack.org/openstack/openstack-helm
+       reference: ${COMMIT_ID}
+       subpath: heat
+       type: git
+
+.. note::
+
+   When specifying a particular version of a Helm chart, it may be necessary to first create the appropriate subsection under "charts". Airship components such as Deckhand and Shipyard belong under "ucp", OpenStack services belong under "osh", and infrastructure components such as MariaDB and RabbitMQ belong under "osh_infra".
```
jgu17: should we have an update airship ucp service section or combine into one?

doc/source/operations.rst
```diff
@@ -55,23 +55,60 @@ Once the appropriate profile has been selected, it can be applied by running the
...
```
jgu17: I think we are using the maridb and RabbitMQ in the osh section instead.

doc/source/operations.rst
Itxaka: is this reflected in a spec somewhere or it was agreed off-meeting?

doc/source/operations.rst
```diff
@@ -55,23 +55,60 @@ Once the appropriate profile has been selected, it can be applied by running the
...
+
+   heat_api: "registry_location/image_directory/image_name:tag"
+
+Similarly, the versions.yaml file can be used to retrieve a specific version of any Helm chart being deployed. To do so, it will be necessary to provide a repository location, type, and a reference. For example, to use a specific version of the Helm chart for Heat, add the following information to the "osh" section under "charts":
+
+.. code-block:: yaml
+
+     heat:
+       location: https://git.openstack.org/openstack/openstack-helm
+       reference: ${COMMIT_ID}
```
Itxaka: can reference be a branch? like `0.1.0` ? if upstream starts using version numbers that may be a much nicer friendly than a commit_id which some end users may not even know how to find

sgrasley:
> Why is this created against the "airship_doc" branch. Shouldn't this endup in master?

The plan is to merge airship_doc branch back into master when it has been flushed out.

toabctl:
> The plan is to merge airship_doc branch back into master when it has been flushed out.

Then we have (again) a huge PR that is not reviewable in the end. Can we make incremental steps please?

doc/source/operations.rst
jgu17: in the ccp build plan spec under cloud9 folder.

doc/source/operations.rst
matthew-fuller: Ah, good catch. Yes, the 'reference' key can be a branch, commit, or reference and it defaults to master if not specified. We've been using commit IDs during development, so I must have been on auto-pilot when I wrote this.

doc/source/operations.rst
matthew-fuller: This was just an example on how and where to change the value, but I can use a newer version in the example.

doc/source/operations.rst
matthew-fuller: This is covered further down in the Updates section.

doc/source/operations.rst
matthew-fuller: I was originally thinking separate sections, but since the process is the same for each, I decided to combine them into a general update section. I can definitely separate them if that makes more sense.

PR #337, opened 2019-05-10
Generating certs for TLS and using them for public endpoints (DO NOT MERGE)

kantarun:
Ingress certs are generated for each new deployment and are
different from airship default certs.

Certs are generated via cfssl (CloudFare Utility). Airship promenade
uses same tool to generate SSL certs which are bundled in airship
treasuremap sample sites.

No TLS for public endpoints in ucp.
Added TLS for keystone, cinder, nova, neutron, hoirzon public endpoints
in openstack namespace.

Added flag for disabling TLS for ucp and osh namespace.

Need to change ingress domain suffix from 'svc.cluster.local' to
'svc.dustycloud.local' so that tls secret reference is added in k8s ingress
definition for public endpoints. This is different from upstream
treasuremap airskiff site reference. Ingress k8s definition is generated
by helm-toolkit _ingress.tpl template and should have domain different
from svc.{cluster_domain_suffix} value to add tls section.


playbooks/roles/airship-setup-deployer/tasks/cfssl-tls-certs.yml
```diff
@@ -0,0 +1,86 @@
...
+- name: Check if need to generate certs
+  set_fact:
+    _generate_cfssl_certs: True
+  when: ( ( _cfssl_certs_stat.results | selectattr('stat.exists') | map(attribute='stat')| list | length !=4)
+          or (_cfssl_config.changed ) )
+
+- name: Download latest stable cfssl (CloudFare pki/tls toolkit)
+  become: yes
+  shell: >-
+    ./cfssl || (curl -L https://pkg.cfssl.org/R1.2/cfssl_linux-amd64 -o cfssl &&
```
jgu17: we are moving things to rpm package based. I could only find this: https://build.opensuse.org/package/show/home:darix:playground/cfssl

playbooks/roles/dev-patcher/defaults/main.yml
```diff
@@ -20,3 +20,5 @@ dev_patcher_patches:
   - 639861
   # osh-infra: Add support for mulitple VIPS
   - 640115
+  # osh: Usage of rally.conf for services with rally test for debugging
```
jgu17: related to this PR?

vars/common-vars.yml
```diff
@@ -64,6 +66,44 @@ suse_airship_components_image_tag: "{%- if suse_airship_build_local_images or su
 
 socok8s_site_name: "soc"
 
+# suffix used in FQDN for ingress (i.e. "publicly facing" access point)
```
jgu17: should document the new options in the deployment guide.

vars/common-vars.yml
```diff
@@ -64,6 +66,44 @@ suse_airship_components_image_tag: "{%- if suse_airship_build_local_images or su
...
+# FQDN -- > {component_api_name}.{namespace}.{cluster_domain_suffix}
+cluster_domain_suffix: "svc.dustycloud.local"
+
+# Only public endpoints are enabled and disabled from TLS perspective.
+# When TLS is enabled for an endpoint, corresponding TLS (tls.ca, tls.crt,
+# tls.key) attributes are made available for TLS secret and component
+# specific helm charts refer that secret in its ingress spec definition.
+
+# Flag to disable/enable TLS for ucp namespace public endpoints
+disable_tls_ucp: True
```
jgu17: why disable tls on ucp by default? It is exposed on vip ingress, so it is expected to be tls enabled.

bryanstephenson: I'd like to see the default be TLS enabled unless there is a good reason not to. "Secure by default" is getting to be a common expectation, and the fewer things customers have to tweak manually to achieve strong security the better our product (or tech preview) looks.

jgu17: I got the answer form Arun. There are a few issues in UCP that blocks us from turning on TLS for UCP services, mainly around the client side cert is not installed correctly on cli side and between internal services. @bryanstephenson I saw Dirk had discussed with you that TLS is not a blocker for TP. In that case, are you okay if we deliver TP with TLS only on openstack service public end points, but not on Airship UCP services?

@kantarun I'd suggest to remove the ucp tls flag since it is not supported yet.

site/soc/software/config/endpoints.yaml
```diff
@@ -76,18 +76,16 @@ data:
...
         default: /v3
       scheme:
-        # default: "https"
         default: "http"
+        public: "{{ ucp_endpoint_http_scheme }}"
         internal: "http"
       port:
         api:
-          # default: 443
           default: 80
```
jgu17: if you choose not to expose admin end points, why don't we leave admin port to 5000 as well?

site/soc/software/config/endpoints.yaml
```diff
@@ -605,18 +602,16 @@ data:
...
         default: /v3
       scheme:
-        #default: "https"
         default: "http"
+        public: "{{ osh_endpoint_http_scheme }}"
         internal: "http"
       port:
         api:
-          #default: 443
           default: 80
```
jgu17: since you took admin end point off the external domain in the substitution, the goal is to leave admin to be the same as internal. I prefer in that case the admin endpoint to have "identity" and port "5000" instead of "identity" and port "80" which leads to beleive that the admin end point should be exposed on the vip. 

This change forces the user to use openstack client 3.18.0 above which no longer uses admin endpoint for keystone. and we should document that requirement in deployment guide and possibly in ansible playbooks.

site/soc/software/config/endpoints.yaml
kantarun: yes, admin endpoint needs to be same as internal endpoint as its primarily used for token validation by services. Will change it to map to internal endpoint.

site/soc/software/config/endpoints.yaml
kantarun: I am not sure if admin endpoint needs to be exposed from deployer as its primarily used for token validation which is used by services as part of its middleware.
Currently only public endpoints are accessible from deployer. So openstack cli should be using public endpoint only. 
I will change admin endpoint to map to internal endpoint as that's what Guang Yee mentioned about general pattern of cloud 9 service entry.

`dana@ardana-qe006-cp1-c1-m1-mgmt:~> openstack endpoint list --service nova
+----------------------------------+---------+--------------+--------------+---------+-----------+----------------------------------------------+
| ID | Region | Service Name | Service Type | Enabled | Interface | URL |
+----------------------------------+---------+--------------+--------------+---------+-----------+----------------------------------------------+
| 41c157b02fb44842b0a2b2eba786aedc | region1 | nova | compute | True | admin | https://192.168.75.1:8774/v2.1/%(tenant_id)s |
| 8b098a70a0084e3da24ddbe9ce84444a | region1 | nova | compute | True | public | https://10.84.81.4:8774/v2.1/%(tenant_id)s |
| e285ae7e0e83452caaaf6c63563224d0 | region1 | nova | compute | True | internal | https://192.168.75.1:8774/v2.1/%(tenant_id)s |
+----------------------------------+---------+--------------+--------------+---------+-----------+----------------------------------------------+`

playbooks/roles/dev-patcher/defaults/main.yml
kantarun: will remove it as its not needed for this change.

vars/common-vars.yml
kantarun: Added comment here ..highlighting that this flag is not supported and so don't change.

vars/common-vars.yml
kantarun: can do once docs branch is merged.

playbooks/roles/airship-setup-deployer/tasks/cfssl-tls-certs.yml
kantarun: As you mentioned as well, I cannot locate a published/supported opensuse rpm package for cfssl. we can look into building a new package in OBS and adding usage thereafter.

vars/common-vars.yml
dirkmueller: My understanding is that this is a large change very late in the game, and I think security is not the primary purpose of the tech preview (but demonstrating a use case). I totally agree that we should default to a secure configuration for Cloud 10.

jgu17: @dirkmueller I agree with what you said. Can you clarify that you are recommending this not to be merged as tech preview v1 entirely, or change the default to tls disabled for both openstank and airship ucp endpoints?

dirkmueller: we can merge it if its reviewed and not breaking the TP1. The default should not be changed at this point in time because it will invalidate existing testing and make CI getting green harder. if there is risk in merging the patch then I'm okay with postponing it after TP1.

Itxaka:
really outdated and with conflicts, closing and we can reopen later.

PR #338, opened 2019-05-10
Introduction of documentation for technical preview.

sgrasley:
This is the first commit for technical preview documentation.


toabctl:
Why do we have now 2 doc source folders? This is creating so much documentation debt that I would prefer to revert this commit and integrate the written docs properly into what we already had. I know that this will be more work but we could also splitting the commit into smaller parts (eg. the contributor guides can easily be in an extra commit)

evrardjp:
I am confused too.

toabctl:
@sgrasley @jgu17 @matthew-fuller any comments on this?

doc/source_tech_preview/contributor.rst
```diff
@@ -0,0 +1,243 @@
...
+* All code requires to go through our :ref:`reviews`.
+* Documentation should be provided with the
+  code directly. See also :ref:`documentation`.
+
+.. _bug_reporting:
+
+Bug reporting process
+---------------------
+
+Bugs should be filed as Github issues.
```
dirkmueller: That's not actually true - we want to leverage bugzilla for the public beta program for this.

doc/source_tech_preview/deployment/setup-deployer.rst
```diff
@@ -0,0 +1,178 @@
...
+   <username> ALL=(ALL) NOPASSWD: ALL
+
+Add the above line after "#includedir /etc/sudoers.d". replace <username> with
+your username.
+
+
+Configure Ansible
+-----------------
+
+Use ARA (recommended)
```
dirkmueller: why is ARA recommended for non developer builds?

doc/source_tech_preview/operations.rst
```diff
@@ -0,0 +1,293 @@
...
+.. code-block:: console
+
+   kubectl logs -n ${NAMESPACE} ${POD_NAME} -p
+
+.. _caaspoperations:
+
+CaaS Platform Operations
+========================
+
+Disable transactional update for development purposes
```
dirkmueller: why are we documenting this?

doc/source_tech_preview/reference.rst
```diff
@@ -0,0 +1,333 @@
...
+
+The OpenStack-Helm project provides shell scripts to deploy the helm charts,
+with overrides per context (for example, multinode).
+
+Those shell scripts calling the helm installation include an environment
+variable, to allow users to pass extra arguments.
+
+See for example https://github.com/openstack/openstack-helm/blob/c869b4ef4a0e95272155c5d5dd893c72976753cd/tools/deployment/multinode/100-glance.sh#L49 .
+
+SUSE-ification of the OSH charts when deploying in OSH only mode
```
dirkmueller: Please don't use the term "SUSE-ification", it is misleading - there is no green paint added.

jgu17:
> Why do we have now 2 doc source folders? This is creating so much documentation debt that I would prefer to revert this commit and integrate the written docs properly into what we already had. I know that this will be more work but we could also splitting the commit into smaller parts (eg. the contributor guides can easily be in an extra commit)

ah, yes, it was a Friday afternoon mistake. We still using the feature branch. The squad already talked about this yesterday that we need to revert it. Now we have comments from Dirk and others, we'll take thosecomments first before revert.

doc/source_tech_preview/contributor.rst
jgu17: The program meeting decided not to let the tech preview users to file bugs directly, but send issues to the Beta mailing list. Is that aligned to what you expect?

doc/source_tech_preview/deployment/setup-deployer.rst
jgu17: copy and pasted from the existing. Agreed it probably should be in the developer guide.

doc/source_tech_preview/operations.rst
jgu17: agreed too.

doc/source_tech_preview/reference.rst
jgu17: "no green paint added." :-)

PR #341, opened 2019-05-12
add deps to tox.ini, add doc/build_tech_preview to .gitignore

csymons-suse:


sgrasley:
This looks like same as 340.

csymons-suse:
> This looks like same as 340.

340 is for master; this is for airship_doc. I'll close whichever is extraneous. I was building docs_tech-preview in both branches.

sgrasley:
Thanks... I missed that point.  That's what I get for reading review before coffee.  I think the airship_doc branch will be removed soon.

PR #342, opened 2019-05-13
Remove references only for SLE12 based deployer

dirkmueller:
The deployer and the CI job launcher have been switched
to Leap15, so we don't need to reference other internal
paths anymore that are only needed for SLE12. This reduces
the amount of red herrings a bit.

### DIFF
--- a/vars/deploy-on-openstack.yml
+++ b/vars/deploy-on-openstack.yml
@@ -18,13 +18,9 @@ deploy_on_openstack_sesnode_userdata: |
 deploy_on_openstack_repos_to_configure:
   SLE12SP3-product: http://provo-clouddata.cloud.suse.de/repos/x86_64/SLES12-SP3-Pool/
   SLE12SP3-update: http://provo-clouddata.cloud.suse.de/repos/x86_64/SLES12-SP3-Updates/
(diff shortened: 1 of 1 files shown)


vars/deploy-on-openstack.yml
```diff
@@ -18,13 +18,9 @@ deploy_on_openstack_sesnode_userdata: |
 deploy_on_openstack_repos_to_configure:
   SLE12SP3-product: http://provo-clouddata.cloud.suse.de/repos/x86_64/SLES12-SP3-Pool/
```
jgu17: is this repo accessible by the external tech preview users? this is not related to this PR :-)

dirkmueller: it is obviously not accessible by externals, but thats unrelated to this PR.

evrardjp:
I don't see extra things to remove or to add for the 12sp3 deprecation in non SES.

toabctl:
Great - our CI now just stops. That's why it is green...

dirkmueller:
@toabctl whenever a PR is being merged it rebuilds all open PRs and that seems to confuse github. it looks like this needs fixing...

PR #343, opened 2019-05-14
Freeze helm repositories to the same version that is used by airship (do not merge)

jsuchome:
Updating helm repo's references to the ones present in 

https://github.com/SUSE-Cloud/socok8s/blob/master/site/soc/software/config/versions.yaml

### DIFF
--- a/vars/manifest.yml
+++ b/vars/manifest.yml
@@ -2,11 +2,11 @@
 upstream_repos:
   openstack-helm-infra:
     src: https://opendev.org/openstack/openstack-helm-infra
-    version: master
+    version: 5e1ecd9840397bf9e8829ce0d98fcb721db1b74e
     dest_subdir: openstack
   openstack-helm:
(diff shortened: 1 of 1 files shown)


jsuchome:
Let's solve the way of applying patches elsewhere

jgu17:
A good start to get general alignment! Airship supports individual version/tag for each service, I don't think we'll need to do that for the osh development purpose, at least not yet.

jsuchome:
@jgu17 well, that means this change does _not_ affect airship, right? And as it is not needed for OSH, I agree with @evrardjp that this is unnecessary after all

jgu17:
> @jgu17 well, that means this change does not affect airship, right? And as it is not needed for OSH, I agree with @evrardjp that this is unnecessary after all

No this patch doesn't affect Airship at all. I thought you wanted to control osh side.

jsuchome:
Not needed for OSH development

PR #345, opened 2019-05-14
Revert "Introduction of documentation for technical preview."

sgrasley:
Reverts SUSE-Cloud/socok8s#338

Needs further refinement.


jgu17:
agree that we should revert and have the end user doc an dev doc separation completed and have only one doc source folder on the feature branch before merging to master

PR #347, opened 2019-05-14
Install helm from package (SOC-9002)

jsuchome:
Requires https://build.opensuse.org/request/show/703158

UPDATE: package is already there


playbooks/roles/common-deployer/templates/helm-serve.service.j2
```diff
@@ -5,7 +5,7 @@ After=network.target
 [Service]
 User={{ ansible_user }}
 Restart=always
-ExecStart=/usr/local/bin/helm serve
+ExecStart=/usr/bin/helm serve
```
Itxaka: Im wondering if we should just put the helm-serve service directly on the package so it can be used as is by other package users?

jsuchome: Oh, that was the plan. See https://build.opensuse.org/request/show/703661
Seems like I forgot to remove the template after all

Itxaka:
FYI I retriggered the jenkins job

jsuchome:
jenkins job will fail, updated package is not there yet

jsuchome:
Package was submitted, theoretically it should work now

jsuchome:
Though the job is green, it did not actually run the code...

jsuchome:
> Need to add the cloud repo in the kvm mechanism.

Is there a place for repo setup on kvm? Didn't we agree to document it instead?

PR #350, opened 2019-05-16
Extra informations about design


evrardjp:
To avoid the "Why did you do it this way" questions, here are
a few extra answers that deserve to be in the documentation.

### DIFF
--- a/doc/source/reference.rst
+++ b/doc/source/reference.rst
@@ -27,7 +27,9 @@ the SUSE Enterprise Storage (SES).
 
 It started as a series of shell scripts and ansible playbooks, choosing the
 simplest and fastest way to bring a test infrastructure for the upstream
-charts.
+charts.  It was easier to start with a shell script
+than writing a CLI in <insert language here>, mostly because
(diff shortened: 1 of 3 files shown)


doc/source/reference.rst
```diff
@@ -54,6 +59,45 @@ This also helps the story of running behind
...
+
+run.sh is a bash script, because it is a very commonly
+installed software on the 'localhost' node, independent
+of the distribution or of its version.
+It allows to install higher level requirements,
+like ansible, until we package socok8s differently.
+
+Each of the steps in `run.sh` are written in a way they represent a
+user facing feature. While what happens behind the scenes could
+change, the user interface is, in theory, stable.
```
jsuchome: It might be good to mention that some those those "user facing features" are actually intended for developers...

evrardjp: A developer is a user? I am afraid if I change this it will be more confusing.

jsuchome: Developer is only one kind of user. The actual end users of the product are not supposed to use developer facing commands. But anyway, you can ignore that.

PR #351, opened 2019-05-16
Create help usage in run script.

rajdharwadkar:
help will show options that is used by external user
for internal/dev, options are still supported but not displayed in help

./run.sh --help
deploy               -- Deploy airship
update_openstack     -- Update openstack deployment
add_compute          -- Add compute node
remove_compute       -- Remove compute node
clean_all            -- Delete all resources related with deployment
clean_openstack      -- Delete resources that are in openstack namespace

./run.sh test
Invalid option, Check --help for valid options

 ./run.sh
./run.sh --help

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -2,12 +2,35 @@
 
 set -o errexit
 
-if [[ "${SOCOK8S_DEVELOPER_MODE:-False}" == "True" ]]; then
-    set -x
+usage() {
+    echo "deploy                                         -- Deploy SUSE Containerized OpenStack."
+    echo "update_openstack                               -- Update OpenStack deployment."
+    echo "add_openstack_compute                          -- Add OpenStack compute node. Add compute node host(s) in inventory file."
+    echo "remove_openstack_compute <compute-node-name>   -- Remove OpenStack compute node. Provide compute node host name with option."
+    echo "remove_deployment                              -- Delete all resources related with deployment including images."
+}
+
+if [ -z "${1:-}" ]; then
+    echo "$0 -h, --help          -- Show help message."
+    exit 1
 fi
 
 scripts_absolute_dir="$( cd "$(dirname "$0")/script_library" ; pwd -P )"
 socok8s_absolute_dir="$( cd "$(dirname "$0")" ; pwd -P )"
+
+deployment_action=$1
+
+while true ; do
+    case "$deployment_action" in
+      -h|--help) usage ; exit 0 ;;
+       *) source ${scripts_absolute_dir}/pre-flight-checks.sh "validate_cli_options $1"
+          break ;;
+    esac
+done
+
+if [[ "${SOCOK8S_DEVELOPER_MODE:-False}" == "True" ]]; then
+    set -x
+fi
 
 # USE an env var to setup where to deploy to
 # by default, ccp will deploy on openstack for inception style fun (and CI).
@@ -31,11 +54,6 @@ pushd ${socok8s_absolute_dir}
 # For simplificity, the following script contains each action for a deploy mechanism, and each action should
 # contain a "master" playbook, which should be named playbooks/${DEPLOYMENT_MECHANISM}-${deployment_action}
 source ${scripts_absolute_dir}/deployment-actions-${DEPLOYMENT_MECHANISM}.sh
-
-# When automation is changed to introduce steps,
-# replace this line with the following line:
-# deployment_action=$1
-deployment_action=${1:-"setup_everything"}
 
 case "$deployment_action" in
     "deploy_network")
@@ -64,10 +82,14 @@ case "$deployment_action" in
     "deploy_osh")
         deploy_osh
         ;;
-    "add_compute")
+    "add_openstack_compute")
         add_compute
         ;;
-    "remove_compute")
+    "remove_openstack_compute")
+        if [ -z ${2+x} ];then
+            echo "Please enter compute node host name"
+            exit 1
+        fi
         remove_compute $2
         ;;
     "setup_caasp_workers_for_openstack")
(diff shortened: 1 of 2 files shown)


run.sh
```diff
@@ -2,6 +2,47 @@
 
 set -o errexit
 
+usage() {
+    echo "deploy               -- Deploy airship"
+    echo "update_openstack     -- Update openstack deployment"
+    echo "add_compute          -- Add compute node"
+    echo "remove_compute       -- Remove compute node"
+    echo "clean_all            -- Delete all resources related with deployment"
+    echo "clean_openstack      -- Delete resources that are in openstack namespace"
```
jgu17: I think we agreed that the purpose of clean_openstack is not clear, and implementation is half baked as the openstack site is still in Airship, and we only remove the pods etc. The user only needs "update OpenStack" is our conclusion. If we still hold the above opinion, I prefer we take out clean_openstack option.

run.sh
```diff
@@ -2,6 +2,47 @@
 
 set -o errexit
 
+usage() {
+    echo "deploy               -- Deploy airship"
```
jgu17: maybe "Deploy SUSE Containerized OpenStack"

run.sh
```diff
@@ -2,6 +2,47 @@
...
+    echo "update_openstack     -- Update openstack deployment"
```
jgu17: openstack->OpenStack.

run.sh
```diff
@@ -2,6 +2,47 @@
...
+    echo "add_compute          -- Add compute node"
```
jgu17: Add OpenStack compute node. I think more specific is better since the description is pretty short here.

run.sh
sgrasley: So, what is the preferred verb for removing deployment?  remove_deployment?

run.sh
jsuchome: I'm wondering, `add_compute` does not need any argument? So any data for a new compute node need to be provided in some extra place? 
Maybe it could be mentioned...

run.sh
rajdharwadkar: it wont be consistent rest of options? I think we either make all small case or all Mixed cases ?

run.sh
rajdharwadkar: > So, what is the preferred verb for removing deployment? remove_deployment?

I like remove_deployment than clean_all . but it does clean everything including images ?

run.sh
rajdharwadkar: > I'm wondering, `add_compute` does not need any argument? So any data for a new compute node need to be provided in some extra place?
> Maybe it could be mentioned...

Yes will add more info.

rajdharwadkar: > Add OpenStack compute node. I think more specific is better since the description is pretty short here.

Yes will fix it.

run.sh
jhesketh: I think James is referring to the documentation side. ie "Update OpenStack deployment".

run.sh
```diff
@@ -64,10 +104,14 @@ case "$deployment_action" in
     "deploy_osh")
         deploy_osh
         ;;
-    "add_compute")
+    "add_openstacK_compute")
```
jhesketh: Typo capitalisation.

run.sh
```diff
@@ -2,6 +2,46 @@
...
+action=$1
+validate_options(){
+   isvalid=false
+   for value in ${OPTIONS[@]} ; do
+      if [ $value == $action ]; then
+         isvalid=true
+         break;
+      fi
+   done
+   if [ $isvalid = false ]; then
```
evrardjp:
```
[[ "$isvalid" == "false" ]];
``` ? (note the `==`).

File: run.sh [hunk 7]
```diff
@@ -2,6 +2,46 @@
...
+   return 0
+}
+
+while true ; do
+    case "$action" in
+      -h|--help) usage ; exit 0 ;;
+       *) validate_options;  break ;;
+    esac
+done
+
```
evrardjp: Does it makes sense to keep L78 then?

run.sh
```diff
@@ -2,6 +2,46 @@
...
+
```
evrardjp: nit: I understand why we are doing this early, before any other logic, but this is harder to read than if the functions were in a separate scripts, sourced. For example: source ${scripts_absolute_dir}/pre-flight-checks.sh validate_cli_arguments

run.sh
rajdharwadkar: > I think James is referring to the documentation side. ie "Update OpenStack deployment".

Yes its already fixed.

run.sh
```diff
@@ -2,6 +2,46 @@
 
 set -o errexit
 
+usage() {
+    echo "deploy                                         -- Deploy SUSE Containerized OpenStack."
+    echo "update_openstack                               -- Update OpenStack deployment."
+    echo "add_openstack_compute                          -- Add OpenStack compute node. User need to provide compute node name in inventory file."
```
jgu17: need->needs?   "add compute node host(s) in the inventory file" -- nits

run.sh
```diff
@@ -2,6 +2,46 @@
...
+    echo "remove_openstack_compute <compute-node-name>   -- Remove OpenStack compute node. User need to provide compute node name with option."
```
jgu17: Can we request confirmation before removing compute or remove deployment?

rajdharwadkar: For cli I think non-interactive is better because they can automate if required.  From UI they can have that option ?

run.sh
rajdharwadkar: will fix it.

run.sh
rajdharwadkar: yes we dont need it . will fix it.

run.sh
rajdharwadkar: Done

run.sh
rajdharwadkar: for option it should be small case.

jesusaurus: `openstacK` -> `openstack`

Notice the last letter

rajdharwadkar: Yes done

PR #353, opened 2019-05-17
Increase test_timeout value and move default values into role defaults

matthew-fuller:
This change increases the default tiller test_timeout value from 300s to 1800s, and it moves the default values for both test_timeout and run_tests into the role's defaults (which is where they probably should have been in the first place). The default value of 300s didn't seem to be enough for anyone's environment, so rather than make the user increase the value in extravars before deploying for the first time, it's better to increase the default.

### DIFF
--- a/playbooks/roles/airship-deploy-ucp/defaults/main.yml
+++ b/playbooks/roles/airship-deploy-ucp/defaults/main.yml
@@ -1,3 +1,5 @@
 ---
 scale_profile: "{{ lookup('env','scale_profile') | default('minimal', true) }}"
 airship_pegleg_image: "{{ suse_airship_registry_location }}/airshipit/pegleg:{{ suse_airship_components_image_tag['pegleg'] }}"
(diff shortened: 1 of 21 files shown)


playbooks/roles/airship-deploy-ucp/defaults/main.yml
```diff
@@ -2,3 +2,5 @@
 scale_profile: "{{ lookup('env','scale_profile') | default('minimal', true) }}"
 airship_pegleg_image: "{{ suse_airship_registry_location }}/airshipit/pegleg:{{ suse_airship_components_image_tag }}"
 suse_osh_image_version: "pike"
+run_tests: true
+test_timeout: 1800
```
jgu17: can we please increase the default timeout to 2700? For the Nova chart, we changed the upstream default to 2700. Since we use a global value for all charts, we should pick the largest value of all chart.

PR #356, opened 2019-05-20
Jenkinsfile.integration: Add deployment parameter

toabctl:
We want to set a default deployment mechanism. Otherwise we do not run
all the needed steps to do a full deployment.
Also rename the parameter value from "helm" to "osh" to make it clear
that we use the "osh" deployment step.

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -11,6 +11,12 @@ pipeline {
             label "cloud-ccp-ci"
         }
     }
+
+    parameters {
+        /* first value in the list is the default */
+        choice(choices: ['airship', 'osh'], description: 'Which deployment mechanism?', name: 'deployment')
+    }
+
     environment {
         /* use lowercase SOCOK8S_ENVNAME. CaaSP Velum doesn't like it otherwise */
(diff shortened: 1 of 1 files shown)


evrardjp:
Blindly approved.

Jenkinsfile.integration
```diff
@@ -11,6 +11,11 @@ pipeline {
             label "cloud-ccp-ci"
         }
     }
+
+    parameters {
+        choice(choices: ['osh', 'airship'], defaultValue: 'osh', description: 'Which deployment mechanism?', name: 'deployment')
```
jsuchome: Didn't we want `airship` as default deployment path?

toabctl: Not sure how the discussion ended. I agree to set 'airship' as default. But afaik, `airship' is broken...

toabctl: Done

jsuchome:
@toabctl what about those caasp4 parts? That's unrelated, is it some test project of yours?

toabctl:
> @toabctl what about those caasp4 parts? That's unrelated, is it some test project of yours?

@jsuchome yes. that was from some experiment. removed.

PR #361, opened 2019-05-21
Update command to retrieve admin key in base 64

sgrasley:
Original command was not creating correct base64 key.

### DIFF
--- a/doc/source_tech_preview/deployment/configure.rst
+++ b/doc/source_tech_preview/deployment/configure.rst
@@ -118,7 +118,7 @@ The Ceph admin keyring can be obtained by running the following on ceph host.
 
 .. code-block:: yaml
 
-  sudo ceph auth get-key client.admin | base64
+  echo $( sudo ceph auth get-key client.admin ) | base64
 
 For example:
 


doc/source_tech_preview/deployment/configure.rst
```diff
@@ -118,7 +118,7 @@ The Ceph admin keyring can be obtained by running the following on ceph host.
 
 .. code-block:: yaml
 
-  sudo ceph auth get-key client.admin | base64
+  echo $( sudo ceph auth get-key client.admin ) | base64
```
jesusaurus: I think this could be further improved by using `base64 <(sudo ceph auth get-key client.admin)`

jgu17: could you explain advantage over "echo"?

sgrasley: adds a \n to string that is getting base64 encoded.  had issues authenticating with ses without /n.

jgu17: sorry the question was for @jesusaurus.

jesusaurus: It's not a very important difference, but it spawns fewer processes/subshells by giving the base64 process a direct file handle for the ceph output.

csymons-suse:
superseded by https://github.com/SUSE-Cloud/socok8s/pull/383
change transferred to doc/source

PR #362, opened 2019-05-21
overiding volume size for all rabbitmq in openstack and postgresql in ucp

rajdharwadkar:
Volume size is configurable and default is 5Gi

QE ran into issue after creating vms.
rabbitmq and postgresql pods were in CrashLoopBackOff state.

### DIFF
--- a/site/soc/software/charts/ucp/core/postgresql.yaml
+++ b/site/soc/software/charts/ucp/core/postgresql.yaml
@@ -31,7 +31,7 @@ data:
   values:
     storage:
       pvc:
-        size: 2Gi
+        size: {{ db_volume_size }}
     pod:
       replicas:
         server: 1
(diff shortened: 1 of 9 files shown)


site/soc/software/charts/ucp/core/postgresql.yaml
```diff
@@ -31,7 +31,7 @@ data:
   values:
     storage:
       pvc:
-        size: 2Gi
+        size: {{ volume_size }}
```
jgu17: should we have a different size var for postgres vs RabbitMQ?

rajdharwadkar: because of different namespaces ?  or each service will have different var for size ?

jgu17: sorry I should be more clear. I think db and rabbitmq are different apps, I don't why they would always have the same storage size .

rajdharwadkar: ok fixed it.

PR #368, opened 2019-05-22
Do not pass empty variable to useradd

jsuchome:
If USER variable is not defined, ansible fails with

  useradd: invalid user name ''

### DIFF
--- a/playbooks/roles/airship-setup-deployer/tasks/main.yml
+++ b/playbooks/roles/airship-setup-deployer/tasks/main.yml
@@ -37,12 +37,17 @@
   tags:
     - install
 
+- name: Set the user name
+  set_fact:
+    user_name: "{{ lookup('env','USER') }}"
+
 - name: Add the user to the group 'docker'
   become: yes
   user:
-    name: "{{ lookup('env','USER') }}"
+    name: user_name
     groups: docker
     append: yes
+  when: user_name
 
 - name: Ensure kubectl is properly configured
   include_role:


jsuchome:
See e.g. https://ci.suse.de/blue/organizations/jenkins/cloud-socok8s-jobs%2Fsocok8s-integration/detail/PR-367/1/pipeline

I'm not sure if the syntax is correct...

jsuchome:
So the pr-merge job fails, but at least it got past the original error I think

playbooks/roles/airship-setup-deployer/tasks/main.yml
```diff
@@ -37,12 +37,17 @@
...
+    user_name: "{{ lookup('env','USER') }}"
+
 - name: Add the user to the group 'docker'
   become: yes
   user:
-    name: "{{ lookup('env','USER') }}"
+    name: user_name
     groups: docker
     append: yes
+  when: user_name
```
jesusaurus: Instead of skipping this step if USER is unset, we should make sure that USER is set to the appropriate value during the jenkins job

jesusaurus:
I've proposed #372 as an alternative approach

playbooks/roles/airship-setup-deployer/tasks/main.yml
jhesketh: I agree

playbooks/roles/airship-setup-deployer/tasks/main.yml
```diff
@@ -37,12 +37,17 @@
   tags:
     - install
 
+- name: Set the user name
+  set_fact:
+    user_name: "{{ lookup('env','USER') }}"
```
jhesketh: (nit) Any particular reason for user_name as opposed to username which afaik is more commonly accepted?

PR #370, opened 2019-05-22
drop some sphinx reqs

Itxaka:
In order to package the repo and build the docs we need to drop some
of the sphinx requirements for building the diagrams as they are not
available for opensuse and it would require us to package and maintain
those ourselves

Instead just use plain images for those diagrams that require extra
deps and provide the original diag files to generate those images
so we can make changes to them in the future


toabctl:
packages are there now in devel:languages:python

evrardjp:
I would add a comment that it's used for generation, but that's fine for me like this.

Itxaka:
woooo, closing this thanks to @toabctl !

PR #371, opened 2019-05-22
 remove sphinx option not available on leap versions

Itxaka:
leap alabaster theme version is a bit old (0.7.10) and it does not
support the show_relbars config as that appears on 0.7.11

Instead of silently ignore unkown parameters, alabaster will instead
die, which is pretty bad.

So we need to drop that option to be able to package docs for leap
15.0/15.1


evrardjp:
> oh no - I really liked the prev/next buttons...

Me too! And you can see yourself progressing in the deploy and all of that!

I was so proud of this docs!

PR #372, opened 2019-05-22
Make sure USER is set in jenkins

jesusaurus:
For some unknown reason, there is no USER variable in the test
environment spawned by jenkins. Manually set the USER variable to the
appropriate value in the jenkins job. This is needed by the
airship-setup-deployer role.

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -26,6 +26,7 @@ pipeline {
         SOCOK8S_DEVELOPER_MODE = "True"
         DEPLOYMENT_MECHANISM = "openstack"
         ANSIBLE_STDOUT_CALLBACK = "yaml"
+        USER = "jenkins" /* Why isn't this set in the jenkins environment? */
     }
 
     stages {


Jenkinsfile.integration
```diff
@@ -26,6 +26,7 @@ pipeline {
         SOCOK8S_DEVELOPER_MODE = "True"
         DEPLOYMENT_MECHANISM = "openstack"
         ANSIBLE_STDOUT_CALLBACK = "yaml"
+        USER = "jenkins" /* Why isn't this set in the jenkins environment? */
```
dirkmueller: are you sure that a trailing "/*" comment is actually going to work? as far as I remember the parsing of this is weird.

jesusaurus: I didn't know that jenkins and/or groovy have buggy comment parsing, is that documented somewhere?

toabctl:
According to https://wiki.jenkins.io/display/JENKINS/Building+a+software+project#Buildingasoftwareproject-belowJenkinsSetEnvironmentVariables , ```$USER``` is not set by intention

Jenkinsfile.integration
toabctl: According to CI, this seems to work fine

PR #374, opened 2019-05-22
Creating pools for airship usage

kantarun:
In openstack-helm charts, some of pools expects a specific application
name which are not same as what's created by SES openstack.integrate
script. So for interim, we are creating pools with different names
so that osh charts can enable the hard-coded application name.

We plan to change upstream to have application name parametrized as
pool name but it will take some time. So this is just added for
tech preview only.


playbooks/roles/common-ses/tasks/main.yml
```diff
@@ -0,0 +1,38 @@
+---
+- name: Include global vars
+  include_vars: "{{ playbook_dir }}/../vars/common-vars.yml"
+
+- name: Create pools in SES for socok8s usage
+  become: yes
+  include_role:
+    name: ses
+    tasks_from: openstack_config
```
jgu17: is this task idempotent? Since it doesn't look like we added code to delete the roles when uninstall, would redeploy still work? I admit that I haven't checked the openstack_config code.

playbooks/roles/common-ses/tasks/main.yml
```diff
@@ -0,0 +1,38 @@
...
+      - "{{ ses_openstack_glance_pool }}"
+      - "{{ ses_openstack_cinder_pool }}"
+      - "{{ ses_openstack_nova_pool }}"
+      - "{{ ses_openstack_cinder_backup_pool }}"
+      - "{{ ses_osh_rbd_pool }}"
+
+- name: Import SES configs to deployer
+  become: yes
+  include_role:
+    name: ses-osh
```
jgu17: eventually, we should have the code in the commons, and have ses-osh to import from commons, since ses-osh is dev mechanism, IMO.

playbooks/roles/ses-osh/defaults/main.yml
```diff
@@ -1,2 +1,18 @@
 ---
 radosgw_keystone: False
+
+# Openstack-helm services expects a specific application name which is
+# different from what's created by ses openstack.integrate script.
+# So as interim solution, creating pools with different names to allow
+# exected application names to be enabled with those pools.
+
+socok8s_openstack_glance_pool_name: socok8s_images
```
jgu17: doesn't have to be in this commit, but it'd be nice to have pool names tied to the workspace env variable. if deploy socok8s to a different workspace, I think would be a good idea that the ses pools are also separated.

playbooks/roles/common-ses/tasks/main.yml
kantarun: Yes, it is as I have ran this multiple times. create pool does fail if it already exists but ansible ignores that with failed_when: false logic.

playbooks/roles/common-ses/tasks/main.yml
kantarun: Yes, makes sense. So will push update to PR to follow this usage model. Actually I can remove usage of ses-osh role altogether and directly invoke play from role common-ses.

playbooks/roles/ses-osh/defaults/main.yml
kantarun: Not sure about the need of having pool name tied to workspace name. 
Purpose for this change is to have different pool name from integrate script which having socok8s_ prefix provides that. Actually having different but fixed name provides predictability in documentation and usage/debugging perspective like we have with ucp and openstack namespaces even though they can be different if needed.

jgu17:
> 
> 
> Can more than one ccp deployment be installed on shared CEPH environment?

yes

hemna:
> Can more than one ccp deployment be installed on shared CEPH environment?

You can provide a prefix to the openstack integration runner with SES and you can get separate pools with that prefix as part of the name of the pools.

playbooks/roles/common-ses/defaults/main.yml
```diff
@@ -0,0 +1,18 @@
...
+socok8s_openstack_glance_pool_name: socok8s_images
+socok8s_openstack_glance_pool_user: glance
+socok8s_openstack_cinder_pool_name: socok8s_volumes
+socok8s_openstack_cinder_pool_user: cinder
+socok8s_openstack_nova_pool_name: socok8s_nova
+socok8s_openstack_libvirt_pool_name: socok8s_vms
+socok8s_openstack_libvirt_pool_user: cinder
+socok8s_openstack_cinder_backup_pool_name: socok8s_cinder_backup
+socok8s_openstack_cinder_backup_pool_user: cinder-backup
+socok8s_default_rbd_pool_name: socok8s_rbd
```
jgu17: I still prefer the os integration salt runner's approach to allow a customizable prefix to make it unique among deployments. In the QE environment at least, SES are shared. It could be true in customer's environment too.  We can make the prefix default to the socok8s workspace prefix, and the user can override it if they choose so.

kantarun:
> You might have a (our first ever!?) legitimate jenkins failure here:
> 
> ```
> [2019-05-23T15:35:32.420Z] TASK [Generate ses-config.yml] *************************************************
> 
> [2019-05-23T15:35:32.420Z] fatal: [cloud-socok8s-pr-374-9-ses -> localhost]: FAILED! => changed=false 
> 
> [2019-05-23T15:35:32.420Z]   msg: |-
> 
> [2019-05-23T15:35:32.420Z]     Could not find or access 'ses-config.yml.j2'
> 
> [2019-05-23T15:35:32.420Z]     Searched in:
> 
> [2019-05-23T15:35:32.420Z]             /home/jenkins/workspace/-jobs_socok8s-integration_PR-374/playbooks/roles/common-ses/tasks/templates/ses-config.yml.j2
> 
> [2019-05-23T15:35:32.420Z]             /home/jenkins/workspace/-jobs_socok8s-integration_PR-374/playbooks/roles/common-ses/tasks/ses-config.yml.j2
> 
> [2019-05-23T15:35:32.420Z]             /home/jenkins/workspace/-jobs_socok8s-integration_PR-374/playbooks/templates/ses-config.yml.j2
> 
> [2019-05-23T15:35:32.420Z]             /home/jenkins/workspace/-jobs_socok8s-integration_PR-374/playbooks/ses-config.yml.j2 on the Ansible Controller.
> 
> [2019-05-23T15:35:32.420Z]     If you are using a module and expect the file to exist on the remote, see the remote_src option
> 
> [2019-05-23T15:35:32.420Z] 	to retry, use: --limit @/home/jenkins/workspace/-jobs_socok8s-integration_PR-374/playbooks/openstack-deploy_ses.retry
> ```
> 
> It seems to be looking for the template in `playbooks/roles/common-ses/tasks/templates` or `playbooks/roles/common-ses/tasks/` but **not** `playbooks/roles/common-ses/templates`.

Indeed, file path needs to be absolute as its invoked from 2 different point of entries. This issue is addressed in subsequent change in PR and now passing check job.

jsuchome:
@kantarun you can already use different app name, the patches were merged upstream https://review.opendev.org/#/c/661722/

playbooks/roles/common-ses/tasks/main.yml
kantarun: The requested change was made in susequent PR push so resolving this now

playbooks/roles/ses-osh/defaults/main.yml
kantarun: Added support for prefix but for now, keeping the same names as used previously to have minimal changes for TP release deadline.

kantarun:
> @kantarun you can already use different app name, the patches were merged upstream https://review.opendev.org/#/c/661722/

Yes, noticed that both upstream reviews were merged and have updated PR yesterday to reflect that.

PR #376, opened 2019-05-22
Add playbook and role for deploying tempest helm chart using airship shipyard workflow

matthew-fuller:
This change is necessary to accommodate the setup tasks required by the tempest helm chart that setup networks and obtain image, flavor and network IDs to pass in as value overrides. 

After rendering the templates, an openstack-tempest directory is created in {clone_folder}/airship/treasuremap/site/soc/software/charts/osh and all appropriate files are moved there. A new task was also added to reprocess the software/manifests/full-site.yaml template so that tempest can be added to the deployment manifest. Once all files are in place, the airship-deploy-osh role is called with the update_airship_osh_site tag and the update_software workflow is run to deploy tempest into the cluster.

### DIFF
--- a/doc/source/deployment/verify.rst
+++ b/doc/source/deployment/verify.rst
@@ -17,3 +17,143 @@ services are functioning as expected by running the following commands:
    export OS_CLOUD='openstack'
    openstack endpoint list
    openstack server list
+
+OpenStack Tempest Testing
+=========================
+
+After the deployment of SUSE Containerized OpenStack has completed, it is possible to run
+OpenStack Tempest tests against the core services in the deployment using the `run.sh` script.
+Before running Tempest tests, it will be necessary to manually configure OpenStack network
+resources and provide a few configuration parameters in the ${WORKDIR}/env/extravars file.
+
+Setting Up An External Network And Subnet in OpenStack
+------------------------------------------------------
+
+To set up an external network and subnet in OpenStack, the following commands can be run from a
+shell on the `Deployer` node.
+
+.. code-block:: console
+
+   export OS_CLOUD=openstack
+   openstack network create --provider-network-type flat --provider-physical-network public \ 
+     --external public
+   openstack subnet create --network public --subnet-range 192.168.100.0/24 --allocation-pool \
+     start=192.168.100.10,end=192.168.100.200 --gateway 192.168.100.1 --no-dhcp public-subnet
+
+.. note::
+
+   The external public network is expected to be able to reach the internet. The above values 
+   will vary based on your network environment. 
+
+Once the public network and subnet have been created in OpenStack, their names will need to be
+made known to Tempest by adding the following keys in the ${WORKDIR}/env/extravars file:
+
+.. code-block:: yaml
+
+   openstack_external_network_name: "public"
+   openstack_external_subnet_name: "public-subnet"
+
+Tempest will also need to know the CIDR block from which to allocate project IPv4 subnets. This
+value should be specified with the following key in the extravars file:
+
+.. code-block:: yaml
+
+   openstack_project_network_cidr: "192.0.4.0/24"
+
+Configuring Tempest Test Parameters
+-----------------------------------
+
+By default, the implementation of Tempest in SUSE Containerized OpenStack will run smoke tests
+for all deployed services including compute, identity, image, network, and volume, using 4
+workers. 
+
+To modify the number of workers, add the following key with a value of your choosing to the
+extravars file:
+
+.. code-block:: yaml
+
+   tempest_workers: 6
+
+To disable tests for specific OpenStack components, any or all of the following keys can be
+added to the extravars file:
+
+.. code-block:: yaml
+
+   tempest_enable_cinder_service: false
+   tempest_enable_glance_service: false
+   tempest_enable_nova_service: false
+   tempest_enable_neutron_service: false
+
+To run all Tempest tests instead of just smoke tests, add the following key to the extravars
+file:
+
+.. code-block:: yaml
+
+   tempest_test_type: "all"
+
+Running Tempest Tests
+---------------------
+
+Once all of the OpenStack network resources have been created and all configuration parameters have
+been provided in ${WORKDIR}/env/extravars, Tempest testing can be started by running the following
+command from the root of the socok8s directory:
+
+.. code-block:: console
+
+   ./run.sh test
+
+Once the Tempest pods have been deployed, testing will begin immediately. You can check the progress
+of the test pod at any time by running
+
+.. code-block:: console
+
+   kubectl get pods -n openstack | grep tempest-run
+
+Example output:
+
+.. code-block:: console
+
+   airship-tempest-run-tests-hq6jg                          1/1     Running       0          33m
+
+A status of 'Running' indicates that testing is still in progress. Once testing is complete, the status
+of the airship-tempest-run-tests pod will change to 'Complete', indicating that all tests passed, or
+'Error', indicating that at least one test has failed.
+
+Tempest Test Results
+--------------------
+
+All test results can be viewed by retrieving the logs from the airship-tempest-run-tests pod by running
+the following command:
+
+.. code-block:: console
+
+   kubectl logs -n openstack airship-tempest-run-tests-hq6jg
+
+.. note::
+
+   The logs can be viewed at any time, even while a current test batch is still running. 
+
+Once testing is complete, the logs will conclude with a summary of all passed, skipped, and failed tests
+similar to the following:
+
+.. code-block:: console
+
+   ======
+   Totals
+   ======
+   Ran: 78 tests in 104.0000 sec.
+    - Passed: 62
+    - Skipped: 16
+    - Expected Fail: 0
+    - Unexpected Success: 0
+    - Failed: 0
+   Sum of execute time for each test: 56.3147 sec.
+
+   ==============
+   Worker Balance
+   ==============
+    - Worker 0 (19 tests) => 0:01:44.140828
+    - Worker 1 (20 tests) => 0:01:02.484599
+    - Worker 2 (18 tests) => 0:00:29.100245
+    - Worker 3 (21 tests) => 0:01:28.449495
+   
--- a/playbooks/roles/airship-deploy-tempest/defaults/main.yml
+++ b/playbooks/roles/airship-deploy-tempest/defaults/main.yml
@@ -1,0 +1,14 @@
+---
+deploy_tempest: true
+tempest_test_type: "smoke"
+tempest_test_args:
+  smoke:  "--smoke"
+  all: "--regex '.*'"
+tempest_workers: 4
+tempest_enable_cinder_service: true
+tempest_enable_glance_service: true
+tempest_enable_nova_service: true
+tempest_enable_neutron_service: true
+openstack_external_network_name: "public"
+openstack_external_subnet_name: "public-subnet"
+openstack_project_network_cidr: "172.0.4.0/24"
--- a/playbooks/roles/airship-deploy-tempest/tasks/main.yml
+++ b/playbooks/roles/airship-deploy-tempest/tasks/main.yml
@@ -1,0 +1,122 @@
+---
+# tasks file for airship-deploy-tempest
+- name: Load standard variables
+  include_vars: "{{ playbook_dir }}/../vars/common-vars.yml"
+  tags:
+    - always
+
+- name: Set site path
+  set_fact:
+    site_path: "{{ upstream_repos_clone_folder }}/airship/treasuremap/site/{{ socok8s_site_name }}"
+  tags:
+    - always
+
+- name: Ensure site directory for {{ socok8s_site_name }} exists
+  become: yes
+  file:
+    path: "{{ site_path }}"
+    state: directory
+
+- name: Ensure tempest chart directory exists
+  become: yes
+  file:
+    path: "{{ site_path }}/software/charts/osh/openstack-tempest"
+    state: directory
+
+- name: Get external network ID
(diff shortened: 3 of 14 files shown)


playbooks/roles/airship-deploy-tempest/defaults/main.yml
```diff
@@ -0,0 +1,7 @@
+---
+deploy_tempest: true
+tempest_enable_cinder_service: true
+tempest_enable_glance_service: true
+tempest_enable_nova_service: false
```
jgu17: Nova and Neutron are still not supported? Is OSH team aware of this and a ticket filed?

playbooks/roles/airship-deploy-tempest/templates/tempest.yaml.j2
```diff
@@ -0,0 +1,100 @@
+---
```
jgu17: why aren't this chart yaml in the soc site directory?

run.sh
```diff
@@ -123,6 +123,9 @@ case "$deployment_action" in
     "gather_logs")
```
jgu17: please add the option in the deployment guide.

playbooks/roles/airship-deploy-tempest/files/tempest_setup_network.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+set -xe
+
+export OS_CLOUD=openstack
+
+export OSH_EXT_NET_NAME="public"
```
jgu17: does the ext network have to actually exist on the host for the test to succeed? If true, do we need to make this user input in extravars?

playbooks/roles/airship-deploy-tempest/defaults/main.yml
matthew-fuller: Nova and neutron are disabled by default in the upstream chart, so I just left it that way in this initial commit because I know tempest deploys and runs the image and volume tests successfully. I'm currently re-testing with them enabled and will change the defaults if everything checks out.

playbooks/roles/airship-deploy-tempest/templates/tempest.yaml.j2
matthew-fuller: I'll move this chart there.

mbelur:
I still think that tempest should not be creating the public network resource.
This should be done outside of tempest since this unique to customer's environment.

For example:
In ardana based cloud install, we have neutron-cloud-configure.yml playbook (called by ardana-cloud-configure.yml) that creates these resources based on the values provided in the input model.

Tempest does not create these resources. It just checks for the existence and reference them if they exist.
If such a resource does not exist, then it fails as in:

- name: tempest | test-resources | Check and fail if there is no network set-up
  fail:
    msg: "No {{ tempest_network_name }} found. Please create one"
  when: (tempest_available.neutron | bool) and
        (public_network_id_result | failed)


I would expect a similar approach in CCP environments too.

playbooks/roles/airship-deploy-tempest/tasks/main.yml
```diff
@@ -0,0 +1,116 @@
...
+
+- name: Ensure tempest chart directory exists
+  become: yes
+  file:
+    path: "{{ site_path }}/software/charts/osh/openstack-tempest"
+    state: directory
+
+- name: Determine tempest test type (smoke or all)
+  set_fact:
+    test_args: "{% if tempest_test_type == 'all' %}--regex 'tempest.api.*'{% else %}--smoke{% endif %}"
```
jesusaurus: This appears to be needed during `TASK [airship-deploy-ucp : Create site software manifests]`, is there somewhere we can put this so that the `airship-deploy-ucp` role can access it as well?

matthew-fuller: Hmm...it shouldn't be. The entire 'values' section of the tempest template should get ignored during airship-deploy-ucp. Is 'deploy_tempest: true' set anywhere in your env? I was able to pass that stage during my 3 test runs.

run.sh
```diff
@@ -5,6 +5,7 @@ set -o errexit
 usage() {
     echo "deploy                                         -- Deploy SUSE Containerized OpenStack."
     echo "update_openstack                               -- Update OpenStack deployment."
+    echo "test                                           -- Deploy and run Tempest tests."
```
jgu17: ->Deploy and run OpenStack Tempest tests.

site/soc/software/charts/osh/openstack-tempest/tempest.yaml
```diff
@@ -0,0 +1,100 @@
...
+  storagePolicy: cleartext
+data:
+  wait:
+    timeout: 3000
+{% if deploy_tempest is defined and deploy_tempest is sameas true %}
+  values:
+    # Adding images overrides here instead of versions.yaml
+    images:
+      tags:
+        tempest_run_tests: "{{ suse_osh_registry_location }}/openstackhelm/tempest:{{ suse_infra_image_version }}"
```
jgu17: the pattern is to define the image versions in the versions.yaml and use substitution to override the default value in the global chart.

site/soc/software/charts/osh/openstack-tempest/tempest.yaml
```diff
@@ -0,0 +1,100 @@
...
+    parentSelector:
+      name: tempest-global
+      component: tempest
+    actions:
+      - method: merge
+        path: .
+  storagePolicy: cleartext
+data:
+  wait:
+    timeout: 3000
```
jgu17: is 3000 sufficient to run full test?

site/soc/software/charts/osh/openstack-tempest/tempest.yaml
```diff
@@ -0,0 +1,100 @@
...
+          default: /v3
+        scheme:
+          default: http
+        port:
+          api:
+            default: 80
+            internal: 5000
+{% endif %}  
+  source:
+    type: git
```
jgu17: chart source and version should also be defined in the versions.yaml and use substitution to override here.

should not use master version, pin it down to a commit sha.

site/soc/software/charts/osh/openstack-tempest/tempest.yaml
```diff
@@ -0,0 +1,100 @@
...
+          network_for_ssh: "{{ openstack_external_network_name }}"
+          floating_ip_range: "{{ openstack_external_subnet_cidr.stdout }}"
+        service_available:
+          cinder: "{{ tempest_enable_cinder_service }}"
+          glance: "{{ tempest_enable_glance_service }}"
+          nova: "{{ tempest_enable_nova_service }}"
+          neutron: "{{ tempest_enable_neutron_service }}"
+    endpoints:
+      cluster_domain_suffix: cluster.local
+      local_image_registry:
```
jgu17: what is the local image registry used for?

site/soc/software/charts/osh/openstack-tempest/tempest.yaml
```diff
@@ -0,0 +1,100 @@
...
+            project_domain_name: default
+          tempest:
+            role: admin
+            region_name: RegionOne
+            username: tempest
+            password: {{ lookup('password', secrets_location + '/osh_tempest_password ' + password_opts) }}
+            project_name: service
+            user_domain_name: service
+            project_domain_name: service
+        hosts:
```
jgu17: much of these should use substation from the endpoints.yaml

playbooks/roles/airship-deploy-tempest/tasks/main.yml
flaviodsr: What about scenario tests, shouldn't they be part of `all`?

In my option `test_args` could be placed at `defaults/main.yml` like this:

```
tempest_test_args:
  smoke:  "--smoke"
  all: "--regex .*"
```

and on `tempest.yaml`:
```
...
script: "tempest run --config-file /etc/tempest/tempest.conf -w 4 {{ tempest_test_args[tempest_test_type] }}"
...
```

site/soc/software/charts/osh/openstack-tempest/tempest.yaml
```diff
@@ -0,0 +1,100 @@
...
+    timeout: 3000
+{% if deploy_tempest is defined and deploy_tempest is sameas true %}
+  values:
+    # Adding images overrides here instead of versions.yaml
+    images:
+      tags:
+        tempest_run_tests: "{{ suse_osh_registry_location }}/openstackhelm/tempest:{{ suse_infra_image_version }}"
+        ks_user: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_openstack_image_version }}"
+    conf:
+      script: "tempest run --config-file /etc/tempest/tempest.conf -w 4 {{ test_args }}"
```
flaviodsr: Could we have a variable for the number of workers? so we could override it if needed.

playbooks/roles/airship-deploy-tempest/tasks/main.yml
matthew-fuller: Excellent idea, I've made that change.

site/soc/software/charts/osh/openstack-tempest/tempest.yaml
matthew-fuller: I created a var in defaults called tempest_workers and set the default value to 4. It can be overridden in extravars if needed.

site/soc/software/charts/osh/openstack-tempest/tempest.yaml
matthew-fuller: Mine finished, but I hadn't included any of the scenario tests as Flavio pointed out below. I increased it to 6000. Should this be enough?

site/soc/software/charts/osh/openstack-tempest/tempest.yaml
matthew-fuller: I pinned the reference to a commit sha, and moved chart and image info to versions.yml. I think I originally put this here while testing and forgot to move it out again.

site/soc/software/charts/osh/openstack-tempest/tempest.yaml
matthew-fuller: Done.

matthew-fuller:
I believe I've incorporated all suggestions and made the requested changes. I've also added documentation in the deployment/verify guide.

doc/source/deployment/verify.rst
```diff
@@ -17,3 +17,143 @@ services are functioning as expected by running the following commands:
...
+------------------------------------------------------
+
+To set up an external network and subnet in OpenStack, the following commands can be run from a
+shell on the `Deployer` node.
+
+.. code-block:: console
+
+   export OS_CLOUD=openstack
+   openstack network create --provider-network-type flat --provider-physical-network public \ 
+     --external ext-net
```
mbelur: Instead of ext-net, match the example command with public which is the default value.
openstack_external_network_name: "public"

doc/source/deployment/verify.rst
```diff
@@ -17,3 +17,143 @@ services are functioning as expected by running the following commands:
...
+   openstack subnet create --network ext-net --subnet-range 192.168.100.0/24 --allocation-pool \
+     start=192.168.100.10,end=192.168.100.200 --gateway 192.168.100.1 --no-dhcp ext-subnet
```
mbelur: Same here we can match the example command to the name public-subnet which is default instead of the ext-subnet.

run.sh
```diff
@@ -155,6 +156,9 @@ case "$deployment_action" in
     "gather_logs")
         gather_logs
         ;;
+    "test")
+        deploy_tempest
```
jesusaurus: This function was only added to deployment-actions-kvm.sh, we also want one in deployment-actions-openstack.sh

playbooks/roles/airship-deploy-tempest/tasks/main.yml
```diff
@@ -0,0 +1,110 @@
...
+  register: subnet_pool_result
+  failed_when:
+    - subnet_pool_result.rc != 0
+    - "'not found' not in subnet_pool_result.stderr"
+  changed_when: false
+  tags:
+    - skip_ansible_lint
+
+- name: Setup subnet pool
+  shell: "{{ role_path }}/files/tempest_setup_subnet.sh"
```
jesusaurus: This is the path to the roles directory on localhost (where run.sh is run from); this will fail if localhost is not the deployer host:

```
TASK [airship-deploy-tempest : Setup subnet pool] *****************************************************************************************************************************************************************************************************************************
fatal: [jharker-beefy-deployer]: FAILED! => changed=true 
  cmd: /home/opensuse/socok8s/playbooks/roles/airship-deploy-tempest/files/tempest_setup_subnet.sh
  delta: '0:00:00.005243'
  end: '2019-06-03 21:08:14.130153'
  msg: non-zero return code
  rc: 127
  start: '2019-06-03 21:08:14.124910'
  stderr: '/bin/sh: /home/opensuse/socok8s/playbooks/roles/airship-deploy-tempest/files/tempest_setup_subnet.sh: No such file or directory'
  stderr_lines:
  - '/bin/sh: /home/opensuse/socok8s/playbooks/roles/airship-deploy-tempest/files/tempest_setup_subnet.sh: No such file or directory'
  stdout: ''
  stdout_lines: <omitted>
```

jgu17: I agree with the request, but it should wait until the tempest tests can all pass (after blacklisting), otherwise, CI will suffer as a result.

jesusaurus: @jgu17 it looks like we need to run an extra action (`run.sh test`) to trigger this code path, so this won't affect CI until we add the test action to the Jenkinsfile

PR #377, opened 2019-05-23
Only install requirements from PyPI when explicitly set

jhesketh:
Requirements should come from packages unless in developer mode or the user has explicitly set SOCOK8S_USE_VIRTUALENV.

### DIFF
--- a/script_library/pre-flight-checks.sh
+++ b/script_library/pre-flight-checks.sh
@@ -36,18 +36,38 @@ check_openstack_environment_is_ready_for_deploy (){
     openstack keypair list | grep ${KEYNAME} > /dev/null || (echo "keyname not found. export KEYNAME=" && exit 2)
 }
 
+check_python_requirement (){
+    if ! python -c "import ${1}" > /dev/null 2>&1; then
+        echo "Missing python requirement ${1}."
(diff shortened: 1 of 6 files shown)


script_library/pre-flight-checks.sh
```diff
@@ -36,18 +36,38 @@ check_openstack_environment_is_ready_for_deploy (){
...
+
 check_ansible_requirements (){
+    if [[ "${SOCOK8S_USE_VIRTUALENV:-False}" == "True" ]]; then
+        install_ansible
+    fi
     # Ansible is required
-    which ansible-playbook > /dev/null || install_ansible
+    if ! which ansible-playbook > /dev/null 2>&1; then
+        echo "Ansible is not installed"
+        echo "OR set SOCOK8S_USE_VIRTUALENV=True to install ansible and other requirements"
```
jsuchome: Does not this miss first part of the sentence? ("Ansible is not installed" does not look like something that should be followed by OR)

jhesketh: Fixed.

evrardjp:
Not tested but the code looks okay.

jsuchome:
Replaced by https://github.com/SUSE-Cloud/socok8s/pull/378

PR #378, opened 2019-05-23
Only install requirements from PyPI when explicitly set

Itxaka:
Requirements should come from packages unless in developer mode or the user has explicitly set SOCOK8S_USE_VIRTUALENV.

This is @jhesketh PR https://github.com/SUSE-Cloud/socok8s/pull/377/ + a small fix for the ara.rc location as `${socok8s_workspace}` is not defined at that point

### DIFF
--- a/script_library/run-ansible.sh
+++ b/script_library/run-ansible.sh
@@ -27,7 +27,7 @@ function run_ansible(){
 
     if [[ ${USE_ARA:-False} == "True" ]]; then
         echo "Loading ARA"
-        source ${socok8s_workspace}/.ansiblevenv/ara.rc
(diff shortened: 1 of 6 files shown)


Itxaka:
btw, tested and working

script_library/run-ansible.sh
```diff
@@ -27,7 +27,7 @@ function run_ansible(){
 
     if [[ ${USE_ARA:-False} == "True" ]]; then
         echo "Loading ARA"
-        source ${socok8s_workspace}/.ansiblevenv/ara.rc
+        source ${socok8s_workspace}/ara.rc
```
evrardjp: Is this consistent with the thing above?

PR #379, opened 2019-05-23
Debug deployment parameter

jesusaurus:


Jenkinsfile.integration
```diff
@@ -33,6 +37,7 @@ pipeline {
         stage('Show environment information') {
             steps {
                 sh 'printenv'
+                echo "Deployment parameter: ${params.deployment}"
```
toabctl: not really needed. The used parameters are visible eg here:  https://ci.suse.de/job/cloud-socok8s-jobs/job/socok8s-integration/job/PR-379/5/parameters/

Not sure if the parameter list is also visible in the "Blue ocean" jenkins UI.

jesusaurus: Thanks, I was struggling to find that link clicking through the various jenkins UIs (I don't know why I struggle so much with jenkins, but I find its UI worse than most)

PR #390, opened 2019-05-27
jenkins: retry SES step

Itxaka:
Seems that sometimes we are hitting an issue when deploying ses which
is fixed by retrying.

As the ses deployment is done by the ses team playbooks and we dont
have any control over those, we cannot directly try to indroduce the
retry on the playbook tasks. Instead we can just retry the jenkins
step, which should be idempotent, to check if its fixed.

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -50,7 +50,9 @@ pipeline {
                 }
                 stage('Setup SES') {
                     steps {
-                        sh "./run.sh deploy_ses"
+                        retry(3) {
+                            sh "./run.sh deploy_ses"
+                        }
                     }
                 }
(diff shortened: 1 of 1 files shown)


jsuchome:
I don't think this is the right approach. We already know which step usually fails ("Run stage 0 - prepare") so retrying that one would be much better then retrying whole `deploy_ses` step.

Even better would be to find out a correct fix instead of retrying....

jsuchome:
An alternative, I believe slightly better although not perfect https://github.com/SUSE-Cloud/automation/pull/3453

Itxaka:
https://github.com/SUSE-Cloud/automation/pull/3453 seems like a nicer approach!

PR #395, opened 2019-05-28
Specify a patchset revision for dev-patcher (SOC-9234)

jsuchome:
To make sure product deployment does not change if a patch is adapted
upstream, make sure specific revision is referenced instead of
always using this latest available one.


jsuchome:
I'm not sure ... AFAIK this is valid way of writing it although I admit it looks bit weird.

I'm just happy this seems to work as expected ...

jsuchome:
Rebased after the merge of https://github.com/SUSE-Cloud/socok8s/pull/389

PR #398, opened 2019-05-29
Do not require developer mode for applying patches

jsuchome:
It's possible patching is required even for non-developers. 
~Introduce new variable for that and set it to true by default.~

I think this might be necessary for the case when we need to release while patches are not merged upstream. It does not seem wise to tell users to use "developer mode", hence the ~new variable and~ new default.

Setting as do-not-merge initially as there might be opposition to this approach.

### DIFF
--- a/playbooks/generic-patch_upstream.yml
+++ b/playbooks/generic-patch_upstream.yml
@@ -2,17 +2,7 @@
 - hosts: soc-deployer
   gather_facts: yes
   any_errors_fatal: true
-  pre_tasks:
-    #http://jinja.pocoo.org/docs/2.10/templates/#default
-    - name: Check if developer mode should be enabled
-      set_fact:
-        developer_mode: "{{ (lookup('env','SOCOK8S_DEVELOPER_MODE') | default('False', true) ) | bool }}"
-      tags:
-        - always
-        - imagebuilder
-        - upstream_patching
   roles:
     - role: dev-patcher
-      when: developer_mode
       tags:
         - upstream_patching


playbooks/generic-patch_upstream.yml
```diff
@@ -3,16 +3,20 @@
   gather_facts: yes
   any_errors_fatal: true
   pre_tasks:
-    #http://jinja.pocoo.org/docs/2.10/templates/#default
-    - name: Check if developer mode should be enabled
+    - name: Check if upstream patches should be applied
```
jhesketh: Should still be "checking if developer mode is set"?

jsuchome: Good point, fixed

jsuchome:
Yes, I propose to apply patches by default. I agree it should be discussed.

jsuchome:
I think some documentation would need to be updated as well

jsuchome:
@evrardjp I'm not sure I understand what you propose.

To apply (available) patches always, and not let it control by any variable?

evrardjp:
> @evrardjp I'm not sure I understand what you propose.
> 
> To apply (available) patches always, and not let it control by any variable?

Correct. Always apply those patches.
Here is the reason:
- We have now split user level patches, and "socok8s sanctioned" ones, so it means all the socok8s have been through our vetting anyway.
- If any of those are empty, the tasks will not take place.

It sounds we are in a good place to just get rid of that conditional.

jsuchome:
I no one is against, I can of course change it that way. 

My idea was that still someone might want to influence the workflow so it goes with or without the patches (of course it's possible to edit the patch list, but that could be said about everything)

evrardjp:
Not really of everything, as we don't expose everything as user variable. or easy override. But this one is, on purpose.

Please note that if we really want to have a feature to influence the workflow to NOT have any patches applied, including the ones we find mandatory (which is kinda weird), then we could still remove that conditional and merge the two lists of patches to apply in dev-patcher :)

jsuchome:
@jhesketh @evrardjp still request to change something? CI is green

PR #399, opened 2019-05-29
Remove compute node is part of remove deployment.

rajdharwadkar:
Issue:
QA noticed nova instances going to ERROR in the new deployment.
"libvirt.libvirtError: operation failed: domain 'instance-00000001' already exists"
This was because nova instances still exists from previous deploy.

### DIFF
--- a/playbooks/roles/openstack-remove-compute/tasks/main.yml
+++ b/playbooks/roles/openstack-remove-compute/tasks/main.yml
@@ -8,11 +8,24 @@
   failed_when: false
   changed_when: false
 
-- name: Validate the compute node id.
+- name: Validate compute node id.
   debug:
     msg: "Compute node {{ compute_node_name }} does not exists in openstack."
   delegate_to: localhost
   when: compute_id.stdout | length | int == 0
+
+- name: Get VM list from compute node.
+  command: "openstack server list --host {{ compute_node_name }}"
+  register: vm_exist
+  delegate_to: localhost
+  failed_when: false
+  changed_when: false
+
+- name: Validate VM(s) exists on compute node.
+  fail:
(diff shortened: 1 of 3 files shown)


playbooks/roles/openstack-remove-compute/tasks/main.yml
```diff
@@ -14,6 +14,10 @@
   delegate_to: localhost
   when: compute_id.stdout | length | int == 0
 
+- debug:
```
jgu17: should this warning to be added to the run.sh and require user confirmation instead?

rajdharwadkar: Fixed

run.sh
```diff
@@ -140,7 +145,12 @@ case "$deployment_action" in
         clean_airship clean_openstack_clean_ucp_clean_rest
         ;;
     "remove_deployment")
-        clean_airship
+        read -r -p "WARNING: Please remove all VM(s) from compute host. Are you sure to continue? [y/n] " user_input
```
jgu17: -> Please remove all VM(s) from all compute host(s)

run.sh
```diff
@@ -91,7 +91,12 @@ case "$deployment_action" in
             echo "Please enter compute node host name"
             exit 1
         fi
-        remove_compute $2
+        read -r -p "WARNING: Please remove all VM(s) from compute host. Are you sure to continue? [y/n] " user_input
```
jgu17: ->Please remove all VM(s) from the compute host.

run.sh
rajdharwadkar: fixed

run.sh
rajdharwadkar: fixed

PR #400, opened 2019-05-29
doc: remove experimental use-case-2, provisioninfra ref

csymons-suse:
case consistency


jsuchome:
Do we really need to remove it? It's still usable for developers...

csymons-suse:
> Do we really need to remove it? It's still usable for developers...

It is still in the documentation in the suse-ecp section. It is only removed from the general sections.

PR #401, opened 2019-05-30
Set individual Jenkins timeouts

jhesketh:
Unfortunately we can't set a global timeout for all the pipeline as it
would also apply to the post stages and hence interrupt our cleanup.

Instead set a timeout on each stage.

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -40,7 +40,7 @@ pipeline {
 
         stage('Create network') {
             options {
-                timeout(time: 45, unit: 'MINUTES', activity: true)
+                timeout(time: 10, unit: 'MINUTES', activity: true)
             }
             steps {
                 sh "./run.sh deploy_network"
@@ -71,7 +71,7 @@ pipeline {
 
         stage('Configure CaaSP workers') {
             options {
-                timeout(time: 45, unit: 'MINUTES', activity: true)
+                timeout(time: 10, unit: 'MINUTES', activity: true)
             }
             steps {
                 sh "./run.sh enroll_caasp_workers"
@@ -81,7 +81,7 @@ pipeline {
 
         stage('Deploy OpenStack Helm') {
             options {
-                timeout(time: 45, unit: 'MINUTES', activity: true)
+                timeout(time: 20, unit: 'MINUTES', activity: true)
             }
             when {
                 expression { params.deployment == "osh" }
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -2,8 +2,10 @@ pipeline {
 
     options {
         timestamps()
-        timeout(time: 45, unit: 'MINUTES', activity: true)
         parallelsAlwaysFailFast()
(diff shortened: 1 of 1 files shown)


Jenkinsfile.integration
```diff
@@ -37,11 +39,17 @@ pipeline {
         }
 
         stage('Setup network') {
+            options {
+                timeout(time: 45, unit: 'MINUTES', activity: true)
```
jsuchome: This should be matter of few minutes tops

Jenkinsfile.integration
```diff
@@ -94,6 +111,9 @@ pipeline {
         }
 
         stage('Deploy Airship') {
+            options {
+                timeout(time: 45, unit: 'MINUTES', activity: true)
```
jsuchome: This however might require more

Jenkinsfile.integration
jhesketh: Agreed, I was just copying in what the default was. Note that it's also based on activity, so it's 45min of no stdout (afaict, as it's not well documented). So copying the same activity timeout to each section isn't giving it any more than what it previously had.

That said, we can find tune them in this PR if you think it's necessary?

toabctl:
@jhesketh can't we just overwrite the global timeout **only** in the post step (instead of setting it now in all steps)?

jsuchome:
... and this needs a rebase

toabctl:
> I like having timeouts per step, so we can finetune them to avoid waiting too much

These timeouts are `activity` timeouts. AFAIU, theses timeouts only trigger if there is no activity (eg. no output to stdout/stderr). So a step can easily take 50 hours if it prints every 30 min something to the console.

Itxaka:
> These timeouts are activity timeouts. AFAIU, theses timeouts only trigger if there is no activity (eg. no output to stdout/stderr). So a step can easily take 50 hours if it prints every 30 min something to the console.

Interesting, in that case, I do agree, only the minimal needed for the teardown should be added!

jhesketh:
> @jhesketh can't we just overwrite the global timeout **only** in the post step (instead of setting it now in all steps)?

Unfortunately once the timeout has occurred we only have a very short grace period in which to exit cleanly. After that Jenkins terminates the whole build.

It is not possible to give another timeout declaration during any of the post stages. Even if it was, I'm not sure if it would superced the previously set one, or be an additional counter (I believe it's the latter). You can wrap commands in a timeout wrapper (like we do for the input part in the `failure` section), but that also has no affect on the already triggered termination.

> > I like having timeouts per step, so we can finetune them to avoid waiting too much
> 
> These timeouts are `activity` timeouts. AFAIU, theses timeouts only trigger if there is no activity (eg. no output to stdout/stderr). So a step can easily take 50 hours if it prints every 30 min something to the console.

Even though that is true, we could still tune the timeouts with that knowledge, or on an individual stage basis choose which ones we want to be a "timer" and which ones we want to be based on inactivity.

Unfortunately because of the aforementioned lack of way of cancelling a triggered global timeout, I think we still need to have these on each section.

PR #402, opened 2019-05-30
Add socok8s package repo in Airship deployer

jgu17:
Add socok8s package repo on the Airship deployer node. Refactored the
repo list from openstack mechanism to common-vars so we can start
sharing the data between different deploy mechanisms.


jsuchome:
Why do we have this PR twice? https://github.com/SUSE-Cloud/socok8s/pull/408

PR #404, opened 2019-05-30
doc: remove Mitogen sections

csymons-suse:
does not work properly without considerable fussing

### DIFF
--- a/doc/source/deployment/setup-deployer.rst
+++ b/doc/source/deployment/setup-deployer.rst
@@ -120,7 +120,6 @@ Add the following.
 Add the above line after "#includedir /etc/sudoers.d". replace <username> with
 your username.
 
-
 Configure Ansible
 -----------------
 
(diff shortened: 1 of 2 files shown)


doc/source/deployment/setup-deployer.rst
```diff
@@ -176,7 +155,3 @@ You can improve SSH connections by enabling pipelining:
    [ssh_connection]
    pipelining = True
    EOF
-
-
-With this done, continue your deployment with the
-:ref:`configuredeployment` page.
```
Itxaka: is this part supposed to be removed?

doc/source/suse-ecp/ose-localhost.rst
```diff
@@ -125,27 +125,6 @@ configuration file loading ARA plugins:
 For more details on ARA's web interface, please read
 https://ara.readthedocs.io/en/stable/webserver.html .
 
-Enable Mitogen (optional)
```
jgu17: this is the part for ecp environment, target audient is devs. this option can be useful here? @Itxaka @evrardjp  your thoughts?

evrardjp: This is optional everywhere.
It's largely making the ansible deploys faster.

The documentation (before rewrite) was not ccp specific, and only ONE part (configure your hosts for socok8s) was partially on ccp (with an option to bring your own). So I think the refactor could have made that more unclear.

I will review this in the whole context of the docs later, if you like.

Itxaka: For me it was hanging on some random tasks so I had to disable it fully to complete a deploy.

If we recommend this in the docs we *need* to support it, so I would say leave it out as it would make troubleshooting more difficult.

evrardjp: Sounds good to me.

doc/source/deployment/setup-deployer.rst
csymons-suse: this is the only page with such an instruction. The deployment guide has SES Integration as the following step. And the Configure Cloud instructions indicate that the output from SES Integration must be present in order to Configure for SES Integration.

PR #410, opened 2019-05-31
Fix Nova missing genisoimage error

jgu17:
SUSE Nova image doesn't have genisoimage installed. Added overrides in
Nova chart to use mkisofs as replacement.

### DIFF
--- a/site/soc/software/charts/osh/openstack-compute-kit/nova.yaml
+++ b/site/soc/software/charts/osh/openstack-compute-kit/nova.yaml
@@ -110,6 +110,8 @@ data:
         libvirt:
           virt_type: qemu
           cpu_mode: host-model
+        DEFAULT:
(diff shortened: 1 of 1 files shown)


jgu17:
corresponding OSH Nova chrt change: https://review.opendev.org/#/c/662385.

jsuchome:
Can't you somehow include those upstream overrides in this yml intended for airship?

evrardjp:
> Can't you somehow include those upstream overrides in this yml intended for airship?

Do you mean directly use upstream override file, by applying the patch first?

jgu17:
> Can't you somehow include those upstream overrides in this yml intended for airship?

Possible doing some ansible maneuver, I wanted to see how Airship will handle the values_overrides. Deckhand doesn't seem to accommodate the pattern. Anyhow, I think that can be a different commit, post tech preview.

PR #412, opened 2019-05-31
remove deployment does not clean properly because of timeout issue.

rajdharwadkar:
Increased timeout for helm delete command because default is 300 secs
Force delete because pods are getting stuck.

Issues:
1. Pods are deleted but still helm charts exists.
2. Pods get stuck in terminating state for long time.


jgu17:
tested locally.

PR #414, opened 2019-05-31
Jenkinsfile: don't run the final deploys in "parallel"

jesusaurus:
Apparently skipping a parallel stage in jenkins as a good way to confuse
the Blue Ocean UI: https://issues.jenkins-ci.org/browse/JENKINS-48879

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -68,29 +68,25 @@ pipeline {
             }
         }
 
-        stage('Deploy') {
-            parallel {
-		stage('Deploy OpenStack Helm') {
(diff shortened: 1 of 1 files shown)


toabctl:
> Is there a use case where we might want to run these in parallel?

Imo currently no. 

> If so, instead of skipping over them with the `when` expression, we could just put in some bash logic on whether to execute the related ./run step. This will allow us to do the steps in parallel, even if one of the steps is effectively a no-op.

But then the UI shows that a step was successfully executed, even if it was skipped (via the bash logic).

PR #416, opened 2019-06-03
Remove the requirement for Leap15develtools repository

jsuchome:
Seems unnecessary, let's find out if it is true


jsuchome:
Currently we require the repo for `phantomjs` package. Adding the package into our project repo here: https://build.opensuse.org/request/show/707108

vars/deploy-on-openstack.yml
```diff
@@ -23,7 +23,6 @@ deploy_on_openstack_ses_repos_per_imagename:
     - SES5-update
 deploy_on_openstack_osh_repos_per_imagename:
   openSUSE-Leap-15.0:
-    - Leap15develtools
```
jhesketh: I don't think this is used anywhere else so we should remove it from common vars

PR #417, opened 2019-06-03
Updated glance bootstrap values to use cirros image 0.4.0 with public visibility.

matthew-fuller:
This change updates the Cirros test image from 0.3.5 to 0.4.0, which brings it up to date with what's used in Cloud 9. It also sets the visibility to "public", which is required for some of the tempest tests to pass.

### DIFF
--- a/site/soc/software/charts/osh/openstack-glance/glance.yaml
+++ b/site/soc/software/charts/osh/openstack-glance/glance.yaml
@@ -37,6 +37,14 @@ data:
     enabled: {{ run_tests }}
     timeout: {{ test_timeout }}
   values:
+    bootstrap:
+      structured:
+        images:
+          cirros:
+            name: "Cirros 0.4.0 64-bit"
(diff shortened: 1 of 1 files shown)


jsuchome:
Does the rally config need an update too? E.g. https://github.com/openstack/openstack-helm/blob/master/glance/values.yaml#L134

jsuchome:
Also, could you point out which tempest tests need public visibility?

matthew-fuller:
> Does the rally config need an update too? E.g. https://github.com/openstack/openstack-helm/blob/master/glance/values.yaml#L134

Ah, good catch. I'll update it here as well.

matthew-fuller:
> Also, could you point out which tempest tests need public visibility?

I don't know the specific tests that QE encountered that were failing, but the issue became apparent after removing the key 'tempest_roles: admin, member' from tempest.conf, since that key defeats the purpose of tests that test the policy definition.

PR #421, opened 2019-06-03
Make sure helm is initialized for root user

jsuchome:


jsuchome:
Well, looks like this might be in conflict with other parts like https://github.com/SUSE-Cloud/socok8s/blob/master/playbooks/roles/common-deployer/tasks/kubectl-setup.yml#L12 where home directory of ansible user (not root) is used....

PR #422, opened 2019-06-03
Install helm-serve unit file directly if ansible is not running as root

jsuchome:
And set User=ansible user so that the service can run as current user

By default the service file is provided with the `helm` package, but that one assumes helm-serve is run by root.

### DIFF
--- a/playbooks/roles/common-deployer/templates/helm-serve.service.j2
+++ b/playbooks/roles/common-deployer/templates/helm-serve.service.j2
@@ -1,0 +1,11 @@
+[Unit]
+Description=Helm Server
+After=network.target
+
+[Service]
+User={{ ansible_user }}
+Restart=always
+ExecStart=/usr/bin/helm serve
+
+[Install]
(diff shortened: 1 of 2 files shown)


toabctl:
And we should clarify what user should call socok8s

jgu17:
ugh.. how long does the CI take to report a yes or no?

playbooks/roles/common-deployer/templates/helm-serve.service.j2
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Helm Server
+After=network.target
+
+[Service]
+User={{ ansible_user }}
+Restart=always
+ExecStart=/usr/local/bin/helm serve
```
sgrasley: helm is installed /usr/bin/

jsuchome: Oh! Good point

jsuchome:
@sgrasley updated

PR #423, opened 2019-06-03
Fix the novncproxy url in host rules

jgu17:
The novncprox host rule in the /etc/hosts was missing "nova-" prefix.

### DIFF
--- a/playbooks/roles/airship-setup-deployer/tasks/main.yml
+++ b/playbooks/roles/airship-setup-deployer/tasks/main.yml
@@ -100,7 +100,7 @@
       {{ socok8s_ext_vip }} compute.openstack.svc.cluster.local
       {{ socok8s_ext_vip }} network.openstack.svc.cluster.local
       {{ socok8s_ext_vip }} dashboard.openstack.svc.cluster.local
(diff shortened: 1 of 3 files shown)


playbooks/roles/airship-setup-deployer/tasks/main.yml
```diff
@@ -100,7 +100,7 @@
       {{ socok8s_ext_vip }} compute.openstack.svc.cluster.local
       {{ socok8s_ext_vip }} network.openstack.svc.cluster.local
       {{ socok8s_ext_vip }} dashboard.openstack.svc.cluster.local
-      {{ socok8s_ext_vip }} novncproxy.openstack.svc.cluster.local
+      {{ socok8s_ext_vip }} nova-novncproxy.openstack.svc.cluster.local
```
kantarun: Can we make similar change in OSH side ?

https://github.com/SUSE-Cloud/socok8s/blob/master/playbooks/roles/deploy-osh/tasks/main.yml#L92

jgu17: done

jgu17:
merged thru command ine

PR #424, opened 2019-06-04
SOC-9435: Run commands from the correct host

jesusaurus:
Run 'openstack' and 'kubectl' commands from the deployer, not from
localhost. In ECP environments localhost does not have kubectl
installed, and openstack commands go to the ECP endpoints by default.

Delegate these commands to the first soc-deployer host, which will give
the expected output in any environment.

### DIFF
--- a/playbooks/roles/openstack-remove-compute/tasks/main.yml
+++ b/playbooks/roles/openstack-remove-compute/tasks/main.yml
@@ -4,27 +4,25 @@
      set -o pipefail
      openstack compute service list | grep {{ compute_node_name }} | awk '{print $2}'
   register: compute_id
-  delegate_to: localhost
+  delegate_to: "{{ groups['soc-deployer'][0] }}"
   failed_when: false
   changed_when: false
 
 - name: Validate compute node id.
   debug:
     msg: "Compute node {{ compute_node_name }} does not exists in openstack."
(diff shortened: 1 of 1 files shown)


playbooks/roles/openstack-remove-compute/tasks/main.yml
```diff
@@ -4,46 +4,44 @@
...
   register: vm_exist
-  delegate_to: localhost
+  delegate_to: "{{ groups['soc-deployer'][0] }}"
   failed_when: false
   changed_when: false
 
 - name: Validate VM(s) exists on compute node.
   fail:
     msg: "VM(s) exists on compute node {{ compute_node_name }}."
-  delegate_to: localhost
```
bryanstephenson: Everywhere else we change the delegate_to, but here we just delete it. Why is this one different?

jesusaurus: There are two places where I removed it instead of changing the target: the debug module above and here at the fail module. Both of these module are kinda special in that they don't really execute anything on the target host, so to me it doesn't make sense to specify a delegate host for those modules.

bryanstephenson: Thank you for the explanation.

PR #426, opened 2019-06-04
Make sure we're using python 3

jhesketh:
We only support python3, and our RPM package installs the python3
dependencies. However, our scripts were calling 'python' which may still
be python2 on the localhost, so instead explicitly call python3
everywhere.


jsuchome:
CI is already past this initial steps, merging

PR #428, opened 2019-06-04
Update docs

sgrasley:
Changed deploy_airship to just deploy. Added instructions to use
virtualenv. Added jmespath to required packages.

### DIFF
--- a/doc/source/deployment/setup-deployer.rst
+++ b/doc/source/deployment/setup-deployer.rst
@@ -46,7 +46,8 @@ The following software must be installed on your `Deployer`:
   * gcc
   * git
   * jq
-  * python3-netaddr
+  * python-jmespath
+  * python-netaddr
   * python-virtualenv
 
 Create SUSE Containerized OpenStack Workspace
@@ -67,6 +68,11 @@ Deployer. Set up your workspace with the following steps:
   export SOCOK8S_ENVNAME=socok8s
   export SOCOK8S_WORKSPACE_BASEDIR=~/socok8s-workspace
 
+Prepare deployer to use virtualenv.
+
+..code-block:: console
+
+  export SOCOK8S_USE_VIRTUALENV=True
 
 Cloning the repository
 ----------------------
--- a/doc/source/deployment/deploy.rst
+++ b/doc/source/deployment/deploy.rst
@@ -34,7 +34,7 @@ To deploy SUSE OpenStack cloud using Airship, run:
 
 .. code-block:: console
 
-   ./run.sh deploy_airship
+   ./run.sh deploy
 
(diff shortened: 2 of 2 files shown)


doc/source/deployment/setup-deployer.rst
```diff
@@ -46,7 +46,8 @@ The following software must be installed on your `Deployer`:
   * gcc
   * git
   * jq
-  * python3-netaddr
+  * python-jmespath
+  * python-netaddr
```
jsuchome: Why this? Package has explicit python3 requirements, see https://build.opensuse.org/package/view_file/Cloud:socok8s/socok8s/socok8s.spec?expand=1

And we just merged another PR that runs explicit `python3` when importing modules: https://github.com/SUSE-Cloud/socok8s/pull/426

doc/source/deployment/setup-deployer.rst
```diff
@@ -67,6 +68,11 @@ Deployer. Set up your workspace with the following steps:
   export SOCOK8S_ENVNAME=socok8s
   export SOCOK8S_WORKSPACE_BASEDIR=~/socok8s-workspace
 
+Prepare deployer to use virtualenv.
+
+..code-block:: console
+
+  export SOCOK8S_USE_VIRTUALENV=True
```
jsuchome: I do not have the whole picture, but this should not be needed at the end. If user has installed from the package, all should work out of the box (and https://github.com/SUSE-Cloud/socok8s/pull/426 should help again)

doc/source/deployment/setup-deployer.rst
sgrasley: Thanks, let me fix that.

evrardjp: I still like to install those myself, so I think it's still worth mentioning. I don't know why it was removed or if I forgot it in the past.

doc/source/deployment/setup-deployer.rst
evrardjp: Agreed with @jsuchome .

doc/source/deployment/setup-deployer.rst
Itxaka: This should be mentioned but probably in a different part? like development? 

Because we are using package for installation, its the package the one that should have those requirements included and pull those, you should not need to install the package + extras for basic functionality!

matthew-fuller:
I'm not sure if it's appropriate to mention here or if I should open a separate PR, but line 88 of deploy.rst should be changed to reflect the new path for the airship shipyard repo:
  /opt/airship/shipyard/tools/shipyard.sh ...

PR #433, opened 2019-06-04
Changing use virtual env default to True if not set

kantarun:
Currently it seems that virtual env setup is *always* needed otherwise
we see errors during deployment at various stages.
So changing default as does not ask user to set it to True if its
always needed.

### DIFF
--- a/doc/source/reference.rst
+++ b/doc/source/reference.rst
@@ -321,9 +321,8 @@ continues the deployment.
 
 ``SOCOK8S_USE_VIRTUALENV`` determines if the script should set up and use a
 virtualenv for python and ansible requirements. Without this it is expected
-that ansible and the requirements are installed via system packages.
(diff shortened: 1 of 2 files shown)


jsuchome:
No, virtual env setup should not be needed. Please report the errors you have without it.

jhesketh:
Right, I agree with jsuchome. The bug here is that it doesn't work without venv.

If you're installing from the package then hopefully it should just work. If it's not, it's a bug.

If you're installing/running from source then you either need to set up the dependencies and your environment yourself, set the developer flag, or set the create venv flag as a convenience.

sgrasley:
I think that the commit message could be better worded.  I think the intent is to reduce the number of pre-requisite packages that need to be sourced and installed before running installation. This would accomplish that.

Itxaka:
Lets close this for now, it can be restored later, doesnt seem to be needed for now.

PR #434, opened 2019-06-04
openstack client: pass through OS_AUTH_TYPE

JanZerebecki:
Because of https://storyboard.openstack.org/#!/story/2005821 to use
token based auth one needs to pass through OS_AUTH_TYPE to any openstack
client call.

Token based authentication allows a deployer that doesn't know the
openstack password.

### DIFF
--- a/playbooks/openstack-caasp_instance.yml
+++ b/playbooks/openstack-caasp_instance.yml
@@ -21,6 +21,7 @@
             state: "present"
             name: "{{ stackname }}"
             template: "{{ playbook_dir }}/../files/caasp-stack.yaml"
+            auth_type: "{{ os_auth_type }}"
             parameters:
               image: "{{ deploy_on_openstack_caasp_image }}"
               external_net: "{{ deploy_on_openstack_external_network }}"
--- a/playbooks/openstack-create_network.yml
+++ b/playbooks/openstack-create_network.yml
@@ -16,6 +16,7 @@
             state: "present"
             name: "{{ deploy_on_openstack_network_stackname }}"
             template: "{{ playbook_dir }}/../heat-templates/openstack-network.yml"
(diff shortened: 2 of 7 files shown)


jsuchome:
Hm. Is the CI failure related?

JanZerebecki:
The error looks as if one tried `OS_AUTH_TYPE=token_endpoint` with a config for type `password`. What does the cloud.yaml look like in the CI?

JanZerebecki:
Ci has that environment variable unset and the cloud.yaml looks like a normal one used with type password. Manually using the openstack cli with ´OS_AUTH_TYPE=password` works on the CI host. Using this patch with ´OS_AUTH_TYPE=token` works for me.

JanZerebecki:
Should work now.

Reading environment variables in Ansible always succeeds and gives an empty string if the variable doesn't exist, so default doesn't trigger unless it is told to also do so on the empty string. Unlike the openstack cli the ansible modules seem to not understand the empty string as the use defaults.

JanZerebecki:
That doesn't work, see the link.

jhesketh:
Ah, I misunderstood it at first glance. Thanks.

JanZerebecki:
Please approve again. Rebased, because I don't know any other way to make the CI retest the merge result (rerunning the job didn't help).

JanZerebecki:
CI previously passed. New failure because of missing job.

evrardjp:
I think there was a misconception of where the problem lies in here.

IMO, shade/openstacksdk is exposing the OS_.* variables to the ansible module by default.

So here, we could have two issues: The ansible modules themselves not using OS_.* variable for auth_type properly, or the sdk/shade library not exposing those the right way. In any case, I don't think it's an openstackclient issue, and this patch should not have existed in the first place.

PR #436, opened 2019-06-04
doc: copyedits, consistency

csymons-suse:
remove unnecessary SCO acronym; limit line length of CLI commands; add clarity to structure; spell out acronym reference.

### DIFF
--- a/doc/source/operations.rst
+++ b/doc/source/operations.rst
@@ -54,8 +54,8 @@ run the following command in the root of the socok8s directory:
 
    ./run.sh remove_deployment
 
-This will delete all Helm releases, all Kubernetes resources in the ucp and
-openstack namespaces, and all persistent volumes that were provisioned for use
+This will delete all Helm releases, all Kubernetes resources in the UCP and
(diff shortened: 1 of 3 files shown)


doc/source/operations.rst
```diff
@@ -54,8 +54,8 @@ run the following command in the root of the socok8s directory:
 
    ./run.sh remove_deployment
 
-This will delete all Helm releases, all Kubernetes resources in the ucp and
-openstack namespaces, and all persistent volumes that were provisioned for use
+This will delete all Helm releases, all Kubernetes resources in the UCP and
```
matthew-fuller: The lack of capitalization was a deliberate choice here since the namespaces are actually 'ucp' and 'openstack'. Would it be better to leave them lower case and italicize or bold them to indicate that those terms are meant to be that way?  It's also possible that I'm being too fussy...

csymons-suse: fixed. lower case, literal. Not too fussy. You're correct.

PR #440, opened 2019-06-05
modified delete subvolume function to delete only content of subvolume

rajdharwadkar:
Earlier approach was clean everything that was configured during deployment because
of cleanup issues , we are keeping fstab entry, subvolumes, mounted dir.

Issue: during remove compute, delete subvolume fails to remove directory in some environment.
this was seen often in qe environment.

### DIFF
--- a/playbooks/roles/airship-configure-worker/templates/caasp_worker_node_set_subvolumes.sh.j2
+++ b/playbooks/roles/airship-configure-worker/templates/caasp_worker_node_set_subvolumes.sh.j2
@@ -40,12 +40,7 @@ restore_and_reboot() {
 
 delete_subvolume(){
   if [ -d "$1" ]; then
-    init
-    if devpath=$(grep $1 /proc/mounts | awk '{print $1}') ; then
-      mount $devpath $2 -o subvol="/@"
-      btrfs subvolume delete "$2/$1" || true
-      umount "$2" || true
-    fi
+    rm -rf "$1/"*
   fi
 }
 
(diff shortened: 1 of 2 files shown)


playbooks/roles/airship-configure-worker/templates/caasp_worker_node_set_subvolumes.sh.j2
```diff
@@ -40,12 +40,7 @@ restore_and_reboot() {
 
 delete_subvolume(){
   if [ -d "$1" ]; then
-    init
-    if devpath=$(grep $1 /proc/mounts | awk '{print $1}') ; then
-      mount $devpath $2 -o subvol="/@"
-      btrfs subvolume delete "$2/$1" || true
-      umount "$2" || true
-    fi
+    rm -rf "$1/"*
```
Itxaka: yikes, I have a feeling that this is prone to errors and somehow could do harm, isnt there an alternative approach? Whats is wrong with the subvolume delete? shouldnt we try to fix that instead?

rajdharwadkar: Issue is mentioned in patch.  Yes I hate this fix too :) . but we had no alternative at this point because node is going to maintenance mode .

jgu17: both create subvolume and delete subvolume are error prone from our experience on the caaps nodes, and require reboot which is time consuming. Deleting content seems to be safter and more efficient.

Itxaka: :( :(

PR #441, opened 2019-06-05
remove DELETE_ANYWAY=YES requirement for remove deployment.

rajdharwadkar:
We already have warning and user input to continue in run.sh

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -146,7 +146,7 @@ case "$deployment_action" in
         clean_airship clean_openstack_clean_ucp_clean_rest
         ;;
     "remove_deployment")
-        read -r -p "WARNING: Please remove all VM(s) from all compute host(s). Are you sure to continue? [y/n] " user_input
+        read -r -p "WARNING: Please remove all VM(s) from all compute host(s). This deletes everything that is deployed. Are you sure to continue? [y/n] " user_input
         if [[ $user_input == "y" ]]; then
             clean_airship
         else
--- a/script_library/deployment-actions-kvm.sh
+++ b/script_library/deployment-actions-kvm.sh
@@ -80,10 +80,8 @@ function clean_airship(){
         clean_action=" -e clean_action=$1"
         action_desc=$1
     fi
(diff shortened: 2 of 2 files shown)


script_library/deployment-actions-kvm.sh
```diff
@@ -80,10 +80,8 @@ function clean_airship(){
         clean_action=" -e clean_action=$1"
         action_desc=$1
     fi
-    echo "DANGER ZONE. Set the env var 'DELETE_ANYWAY' to 'YES' to delete airship artifacts ( ${action_desc} ) in your userspace."
-    if [[ ${DELETE_ANYWAY:-"NO"} == "YES" ]]; then
-        run_ansible ${socok8s_absolute_dir}/playbooks/generic-clean_airship.yml ${clean_action}
-    fi
+    echo "DANGER ZONE. This delete airship artifacts ( ${action_desc} ) in your userspace."
```
Itxaka: `This will delete all ariship artifacts...`

Itxaka: should we also remove  the `DANGER ZONE` ?

run.sh
```diff
@@ -146,7 +146,7 @@ case "$deployment_action" in
         clean_airship clean_openstack_clean_ucp_clean_rest
         ;;
     "remove_deployment")
-        read -r -p "WARNING: Please remove all VM(s) from all compute host(s). Are you sure to continue? [y/n] " user_input
+        read -r -p "WARNING: Please remove all VM(s) from all compute host(s). This Deletes everything that is deployed. Are you sure to continue? [y/n] " user_input
```
jgu17: Deletes->deletes

script_library/deployment-actions-kvm.sh
jgu17: Do we still want to keep action_desc for dev options?

rajdharwadkar: > should we also remove the `DANGER ZONE` ?

Keeping it because we need that for internal dev environment. because that clean airship function is also called from other options.  Currently  user input / message only exist in remove_deployment. option.

rajdharwadkar: > Do we still want to keep action_desc for dev options?

Yeah we need action_dev for internal dev env .

PR #442, opened 2019-06-05
Jenkinsfile: Abort previous builds

toabctl:
If a new build starts, abort already running builds to save
resources.

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -38,6 +38,14 @@ pipeline {
             }
         }
 
+        stage('Abort previous builds') {
+            steps {
+                script {
+                    abortPreviousRunningBuilds()
+                }
+            }
+        }
+
         stage('Create network') {
             options {
                 timeout(time: 10, unit: 'MINUTES', activity: true)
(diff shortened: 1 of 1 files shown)


toabctl:
> Won't this cancel any build of this job? We probably only want to cancel builds of the same PR/branch. 

I'm not sure what it is doing exactly. This PR is here to find out :)
> Also I'm not sure if Jenkins already does this, or if it waits for a previous build to finish if one is already running against the branch (therefore we won't get a chance to cancel it).

I think it does not. That's why we have this long waiting queue in jenkins.

But currently it does anyway not work because we don't have the permissions on jenkins to run the groovy script (I talked to Daniel to change this. let's see if we get the perms). See https://jenkins.io/doc/book/managing/script-approval/

Jenkinsfile.integration
```diff
@@ -129,3 +137,25 @@ pipeline {
         }
     }
 }
+
+
+def abortPreviousRunningBuilds() {
+    def hi = Hudson.instance
+    def pname = env.JOB_NAME.split('/')[0]
```
Itxaka: will this fail with master? 

Also, should we skip master?

Itxaka:
Lets close this for now, it can be restored later if needed.

PR #443, opened 2019-06-05
Fix Neutron tenant networking

jgu17:
Allows the user to define neutron tunnel device and external interface for private and floating ip. Also added python3 l3 and dhcp rootwrap filters.

### DIFF
--- a/playbooks/roles/airship-deploy-ucp/defaults/main.yml
+++ b/playbooks/roles/airship-deploy-ucp/defaults/main.yml
@@ -2,3 +2,5 @@
 airship_pegleg_image: "{{ suse_airship_registry_location }}/airshipit/pegleg:{{ suse_airship_components_image_tag['pegleg'] }}"
 run_tests: true
 test_timeout: 2700
+neutron_external_interface: ""
+neutron_tunnel_device: null
--- a/site/soc/software/charts/osh/openstack-compute-kit/neutron.yaml
+++ b/site/soc/software/charts/osh/openstack-compute-kit/neutron.yaml
@@ -73,7 +73,7 @@ data:
             # for the migration process
             kill_metadata: KillFilter, root, python, -15, -9
             kill_metadata7: KillFilter, root, python2.7, -15, -9
-            kill_metadata35: KillFilter, root, python3.5, -15, -9
+            # Added socok8s downstream to support python 3.6 in openSUSE Neutron image
             kill_metadata36: KillFilter, root, python3.6, -15, -9
             kill_radvd_usr: KillFilter, root, /usr/sbin/radvd, -15, -9, -HUP
             kill_radvd: KillFilter, root, /sbin/radvd, -15, -9, -HUP
@@ -159,7 +159,7 @@ data:
             # for the migration process
             kill_metadata: KillFilter, root, python, -9
             kill_metadata7: KillFilter, root, python2.7, -9
-            kill_metadata35: KillFilter, root, python3.5, -9
+            # Added in socok8s downstream to support Pythong 3.6 in openSUSE Neuton image
             kill_metadata36: KillFilter, root, python3.6, -9
             
             # ip_lib
--- a/site/soc/software/charts/osh/openstack-compute-kit/neutron.yaml
+++ b/site/soc/software/charts/osh/openstack-compute-kit/neutron.yaml
@@ -43,4 +43,127 @@ data:
       paste:
         app:neutronversions:
           paste.app_factory: neutron.pecan_wsgi.app:versions_factory
+      rootwrap_filters:
+        l3:
+          content: |
+            # neutron-rootwrap command filters for nodes on which neutron is
+            # expected to control network
+            #
+            # This file should be owned by (and only-writeable by) the root user
+           
+            # format seems to be
+            # cmd-name: filter-name, raw-command, user, args
+           
+            [Filters]
+            
+            # arping
+            arping: CommandFilter, arping, root
+           
+            # l3_agent
+            sysctl: CommandFilter, sysctl, root
+            route: CommandFilter, route, root
+            radvd: CommandFilter, radvd, root
+            
+            # haproxy
+            haproxy: RegExpFilter, haproxy, root, haproxy, -f, .*
+            kill_haproxy: KillFilter, root, haproxy, -15, -9, -HUP
+            # RHEL invocation of the metadata proxy will report /usr/bin/python
+            # TODO(dalvarez): Remove kill_metadata* filters in Q release since
+            # neutron-ns-metadata-proxy is now replaced by haproxy. We keep them for now
+            # for the migration process
+            kill_metadata: KillFilter, root, python, -15, -9
+            kill_metadata7: KillFilter, root, python2.7, -15, -9
+            kill_metadata35: KillFilter, root, python3.5, -15, -9
+            kill_metadata36: KillFilter, root, python3.6, -15, -9
+            kill_radvd_usr: KillFilter, root, /usr/sbin/radvd, -15, -9, -HUP
+            kill_radvd: KillFilter, root, /sbin/radvd, -15, -9, -HUP
+            
+            # ip_lib
+            ip: IpFilter, ip, root
+            find: RegExpFilter, find, root, find, /sys/class/net, -maxdepth, 1, -type, l, -printf, %.*
+            ip_exec: IpNetnsExecFilter, ip, root
+           
+            # l3_tc_lib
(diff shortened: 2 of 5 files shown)


playbooks/roles/airship-deploy-ucp/defaults/main.yml
```diff
@@ -2,3 +2,5 @@
 airship_pegleg_image: "{{ suse_airship_registry_location }}/airshipit/pegleg:{{ suse_airship_components_image_tag['pegleg'] }}"
 run_tests: true
 test_timeout: 2700
+neutron_external_interface: public
+neutron_tunnel_device: docker0
```
jsuchome: Does it make sense to have `docker0` as a default value? I was under the impression that no value is better ...

rhafer: Yeah, I think it's better to leave it empty (or set to `null`) than to put `docker0` here. `docker0` would only work in single node deployments. But in that case neutron doesn't even need to create any tunnels.

script_library/requirements.txt
```diff
@@ -1,6 +1,6 @@
 ansible==2.7.8
 netaddr==0.7.19
-jmespath==0.9.3
+jmespath>=0.9.3
```
Itxaka: ????
Any reason why this, why is not mentioned and why are we unfreezing a known version?

stephenbma: This change is not explained in the commit text.

site/soc/software/charts/osh/openstack-compute-kit/neutron.yaml
```diff
@@ -40,33 +34,12 @@ data:
...
-          ml2_type_flat:
-            flat_networks: public
-        openvswitch_agent:
-          agent:
-            tunnel_types: vxlan
-          ovs:
-            bridge_mappings: public:br-ex
-        linuxbridge_agent:
-          linux_bridge:
-            bridge_mappings: public:br-ex..
```
stephenbma: Why are L52-L69 removed?

playbooks/roles/airship-deploy-ucp/defaults/main.yml
```diff
@@ -2,3 +2,5 @@
 airship_pegleg_image: "{{ suse_airship_registry_location }}/airshipit/pegleg:{{ suse_airship_components_image_tag['pegleg'] }}"
 run_tests: true
 test_timeout: 2700
+neutron_external_interface: public
```
rhafer: This interface does not exist by default. So this will just break when deploying the ovs-agent.

site/soc/software/charts/osh/openstack-compute-kit/neutron.yaml
jgu17: > Why are L52-L69 removed?

because they are provided in the global definition. The removed lines are more settings for dev environment.

playbooks/roles/airship-deploy-ucp/defaults/main.yml
jgu17: done

script_library/requirements.txt
jgu17: no reason. just by accident :-)

jgu17: fixed

jgu17: fixed

playbooks/roles/airship-deploy-ucp/defaults/main.yml
jgu17: changed default to empty.

jgu17:
includes the dhcp and l3 filter changes.

jgu17:
Tested successful deployment after the inclusion of l3 and dhcp filters.

site/soc/software/charts/osh/openstack-compute-kit/neutron.yaml
```diff
@@ -40,34 +34,134 @@ data:
...
+            # haproxy
+            haproxy: RegExpFilter, haproxy, root, haproxy, -f, .*
+            kill_haproxy: KillFilter, root, haproxy, -15, -9, -HUP
+            # RHEL invocation of the metadata proxy will report /usr/bin/python
+            # TODO(dalvarez): Remove kill_metadata* filters in Q release since
+            # neutron-ns-metadata-proxy is now replaced by haproxy. We keep them for now
+            # for the migration process
+            kill_metadata: KillFilter, root, python, -15, -9
+            kill_metadata7: KillFilter, root, python2.7, -15, -9
+            kill_metadata35: KillFilter, root, python3.5, -15, -9
```
jsuchome: Shouldn't it be `python3.6` ?

site/soc/software/charts/osh/openstack-compute-kit/neutron.yaml
```diff
@@ -40,34 +34,134 @@ data:
...
+            # haproxy
+            haproxy: RegExpFilter, haproxy, root, haproxy, -f, .*
+            kill_haproxy: KillFilter, root, haproxy, -15, -9, -HUP
+            # RHEL invocation of the metadata proxy will report /usr/bin/python
+            # TODO(dalvarez): Remove kill_metadata* filters in Q release since
+            # neutron-ns-metadata-proxy is now replaced by haproxy. We keep them for now
+            # for the migration process
+            kill_metadata: KillFilter, root, python, -9
+            kill_metadata7: KillFilter, root, python2.7, -9
+            kill_metadata35: KillFilter, root, python3.5, -9
```
jsuchome: Same here, do we need python3 or python3.6 ?

site/soc/software/charts/osh/openstack-compute-kit/neutron.yaml
jgu17: added 3.5 line

site/soc/software/charts/osh/openstack-compute-kit/neutron.yaml
jgu17: yes, added 3.6 line

site/soc/software/charts/osh/openstack-compute-kit/neutron.yaml
```diff
@@ -40,34 +34,136 @@ data:
...
+            conntrack: CommandFilter, conntrack, root
+            
+            # keepalived state change monitor
+            keepalived_state_change: CommandFilter, neutron-keepalived-state-change, root
+            # The following filters are used to kill the keepalived state change monitor.
+            # Since the monitor runs as a Python script, the system reports that the
+            # command of the process to be killed is python.
+            # TODO(mlavalle) These kill filters will be updated once we come up with a
+            # mechanism to kill using the name of the script being executed by Python
+            kill_keepalived_monitor_py: KillFilter, root, python, -15
```
Itxaka: Is this what we are adding? Can we indicate that so its easier to track and upstream it?

site/soc/software/charts/osh/openstack-compute-kit/neutron.yaml
```diff
@@ -40,34 +34,136 @@ data:
...
+            # haproxy
+            haproxy: RegExpFilter, haproxy, root, haproxy, -f, .*
+            kill_haproxy: KillFilter, root, haproxy, -15, -9, -HUP
+            # RHEL invocation of the metadata proxy will report /usr/bin/python
+            # TODO(dalvarez): Remove kill_metadata* filters in Q release since
+            # neutron-ns-metadata-proxy is now replaced by haproxy. We keep them for now
+            # for the migration process
+            kill_metadata: KillFilter, root, python, -9
+            kill_metadata7: KillFilter, root, python2.7, -9
+            kill_metadata35: KillFilter, root, python3.5, -9
```
Itxaka: can we indicate here that this lines were added by us downstream so we can upstream them and track the changes better?

Itxaka: also, our neutron images have 3.6 so do we still need to add the 3.5?

site/soc/software/charts/osh/openstack-compute-kit/neutron.yaml
jsuchome: Yep, AFAIK only those lines with python3* are new. It's unfortunate we can't really add just a snippet, but the whole config file is a value :-(

And yes, we should add it upstream for sure

site/soc/software/charts/osh/openstack-compute-kit/neutron.yaml
jsuchome: It's true that 3.5 lines (not just here but in whole file) seem unnecessary

PR #444, opened 2019-06-05
ci: add linting for playbooks in the pipeline

Itxaka:
Will trigger an ansible-lint run if we detect changes under the
playbooks dir

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -38,6 +38,15 @@ pipeline {
             }
         }
 
+        stage('Lint Ansible playbooks') {
+            options {
+                timeout(time: 5, unit: 'MINUTES', activity: true)
+            }
+            steps {
+                sh "tox -e lint"
+            }
+        }
+
         stage('Create network') {
             options {
                 timeout(time: 10, unit: 'MINUTES', activity: true)


Itxaka:
the `when` should execute correctly as we can see the changes properly in jenkins, it should detect that something has changed there and run the step. We should probably do the same for the docs :)

evrardjp:
> the `when` should execute correctly as we can see the changes properly in jenkins, it should detect that something has changed there and run the step. We should probably do the same for the docs :)

For docs we should be only running the docs jobs if it's a docs only change to also save resources.

Jenkinsfile.integration
```diff
@@ -38,6 +38,18 @@ pipeline {
             }
         }
 
+        stage('Lint') {
```
evrardjp: nit: I would say lint playbooks to make it clear it's only linting plays.

Jenkinsfile.integration
```diff
@@ -38,6 +38,18 @@ pipeline {
...
+            options {
+                timeout(time: 5, unit: 'MINUTES', activity: true)
+            }
+            when {
+                changeset "playbooks/**"
```
toabctl: maybe also when tox.ini changes?

Itxaka:
> For docs we should be only running the docs jobs if it's a docs only change to also save resources.

Yep, I was trying to imply that, sorry for not making it clearer!

Jenkinsfile.integration
Itxaka: done

Jenkinsfile.integration
Itxaka: done

toabctl:
@Itxaka you sould change something in tox.ini or playbooks/ to actually trigger the build step :)

Itxaka:
> @Itxaka you sould change something in tox.ini or playbooks/ to actually trigger the build step :)

Already done :P
https://github.com/SUSE-Cloud/socok8s/pull/445 and https://github.com/SUSE-Cloud/socok8s/pull/446

Itxaka:
did not work :(

Itxaka:
`14:00:50  Warning, empty changelog. Probably because this is the first build.`

¬_¬ so on first build there is no way of checking what has changed? what?

Itxaka:
opened since 2009.... https://issues.jenkins-ci.org/browse/JENKINS-4162

Itxaka:
we need tox in those nodes, not sure where to do that or request that :/

toabctl:
> we need tox in those nodes, not sure where to do that or request that :/

you can ask me :) But it seems to be already installed ....

Itxaka:
> you can ask me :) But it seems to be already installed ....

woops, yeah I installed it yesterday with @jhesketh help :)

And the job went green as well (at least the linting part)

PR #447, opened 2019-06-05
Move common deployment helpers into common file

jsuchome:


script_library/deployment-actions-kvm.sh
```diff
@@ -63,39 +36,6 @@ function remove_compute(){
...
     echo "Not implemented"
 }
-function clean_userfiles(){
-    echo "DANGER ZONE. Set the env var 'DELETE_ANYWAY' to delete everything in your userspace."
-    if [[ ${DELETE_ANYWAY:-"NO"} == "YES" ]]; then
-        extra_arg="-e delete_anyway='yes'"
-    fi
-    run_ansible ${socok8s_absolute_dir}/playbooks/generic-clean_userfiles.yml ${extra_arg:-}
-}
-function teardown(){
```
jhesketh: This is different to the openstack or common implementations.

script_library/deployment-actions-openstack.sh
```diff
@@ -36,66 +37,6 @@ function enroll_caasp_workers() {
     echo "Enrolling caasp worker nodes into the cluster"
     run_ansible ${socok8s_absolute_dir}/playbooks/generic-enroll_caasp_workers.yml
 }
-function patch_upstream(){
-    echo "Running dev-patcher"
-    echo "Nothing will happen if developer mode is not set"
-    run_ansible ${socok8s_absolute_dir}/playbooks/openstack-patch_upstream.yml
```
jhesketh: This playbook (like some of the others following) just calls the generic playbook. We should remove the file if they are no longer needed.

script_library/deployment-actions-kvm.sh
jsuchome: Yes, but I do not think adding `gather_logs` option into common part hurts.

script_library/deployment-actions-openstack.sh
jsuchome: OK

jsuchome:
@jhesketh updated

script_library/deployment-actions-kvm.sh
jhesketh: No, but this is doing clean_kvm, and the new function does clean_openstack

script_library/deployment-actions-common.sh
```diff
@@ -0,0 +1,81 @@
...
+    fi
+    run_ansible ${socok8s_absolute_dir}/playbooks/generic-clean_userfiles.yml ${extra_arg:-}
+}
+
+function teardown(){
+    if [[ ${SOCOK8S_GATHER_LOGS:-"NO"} == "YES" ]]
+    then
+        gather_logs
+    fi
+    clean_openstack
```
jhesketh: This shouldn't be generic

script_library/deployment-actions-kvm.sh
jsuchome: Ah, true, I missed that.

Itxaka:
looks good but...should we wait to merge this until we cut the stable release in case we break something?

toabctl:
retriggered CI.

jsuchome:
@Itxaka or let's wait until the CI is green :-)

jsuchome:
Fine. Better to use a tag then change request though ;-)

jsuchome:
Rebased, ready for new reviews

PR #449, opened 2019-06-05
Make it possible for the final task to fail

jesusaurus:
The final task is a debug task, which normally displays OK (in green if
color is enabled). This will mark the debug task as failed (in red) if
the message contains the word 'failed'.


jsuchome:
Yes, it failed the way it should:

https://ci.suse.de/blue/organizations/jenkins/cloud-socok8s-jobs%2Fsocok8s-integration/detail/PR-449/2/pipeline

PR #450, opened 2019-06-06
Fix conflciting endpoint definition in endpoints.yml for novncproxy

jgu17:
Fixed a conflicting endpoint definition for public and default for novncproxy
in endpoints.yaml. Both public an ddefault had the same value nova-novncproxy that Airship didnt like,

### DIFF
--- a/site/soc/software/config/endpoints.yaml
+++ b/site/soc/software/config/endpoints.yaml
@@ -960,7 +960,6 @@ data:
       name: nova
       hosts:
         default: nova-novncproxy
-        public: nova-novncproxy
       host_fqdn_override:
         default: null
         public:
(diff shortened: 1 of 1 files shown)


playbooks/roles/deploy-osh/tasks/main.yml
```diff
@@ -89,7 +89,7 @@
       {{ socok8s_ext_vip }} neutron-server.openstack.svc.cluster.local neutron-server
       {{ socok8s_ext_vip }} horizon.openstack.svc.cluster.local horizon
       {{ socok8s_ext_vip }} heat.openstack.svc.cluster.local heat
-      {{ socok8s_ext_vip }} nova-novncproxy.openstack.svc.cluster.local nova-novncproxy
+      {{ socok8s_ext_vip }} nova-novncproxy.openstack.svc.cluster.local novncproxy
```
jesusaurus: Why are we changing only the short name and not the full name?

jgu17: fixed

PR #452, opened 2019-06-06
Remove duplicate step list

toabctl:
Commit e7355ff4c910af28b428 added some help instructions when running
'run.sh' but it also duplicated the list of possible actions (steps).
Remove that duplicate list but still allow to see the help if
requested.

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -21,13 +21,13 @@ socok8s_absolute_dir="$( cd "$(dirname "$0")" ; pwd -P )"
 
 deployment_action=$1
 
-while true ; do
-    case "$deployment_action" in
-      -h|--help) usage ; exit 0 ;;
-       *) source ${scripts_absolute_dir}/pre-flight-checks.sh "validate_cli_options $1"
-          break ;;
-    esac
-done
+# be able to show the help. Later we have the same case (which includes
+# the help) but the pre-flight checks might already abort before
+# being able to show the help
+if [[ ${deployment_action} =~ -h|--help ]] ; then
+    usage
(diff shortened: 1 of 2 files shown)


run.sh
```diff
@@ -21,13 +21,15 @@ socok8s_absolute_dir="$( cd "$(dirname "$0")" ; pwd -P )"
...
-    case "$deployment_action" in
-      -h|--help) usage ; exit 0 ;;
-       *) source ${scripts_absolute_dir}/pre-flight-checks.sh "validate_cli_options $1"
-          break ;;
-    esac
-done
+# be able to show the help. Later we have the same case (which includes
+# the help) but the pre-flight checks might already abort before
+# being able to show the help
+case "$deployment_action" in
```
jsuchome: Why `case` for just one option?

toabctl: fixed

toabctl:
The downside of this PR is, that if you run a non-valid deployment action (eg `./run.sh xyz`) then you get the output from the preflight checks and other stuff before you actually see the `Invalid option foo, Check --help for valid options` message.

rajdharwadkar:
> The downside of this PR is, that if you run a non-valid deployment action (eg `./run.sh xyz`) then you get the output from the preflight checks and other stuff before you actually see the `Invalid option foo, Check --help for valid options` message.

Yes thats reason this is duplicated . User expects validation right at beginning not after seeing those output from source files.

Itxaka:
Lets close this for now, it can be restored later if needed

PR #453, opened 2019-06-06
CI: only run lint when playbook files have changed

Itxaka:
Provides 3 different vars to know when tox, playbooks or docs
have been modified so we can trigger the following steps according
to those.

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -1,3 +1,11 @@
+#!groovy
+
+// Global vars so they can be used in different steps
+DocsChanged = false
+PlaybooksChanged = false
+ToxChanged = false
+
+
 pipeline {
 
     options {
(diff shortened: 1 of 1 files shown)


Jenkinsfile.integration
```diff
@@ -37,8 +45,33 @@ pipeline {
...
+                        if (sourceChanged[i].contains("tox.ini")) {
+                            isToxChanged = true
+                        }
+                    }
+                }
+            }
+        }
 
         stage('Lint Ansible playbooks') {
+            when { expression { isPlaybooksChanged == true } }
```
Itxaka: Im not sure but this may be changed into `when { expression { return isPlaybooksChanged } }` and work the same

Itxaka: this works so I changed it

Itxaka:
shows the files changed properly:
```
[2019-06-06T14:06:21.199Z] Files changed for this PR:
[2019-06-06T14:06:21.199Z] Jenkinsfile.integration
```

And skips linting because there was no changes:
```
[Pipeline] {
Stage "Lint Ansible playbooks" skipped due to when conditional
[Pipeline] }
```

Itxaka:
You can see a more complete change https://ci.suse.de/job/cloud-socok8s-jobs/job/socok8s-integration/job/PR-453/7/consoleFull

Which has several files and changes the vars accordingly

Jenkinsfile.integration
```diff
@@ -1,3 +1,11 @@
+#!groovy
+
+// Global vars so they can be used in different steps
+isDocsChanged = false
+isPlaybooksChanged = false
```
Itxaka: is + plurals = bad engrish

Will change those to a more decent sounding vars

Itxaka:
> I am not sure if the conditional on line 49 should apply, but it's already an improvement.

So that is in there because when we check the master branch it makes no sense to run this, as the changelog will be empty anyways :)

It just saves us a few seconds, nothing else :)

PR #456, opened 2019-06-06
Update tempest config params to increase smoke test success rate

matthew-fuller:
This change incorporates tempest.conf changes recommended by QE to increase passing rate in tempest smoke tests.

This change also updates the network setup example in verify.rst to correct the name of the --provider-physical-network value from 'public' to 'external'.

### DIFF
--- a/site/soc/software/charts/osh/openstack-tempest/tempest.yaml
+++ b/site/soc/software/charts/osh/openstack-tempest/tempest.yaml
@@ -15,7 +15,7 @@ metadata:
   storagePolicy: cleartext
 data:
   wait:
-    timeout: 6000
+    timeout: 12000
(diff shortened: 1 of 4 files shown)


site/soc/software/charts/osh/openstack-tempest/tempest.yaml
```diff
@@ -27,9 +27,15 @@ data:
...
           tempest_roles: []
+          admin_domain_name: "Default"
+        identity:
+          admin_domain_scope: true
+          default_domain_id: "default"
         identity-feature-enabled:
           api_v2: false
           domain_specific_drivers: true
+        image:
+          http_image: "{{ cirros_test_image_url }}"
```
Itxaka: while this is ok here, in the CI I think it will fail. Dont we have an already setup public image due to changes tou the chats already?

PR #459, opened 2019-06-06
Update neutron l3 rootwrap filters and dhcp filters

sgrasley:
Missing python3 filters.


jgu17:
@sgrasley I included the commit into PR 443 due to the partial inclusion of the 443 in our commit.

PR #460, opened 2019-06-06
doc: add socok8s installation alternatives

csymons-suse:


csymons-suse:
SCO-install-Screenshot_20190606

csymons-suse:
> The requirements packages are installed automatically if rpm or iso. So manual install is required.

It's not clear to me how this should be handled in the documentation. Does this mean that required dependencies are handled automatically when rpm or is? And cloning the GitHub repository involves manually installing the requirements?

csymons-suse:
> Installed from iso or rpm put the socok8s at /usr/share/socok8s

Why is this important in the documentation? Seems like the user might like some context.

csymons-suse:
SCO-install-Screenshot_20190606_V2

jgu17:
> The requirements packages are installed automatically if rpm or iso. So manual install is required.
> 
> It's not clear to me how this should be handled in the documentation. Does this mean that required dependencies are handled automatically when rpm or is? And cloning the GitHub repository involves manually installing the requirements?

Yes.

jgu17:
> Installed from iso or rpm put the socok8s at /usr/share/socok8s
> 
> Why is this important in the documentation? Seems like the user might like some context.

For people not familiar with suse convention, it's helpful to inform the user where the installed folder is. QE for example, was looking for socok8s folder in the user home.

csymons-suse:
SCO-install-Screenshot_20190606_V3

doc/source/deployment/setup-deployer.rst
```diff
@@ -68,22 +68,37 @@ Deployer. Set up your workspace with the following steps:
...
-This repository uses submodules, so you need to get all the code to make sure
-the playbooks work.
+There are several ways to install the SUSE Containerized OpenStack software.
 
-::
+1. Install from the openSUSE repository including required dependencies:
 
-   git clone --recursive https://github.com/SUSE-Cloud/socok8s.git
+   a. sudo zypper addrepo \\
+      https://download.opensuse.org/repositories/Cloud:/socok8s/openSUSE_Leap_15.0/ socok8s
```
jhesketh: You need to add `-f` or `zypper refresh`

doc/source/deployment/setup-deployer.rst
```diff
@@ -68,22 +68,37 @@ Deployer. Set up your workspace with the following steps:
...
+   b. sudo zypper install socok8s (installs to /usr/share/socok8s)
 
-Alternatively, one can fetch/update the tree of the submodules by running:
+2. Install with an ISO image including required dependencies:
 
-::
+   a. Download openSUSE-Addon-socok8s-x86_64-Media.iso from
+      https://download.opensuse.org/repositories/Cloud:/socok8s/images/iso/
+   b. sudo zypper addrepo <ISO_IMAGE> socok8s-iso
```
jhesketh: ditto

doc/source/deployment/setup-deployer.rst
jsuchome: I think it's not technically necessary, as following `zypper install` should refresh new repo as well

jhesketh: ah, I thought it only did that when the last refresh was a certain age. It makes sense to do it on new repos though.

doc/source/deployment/setup-deployer.rst
```diff
@@ -68,22 +68,37 @@ Deployer. Set up your workspace with the following steps:
...
-::
+1. Install from the openSUSE repository including required dependencies:
 
-   git clone --recursive https://github.com/SUSE-Cloud/socok8s.git
+   a. sudo zypper addrepo \\
+      https://download.opensuse.org/repositories/Cloud:/socok8s/openSUSE_Leap_15.0/ socok8s
+   b. sudo zypper install socok8s (installs to /usr/share/socok8s)
 
-Alternatively, one can fetch/update the tree of the submodules by running:
+2. Install with an ISO image including required dependencies:
```
jgu17: One more nit: could we please move ISO as the first preferred approach we agreed with the PM and architect.

jsuchome: Right, the iso should be available elsewhere for download.
But I would think documentation could be adapted once we know the right location

jsuchome:
> 2) In the ISO and RPM method, inform the user where the socok8s install dir is.

I believe this part is included: _sudo zypper install socok8s (installs to /usr/share/socok8s)_

doc/source/deployment/setup-deployer.rst
csymons-suse: done, refresh added

doc/source/deployment/setup-deployer.rst
csymons-suse: ditto

PR #462, opened 2019-06-07
Add missing requirement

jhesketh:
Which isn't installed by default on minimal openSUSE installs such as
the container images.

### DIFF
--- a/doc/source/suse-ecp/ose-localhost.rst
+++ b/doc/source/suse-ecp/ose-localhost.rst
@@ -57,6 +57,7 @@ Base software
 Install the following software on your `localhost`:
 
   * jq
+  * which
   * ipcalc
   * git
   * python-virtualenv


Itxaka:
didnt Dirk dropped the dependency on which?

jhesketh:
Not sure. Not that I can see with a quick look in git, but it'd be easy to miss with all the docs reorgs.

Either way, it is a requirement as it's called from our bash scripts on localhost. I'm not sure we can assume it's installed, as it isn't inside the leap15 OCI container.

doc/source/suse-ecp/ose-localhost.rst
```diff
@@ -57,6 +57,7 @@ Base software
 Install the following software on your `localhost`:
 
   * jq
+  * which
```
jgu17: is this a newly introduced requirement?

jsuchome: No, it has been used quite some time, but it never failed because most of the installations have which already present. So as @jhesketh says it might be useful for some  minimal setups

jhesketh: The RPM has been updated to depend on which as well.

jhesketh:
This was opened before #537 and is no longer required.

PR #463, opened 2019-06-07
Preparating for caaspv4 and cleanup

toabctl:
This PR adds an extra step "run.sh deploy_caasp4". It creates a k8s cluster but it does not yet integrate with the rest of the deployment (so you can not install osh on top of the caasp4 cluster).

I think it would be good to get this merged and then work on the integration

### DIFF
--- a/playbooks/openstack-create_network.yml
+++ b/playbooks/openstack-create_network.yml
@@ -36,3 +36,33 @@
           socok8s_{{ item.output_key }}: {{ item.output_value }}
       with_items: "{{ network_stack_create_output.stack.outputs }}"
       when: ( item.output_key == 'dcm_vip') or ( item.output_key == 'ext_vip' )
+
+    - name: Add internal network ID to group_vars
+      blockinfile:
+        path: "{{ socok8s_workspace }}/inventory/group_vars/all/openstack-network.yml"
+        create: yes
+        marker: "# {mark} ANSIBLE MANAGED BLOCK FOR deploy_on_openstack_internal_network_id"
+        block: |
+          deploy_on_openstack_internal_network_id: {{ item.output_value }}
+      with_items: "{{ network_stack_create_output.stack.outputs }}"
+      when: ( item.output_key == 'network_id')
+
+    - name: Add internal subnet ID to group_vars
+      blockinfile:
+        path: "{{ socok8s_workspace }}/inventory/group_vars/all/openstack-network.yml"
+        create: yes
+        marker: "# {mark} ANSIBLE MANAGED BLOCK FOR deploy_on_openstack_internal_subnet_id"
+        block: |
+          deploy_on_openstack_internal_subnet_id: {{ item.output_value }}
+      with_items: "{{ network_stack_create_output.stack.outputs }}"
+      when: ( item.output_key == 'subnet_id')
+
+    - name: Add router ID to group_vars
+      blockinfile:
+        path: "{{ socok8s_workspace }}/inventory/group_vars/all/openstack-network.yml"
+        create: yes
+        marker: "# {mark} ANSIBLE MANAGED BLOCK FOR deploy_on_openstack_router_id"
+        block: |
+          deploy_on_openstack_router_id: {{ item.output_value }}
+      with_items: "{{ network_stack_create_output.stack.outputs }}"
+      when: ( item.output_key == 'router_id')
--- a/playbooks/openstack-create_network.yml
+++ b/playbooks/openstack-create_network.yml
@@ -66,3 +66,20 @@
           deploy_on_openstack_router_id: {{ item.output_value }}
       with_items: "{{ network_stack_create_output.stack.outputs }}"
       when: ( item.output_key == 'router_id')
+
+    - name: Reload group_vars openstack-network to have the router ID available
+      include_vars:
+        file: "{{ socok8s_workspace }}/inventory/group_vars/all/openstack-network.yml"
+
+    - name: Get router interface port ID
+      command: "openstack port list --router {{ deploy_on_openstack_router_id }} -f value -c ID"
+      register: _router_interface_port_id
+      changed_when: False
+
+    - name: Add router interface port ID to group_vars
+      blockinfile:
+        path: "{{ socok8s_workspace }}/inventory/group_vars/all/openstack-network.yml"
+        create: yes
+        marker: "# {mark} ANSIBLE MANAGED BLOCK FOR deploy_on_openstack_router_interface_id"
+        block: |
+          deploy_on_openstack_router_interface_id: {{ _router_interface_port_id.stdout }}
--- a/playbooks/openstack-delete_caasp4.yml
+++ b/playbooks/openstack-delete_caasp4.yml
@@ -1,0 +1,29 @@
+---
+#
+# (c) Copyright 2019 SUSE Linux GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+
+- hosts: localhost
+  gather_facts: yes
+  any_errors_fatal: true
+  vars_files:
+    - "{{ playbook_dir }}/../vars/common-vars.yml"
+    - "{{ playbook_dir }}/../vars/deploy-on-openstack.yml"
+  tasks:
+    - import_role:
+        name: caasp4
+        tasks_from: cleanup_nodes
+      vars:
+        skuba_ci_terraform_workspace: "{{ socok8s_workspace }}/skuba-terraform/"
--- a/playbooks/roles/caasp4/tasks/cleanup_nodes.yml
+++ b/playbooks/roles/caasp4/tasks/cleanup_nodes.yml
@@ -1,0 +1,39 @@
+---
+#
+# (c) Copyright 2019 SUSE Linux GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+
+# NOTE(toabctl): We import the resources (which are created by the network heat
+# stack in setup_nodes.yml so we need to remove the resources again from
+# terraform state and let heat handle the reource cleanup
+
+- name: Does the terraform state exist
+  stat:
+    path: "{{ skuba_ci_terraform_workspace }}/terraform.tfstate"
+  register: _terraform_state_file
+
+- name: Remove existing networking resources from terraform state
+  include_tasks: _cleanup_nodes_terraform_remove_resources.yml
+  when: _terraform_state_file.stat.exists
+
+- name: Destroy terraform plan
+  terraform:
+    project_path: "{{ skuba_ci_terraform_workspace }}"
+    state: absent
+
+- name: Drop skuba openstack terraform files
+  file:
+    path: "{{ skuba_ci_terraform_workspace }}"
+    state: absent
--- a/playbooks/roles/caasp4/tasks/setup_k8s.yml
+++ b/playbooks/roles/caasp4/tasks/setup_k8s.yml
@@ -1,0 +1,71 @@
+---
+#
+# (c) Copyright 2019 SUSE Linux GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+
+- name: Get skuba version
+  command: skuba version
+  changed_when: False
+  register: _skuba_version
+
+- name: Show skuba version
+  debug:
+    # NOTE: skuba is currently showing the version at stderr (see https://github.com/SUSE/skuba/issues/412)
+    msg: "{{ _skuba_version.stdout }}{{ _skuba_version.stderr }}"
+
+- name: Get terraform output variables
+  command: terraform output -json
+  args:
+    chdir: "{{ skuba_ci_terraform_workspace }}"
+  changed_when: False
+  register: _terraform_json_output
+
+- name: Set terraform json output fact
+  set_fact:
+    _terraform_json_output: "{{ _terraform_json_output.stdout|from_json }}"
+
+- name: Is skuba cluster dir available
+  stat:
+    path: "{{ skuba_cluster_basedir }}/{{ skuba_cluster_name }}"
+  register: _skuba_cluster_dir
+
+- name: Init skuba cluster
+  command: "skuba cluster init --control-plane {{ _terraform_json_output.ip_load_balancer.value }} {{ skuba_cluster_name }}"
+  args:
+    chdir: "{{ skuba_cluster_basedir }}"
+  when: not _skuba_cluster_dir.stat.exists
+
+- name: Get skuba cluster status
+  command: "skuba cluster status"
+  args:
+    chdir: "{{ skuba_cluster_basedir }}/{{ skuba_cluster_name }}"
+  register: _skuba_cluster_status
+  changed_when: False
+
+# TODO(toabctl): support more than one master
+- name: Bootstrap first master node
+  command: "skuba node bootstrap master-0 -s -t {{ _terraform_json_output.ip_masters.value[0] }} -u {{ skuba_ci_terraform_image_username }}"
+  args:
+    chdir: "{{ skuba_cluster_basedir }}/{{ skuba_cluster_name }}"
+  when: not _skuba_cluster_status.stdout | regex_search("^master-0.*", multiline=True)
+
+- name: Join worker nodes
+  command: "skuba node join worker-{{ idx }} -r worker -s -t {{ item }} -u {{ skuba_ci_terraform_image_username }}"
+  args:
+    chdir: "{{ skuba_cluster_basedir }}/{{ skuba_cluster_name }}"
+  loop: "{{ _terraform_json_output.ip_workers.value }}"
+  loop_control:
+    index_var: idx
+  when: not _skuba_cluster_status.stdout | regex_search('^worker-' ~ idx ~ '.*', multiline=True)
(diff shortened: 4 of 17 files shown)


toabctl:
Just thought it might be good to get this in already while working on caasp 4....

heat-templates/openstack-network.yml
```diff
@@ -64,14 +68,14 @@ resources:
       port_security_enabled: False
 
 outputs:
-  network_name:
-    description: Network name
+  network_id:
+    description: Network ID
```
jesusaurus: I went to look for places we need to update references to these outputs and couldn't find any. Are these three outputs (network, subnet, router) used anywhere?

toabctl: Not yet. But I don't want to remove them because the vars are needed when using terraform to deploy caasp 4

playbooks/roles/caasp4/tasks/cleanup_nodes.yml
```diff
@@ -0,0 +1,55 @@
...
+# terraform state and let heat handle the reource cleanup
+
+- name: Does the terraform state exist
+  stat:
+    path: "{{ socok8s_workspace }}/skuba-terraform/terraform.tfstate"
+  register: _terraform_state_file
+
+
+- name: Remove existing networking resources from terraform state
+  block:
```
evrardjp: nit: You don't need a block if you don't want to handle exceptions. Also this will result in a lot of "skipped" tasks, so you can probably do an include_tasks of the whole series of tasks from L28 to L55 when ``_terraform_state_file.stat.exists``. It would dynamically evaluate the include, and bypass some of the exec loop, so it would run faster, and is generally cleaner.

playbooks/roles/caasp4/tasks/setup_nodes.yml
```diff
@@ -0,0 +1,87 @@
...
+    image_name: "SLE-15-SP1-JeOS-GM"
+    # NOTE(toabctl):  net/subnet/router variable names must match the names from heat-templates/openstack-network
+    internal_net: "{{ socok8s_envname }}-net"
+    internal_subnet: "{{ socok8s_envname }}-subnet"
+    internal_router: "{{ socok8s_envname }}-router"
+    external_net: "{{ deploy_on_openstack_external_network }}"
+    stack_name: "{{ socok8s_envname }}-caasp4"
+    subnet_cidr: "{{ deploy_on_openstack_internal_subnet_cidr }}"
+    authorized_keys:
+      - "{{ lookup('file', '{{ ansible_env.HOME }}/.ssh/id_rsa.pub', errors='ignore') }}"
```
evrardjp: I would prefer exposing this as a var, which could be in the role defaults, with the current value. It would allow users to override it (should they have a different file, or want to ensure failures if absent).

playbooks/roles/caasp4/tasks/setup_nodes.yml
```diff
@@ -0,0 +1,87 @@
...
+    mode: 0775
+    force: no
+
+- name: Create a terraform variable file
+  template:
+    src: skuba-terraform.tfvars.j2
+    dest: "{{ socok8s_workspace }}/skuba-terraform/terraform.tfvars"
+  vars:
+    # TODO(toabctl): For now just use the image that the CaaSP team is using
+    image_name: "SLE-15-SP1-JeOS-GM"
```
evrardjp: I would prefer using this as a defaults/main.yml variable, so that people can override it if they want.

playbooks/roles/caasp4/templates/skuba-terraform.tfvars.j2
```diff
@@ -0,0 +1,60 @@
+# This file is managed by ansible and based on terraform.tfvars.ci.example from github.com/SUSE/skuba
+image_name = "{{ image_name }}"
+internal_net = "{{ internal_net }}"
+internal_subnet = "{{ internal_subnet }}"
+internal_router = "{{ internal_router }}"
+external_net = "{{ external_net }}"
+stack_name = "{{ stack_name }}"
+subnet_cidr = "{{ subnet_cidr }}"
+"dns_nameservers" = [
```
evrardjp: This should also be a var. Some people don't want to use google dns resolvers :p

playbooks/roles/caasp4/templates/skuba-terraform.tfvars.j2
```diff
@@ -0,0 +1,60 @@
...
+    "8.8.8.8",
+    "8.8.8.4"
+]
+masters = 1
```
evrardjp: I think this is fine being static, but we might need to think about that later.

playbooks/roles/caasp4/templates/skuba-terraform.tfvars.j2
```diff
@@ -0,0 +1,60 @@
...
+workers = 2
```
evrardjp: are 2 workers enough? We are using 3 right now. Might be worth using deploy_on_openstack_caasp_workers.

playbooks/roles/caasp4/templates/skuba-terraform.tfvars.j2
```diff
@@ -0,0 +1,60 @@
...
+master_size = "m1.medium"
+worker_size = "m1.medium"
```
evrardjp: Is that large enough? Can those two be a default role var?

playbooks/roles/caasp4/templates/skuba-terraform.tfvars.j2
```diff
@@ -0,0 +1,60 @@
...
+
+username = "sles"
+password = "linux"
```
evrardjp: Same applies on vars here.

playbooks/roles/caasp4/templates/skuba-terraform.tfvars.j2
```diff
@@ -0,0 +1,60 @@
...
+
+packages = [
+  "kernel-default",
+  "-kernel-default-base",
+  "ca-certificates-suse",
+  "patterns-caasp-Node",
+]
+
+authorized_keys = [
+{% for key in authorized_keys %}
```
evrardjp: probably easier to write:

``` {{ authorized_keys | join(',') }}```

File: playbooks/roles/caasp4/templates/skuba-terraform.tfvars.j2 [hunk 11]
```diff
@@ -0,0 +1,60 @@
...
+  "{{ key }}",
+{% endfor %}
+]
+
+ntp_servers = [
+{% for server in ntp_servers %}
```
evrardjp: same as authorized keys

playbooks/roles/caasp4/templates/skuba-terraform.tfvars.j2
toabctl: Removed the nameservers. I think theses should be added (if needed) to our network heat stack when creating the subnet.

playbooks/roles/caasp4/templates/skuba-terraform.tfvars.j2
toabctl: agreed

playbooks/roles/caasp4/templates/skuba-terraform.tfvars.j2
toabctl: I think it's not the same. I want a new line per entry

playbooks/roles/caasp4/templates/skuba-terraform.tfvars.j2
toabctl: I think it's not the same. I want a new line per entry

playbooks/roles/caasp4/templates/skuba-terraform.tfvars.j2
toabctl: That's what the caasp team uses in the CI. added as default role var

playbooks/roles/caasp4/templates/skuba-terraform.tfvars.j2
toabctl: 2 is the number that the caasp team uses. Are you ok with leaving this for now to see how it goes?

playbooks/roles/caasp4/tasks/setup_nodes.yml
toabctl: Done

playbooks/roles/caasp4/tasks/setup_nodes.yml
toabctl: Done

playbooks/roles/caasp4/tasks/cleanup_nodes.yml
toabctl: Done. Thanks for the hint!

playbooks/roles/caasp4/templates/skuba-terraform.tfvars.j2
toabctl: Done

playbooks/roles/caasp4/tasks/setup_nodes.yml
```diff
@@ -0,0 +1,97 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+
+- name: Copy skuba openstack terraform files to socok8s workspace
+  copy:
+    src: "{{ playbook_dir }}/../submodules/skuba/ci/infra/openstack/"
```
jesusaurus: Did you intend to add this as a submodule?

playbooks/roles/caasp4/templates/skuba-terraform.tfvars.j2
jesusaurus: We currently use the first two caasp workers for the airship control plane, and the third caasp worker is the airship compute worker. Do we have a reason to ignore `deploy_on_openstack_caasp_workers`?

playbooks/roles/caasp4/tasks/setup_nodes.yml
toabctl: Yes (see the commit messages). Currently skuba is a private repo so I can not add it as a submodule. But the skuba maintainers plan to make it public (in the next week(s)) and then we can switch to a submodule.

jsuchome: I would assume ideally we'd like to install it from package...

toabctl: If we want to use the terraform form, I guess we need to use a submodule or something similar. Not sure if terraform forms will be available via packages.
For the `skuba` executable, I agree (but for now I would just build it from git)

playbooks/roles/caasp4/templates/skuba-terraform.tfvars.j2
toabctl: ok. Fixed and it uses `deploy_on_openstack_caasp_workers` now

playbooks/roles/caasp4/templates/skuba-terraform.tfvars.j2
evrardjp: Got it.

playbooks/roles/caasp4/templates/skuba-terraform.tfvars.j2
evrardjp: Got it. It's cleaner this way then.

playbooks/openstack-create_network.yml
```diff
@@ -19,27 +19,70 @@
...
         - name: show network stack output
           debug:
             var: network_stack_create_output
 
-    - name: Get the internal VIP from the network stack
-      command: openstack stack output show {{ deploy_on_openstack_network_stackname }} dcm_vip -c output_value -f value
-      changed_when: false
-      register: _network_stack_dcm_vip_output
+    - name: Add vip with cidr in extravars
+      delegate_to: localhost
```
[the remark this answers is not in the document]
toabctl: I don't think it's needed. copied from the task before

playbooks/openstack-create_network.yml
```diff
@@ -19,27 +19,70 @@
...
+      blockinfile:
```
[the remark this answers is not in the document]
toabctl: I like the blockinfile because it adds this nice marker so you know that theses lines are edited by ansible (you don't know that with `lineinfile`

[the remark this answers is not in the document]
evrardjp: ok for me.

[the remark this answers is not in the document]
jesusaurus: Could we combine all of the openstack IDs into a single block? That would give us better signal-to-noise in the target file since each block comes with two comment lines.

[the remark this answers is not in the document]
toabctl: That doesn't work because we loop over all output entries for every block.

playbooks/openstack-create_network.yml
```diff
@@ -19,27 +19,66 @@
...
     - name: Add vip with cidr in extravars
       delegate_to: localhost
       blockinfile:
         path: "{{ socok8s_extravars }}"
         create: yes
-        marker: "# {mark} ANSIBLE MANAGED BLOCK FOR VIP"
+        marker: "# {mark} ANSIBLE MANAGED BLOCK FOR {{ item.output_key }}"
+        block: |
+          socok8s_{{ item.output_key }}: {{ item.output_value }}
+      with_items: "{{ network_stack_create_output.stack.outputs }}"
```
evrardjp: nit1: you should use loop instead of with_.* it will be deprecated soon. But we can do that in another patch.
nit2: for this case, you could even loop on something different than the list lookup (fetch 
 only the subelements matching what you need without requiring a "when" condition). This is just an improvement.

playbooks/openstack-create_network.yml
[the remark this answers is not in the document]
evrardjp: Well we could technically have a marker, and then add lineinfiles based on previous expected lines. But it's not important for me.

toabctl:
Some cleanup commits are now also available at https://github.com/SUSE-Cloud/socok8s/pull/563 so we can merge theses commits first to make the review here simpler.

toabctl:
> I don't see a port being created in that commit at all. It really just fetches the id.

Ah. That was a leftover of a previous approach. Solved.

toabctl:
> You are right. The CI code is also in the package. See https://build.suse.de/package/view_file/Devel:CaaSP:4.0/skuba/skuba.spec?expand=1

Hm. But then we would need to install the package to the CI node (which does setup the cloud nodes via terraform)

rhafer:
> Hm. But then we would need to install the package to the CI node (which does setup the cloud nodes via terraform)

Yeah, right. But that's what you're doing now as well (just from git instead of packages). Not sure which approach is better.

In the end I am fine with keeping it as it is for now. And maybe revisit later.

rhafer:
You might want to rebase again. As the commit to bump the automation submodule got merged meanwhile

toabctl:
> You might want to rebase again. As the commit to bump the automation submodule got merged meanwhile

Rebase done

playbooks/openstack-delete_caasp4.yml
```diff
@@ -0,0 +1,29 @@
...
+  any_errors_fatal: true
+  vars_files:
+    - "{{ playbook_dir }}/../vars/common-vars.yml"
+    - "{{ playbook_dir }}/../vars/deploy-on-openstack.yml"
+  tasks:
+    - import_role:
+        name: caasp4
+        tasks_from: cleanup_nodes
+      vars:
+        - skuba_ci_terraform_workspace: "{{ socok8s_workspace }}/skuba-terraform/"
```
evrardjp: do you really want vars to be an unnamed list?

toabctl: @evrardjp No. Fixed (hopefully) in all the commits included here

run.sh
```diff
@@ -65,8 +65,15 @@ case "$deployment_action" in
         deploy_ses
         ;;
     "deploy_caasp")
+        # this is CaaSP 3
         deploy_caasp
         ;;
+    "deploy_caasp4")
+        # this is CaaSP 4
+        source ${scripts_absolute_dir}/pre-flight-checks.sh check_caasp4_skuba_available
+        source ${scripts_absolute_dir}/pre-flight-checks.sh check_caasp4_terraform_available
```
jesusaurus: Our other pre-flight checks are called from the deployment actions scripts, can we move these there as well for consistency?

run.sh
```diff
@@ -139,6 +146,11 @@ case "$deployment_action" in
     "teardown")
         teardown
         ;;
+    "clean_caasp4")
+        source ${scripts_absolute_dir}/pre-flight-checks.sh check_caasp4_skuba_available
+        source ${scripts_absolute_dir}/pre-flight-checks.sh check_caasp4_terraform_available
```
jesusaurus: Same as above

playbooks/roles/caasp4/tasks/setup_k8s.yml
```diff
@@ -0,0 +1,70 @@
...
+#
+
+- name: Get skuba version
+  command: skuba version
+  changed_when: False
+  register: _skuba_version
+
+- name: Show skuba version
+  debug:
+    var: _skuba_version.stdout
```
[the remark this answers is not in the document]
toabctl: oh. Wow. Fixed and created https://github.com/SUSE/skuba/issues/412

run.sh
toabctl: Fixed

run.sh
toabctl: Fixed

playbooks/roles/caasp4/templates/skuba-terraform.tfvars.j2
```diff
@@ -0,0 +1,46 @@
+# This file is managed by ansible and based on terraform.tfvars.ci.example from github.com/SUSE/skuba
+image_name = "{{ image_name }}"
+internal_net = "{{ internal_net }}"
+internal_subnet = "{{ internal_subnet }}"
+internal_router = "{{ internal_router }}"
+external_net = "{{ external_net }}"
+stack_name = "{{ stack_name }}"
+subnet_cidr = "{{ subnet_cidr }}"
+"dns_nameservers" = [
```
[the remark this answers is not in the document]
toabctl: `dns_nameservers` got removed from terraform (see https://github.com/SUSE/skuba/pull/369 ) so this was a leftover. Dropped now.

playbooks/roles/caasp4/tasks/setup_nodes.yml
```diff
@@ -0,0 +1,88 @@
...
+    dest: "{{ skuba_ci_terraform_workspace }}"
```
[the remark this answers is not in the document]
toabctl: Cleanup is done now in the playbook

playbooks/roles/caasp4/tasks/setup_k8s.yml
```diff
@@ -0,0 +1,70 @@
...
+    path: "{{ skuba_cluster_basedir }}/{{ skuba_cluster_name }}"
+  register: _skuba_cluster_dir
+
+- name: Init skuba cluster
+  command: "skuba cluster init --control-plane {{ _terraform_json_output.ip_load_balancer.value }} {{ skuba_cluster_name }}"
+  args:
+    chdir: "{{ skuba_cluster_basedir }}"
+  when: not _skuba_cluster_dir.stat.exists
+
+- name: Get skuba cluster status
```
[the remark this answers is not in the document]
toabctl: I ran into the same problem now. Looking how to fix it...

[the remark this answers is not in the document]
toabctl: Seems to be an issue with `skuba`. I updated the skuba git repo and rebuild (with `make`) the binary and then it works

toabctl:
@jesusaurus I think I addressed all your comments. Please re-review and feel free to press the merge button if this looks ok to you.

toabctl:
rebased...

playbooks/roles/caasp4/tasks/setup_k8s.yml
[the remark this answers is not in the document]
jesusaurus: This is still happening for me with the latest version of skuba: `skuba version: 0.6.1 (development) d55e3f9 20190621 go1.11.10`

PR #464, opened 2019-06-07
Save shipyard env values to .profile to help later debugging

jsuchome:
These variables are used by tools/shipyard.sh script and as
we already know their values, we might as well provide them to
users for possible later use.

This requires https://review.opendev.org/#/c/663698/


jsuchome:
@evrardjp Why is ansible complaining?
AFAIK template does not have to be YAML file, it seems like it is expecting one...

Itxaka:
>  AFAIK template does not have to be YAML file, it seems like it is expecting one...

Actually its your ansible code the one complaining, not the template ;)

playbooks/roles/airship-setup-deployer/tasks/main.yml
```diff
@@ -117,3 +117,8 @@
   template:
     src: clouds.yml.j2
     dest: /etc/openstack/clouds.yaml
+
+- name: Save Shipyard env variables to ~/.profile
+  template:
+    src: profile.j2
+    dest: dest: "{{ ansible_user_dir }}/.profile"
```
Itxaka:
```suggestion
    dest: "{{ ansible_user_dir }}/.profile"
```

jsuchome: Oh, I must be blind. Thanks!

playbooks/roles/airship-setup-deployer/tasks/main.yml
```diff
@@ -116,3 +116,8 @@
   template:
     src: clouds.yml.j2
     dest: /etc/openstack/clouds.yaml
+
+- name: Save Shipyard env variables to ~/.profile
+  template:
+    src: profile.j2
+    dest: "{{ ansible_user_dir }}/.profile"
```
evrardjp: Do we really want to override a user's profile?
That's very intrusive.

Are you sure you don't want to ensure those lines are part of the profile instead?
For that, you can use "lineinfile" module.

jsuchome: Yes, good point, I've actually realized yesterday as well. I'll change it

playbooks/roles/airship-setup-deployer/tasks/main.yml
```diff
@@ -116,3 +116,10 @@
   template:
     src: clouds.yml.j2
     dest: /etc/openstack/clouds.yaml
+
+- name: Save Shipyard env variables to ~/.profile
+  blockinfile:
+    path: "{{ ansible_user_dir }}/.profile"
```
rhafer: I think `blockinfile` will not create the file if it doesn't exist yet by default. Also I think we should not be putting passwords into `.profile` (it often is word-readable). How about creating some `.airshiprc`  (properly protected)? And either source that from `.profile` or drop a file into `/etc/profile.d` that sources it if it exists. Or not source it at all automatically and leave it as an excerise for the user to do so.

I know that this is a bit moot as long as the deployer is still some well separated stand-alone node. But this might change in the future.

jsuchome: I like e.g. `/etc/profile.d/airship.sh`. But that would make it available for all users, right?

jsuchome: So `.airshiprc` in `"{{ ansible_user_dir }}` seems to be safest bet I think

jsuchome: And with custom file like this I could go back to using template instead of `blockinfile`

rhafer: > I like e.g. `/etc/profile.d/airship.sh`. But that would make it available for all users, right?

Well. In theory that file could just source `~/.airshiprc` if it exists, the it would keep the password "secret" and only work for users that actually have that file. But I am rather unsure if that is really needed at the current stage.

> So `.airshiprc` in `"{{ ansible_user_dir }}` seems to be safest bet I think

Yes

> And with custom file like this I could go back to using template instead of `blockinfile`

Yes. (Note though, that `blockinfile` has an option that allows to create the file if it doesn't exist)

jsuchome: OK, done (with blockinfile as I was lazy to look up the template again)

playbooks/roles/airship-setup-deployer/tasks/main.yml
```diff
@@ -116,3 +116,12 @@
   template:
     src: clouds.yml.j2
     dest: /etc/openstack/clouds.yaml
+
+- name: Save Shipyard env variables to ~/.airshiprc
+  blockinfile:
+    path: "{{ ansible_user_dir }}/.airshiprc"
+    create: yes
+    mode: "0700"
```
rhafer: nit: As this is supposed to be sourced, not executed, `0600` should be enough here

jsuchome: OK

playbooks/roles/airship-setup-deployer/tasks/main.yml
```diff
@@ -116,3 +116,12 @@
   template:
     src: clouds.yml.j2
     dest: /etc/openstack/clouds.yaml
+
+- name: Save Shipyard env variables to ~/.airshiprc
+  blockinfile:
+    path: "{{ ansible_user_dir }}/.airshiprc"
```
evrardjp: I would rather use the workspace instead. ``socok8s_workspace`` so that we continue to be self contained.

jsuchome: workspace is on localhost, but I want this on deployer

evrardjp: Got it.

jsuchome:
As for "leaking of secrets", we're already saving OS_PASSWORD into clouds.yaml on the deployer node. I do not see a problem of saving OS_SHIPYARD_PASSWORD into the file that is only readable by ansible user.

Actually this (= saving the password on deployer) is the very reason I created this PR, so that the debugging and development is easier and user does not have to dig it manually from secrets/ucp_shipyard_keystone_password

Itxaka:
is this still needed @jsuchome ?

jsuchome:
I still think it makes sense (but it would require @evrardjp to change his vote).

Although ... if Airship 2.0 will look completely differently than this would not be used I guess

jsuchome:
rebased

PR #465, opened 2019-06-07
Doc change for recover controller and compute host node.

rajdharwadkar:
Add in troubleshooting section . to recover  controller and compute node if it entered in maintenance mode during deployment.

### DIFF
--- a/doc/source/operations.rst
+++ b/doc/source/operations.rst
@@ -339,6 +339,51 @@ instance:
 
    kubectl logs -n ${NAMESPACE} ${POD_NAME} -p
 
+Recover Controller Host Node
+----------------------------
+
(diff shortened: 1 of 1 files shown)


doc/source/operations.rst
```diff
@@ -339,6 +339,51 @@ instance:
...
+
+   btrfs property set -ts /.snapshots/$mounted_snapshot/snapshot ro true
+
+   reboot
+
+Recover Compute Host Node
+-------------------------
+
+If deployment failed with error of compute host not reachable ( has entered maintenance mode )
+
```
csymons-suse: same as above.

rajdharwadkar: > same as above.

Same as what ?

PR #473, opened 2019-06-10
Correcting domain name as identified in SCRD-9468

kantarun:
Looks like current functionality is working because system has default
domain defined and that is used instead.


Itxaka:
change is not needed anymore, closing this

PR #475, opened 2019-06-10
Disable tempest cleanup and ensure tempest run always returns 0

matthew-fuller:
This change is not critical, but it's nice to have.

First, since 'tempest cleanup' immediately throws an exception when trying to list roles using a v2 URL, cleanup is disabled to keep this error out of the tempest logs.

Second, since any failed test during tempest run will cause the container to terminate with an exit code of 1, k8s moves the pod to 'Error' status and the shipyard workflow will attempt to retry the installation multiple times. This creates the undesirable behavior where the tempest test run is completely finished, but the ansible playbook will continue to wait for the shipyard workflow to enter 'Completed' status, which could take > 45 mins. This change ensures that 'tempest run' always exits with 0 so shipyard will know the deployment is completed.

It should be noted that making the 'tempest run' command exit with 0 should not mask any underlying errors. All container errors and tempest test failures will still be captured in the logs.

### DIFF
--- a/site/soc/software/charts/osh/openstack-tempest/tempest.yaml
+++ b/site/soc/software/charts/osh/openstack-tempest/tempest.yaml
@@ -29,7 +29,10 @@ data:
         -w {{ tempest_workers }} \
         {{ tempest_test_args[tempest_test_type] }} \
 {% if use_blacklist %}
-        --blacklist-file /etc/tempest/test-blacklist
+        --blacklist-file /etc/tempest/test-blacklist \
+{% endif %}
+        || true
+{% if use_blacklist %}
       blacklist:
         {{ lookup('file', 'tempest_blacklist') | indent(8) }}
 {% endif %}
@@ -66,5 +69,7 @@ data:
           glance: "{{ tempest_enable_glance_service }}"
           nova: "{{ tempest_enable_nova_service }}"
           neutron: "{{ tempest_enable_neutron_service }}"
+      cleanup:
+        enabled: false
 {% endif %}  
 ...
--- a/doc/source/deployment/verify.rst
+++ b/doc/source/deployment/verify.rst
@@ -35,15 +35,15 @@ shell on the `Deployer` node.
 .. code-block:: console
 
    export OS_CLOUD=openstack
-   openstack network create --provider-network-type flat --provider-physical-network external \ 
+   openstack network create --provider-network-type flat --provider-physical-network external \
      --external public
    openstack subnet create --network public --subnet-range 192.168.100.0/24 --allocation-pool \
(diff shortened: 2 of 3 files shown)


site/soc/software/charts/osh/openstack-tempest/tempest.yaml
```diff
@@ -29,7 +29,10 @@ data:
         -w {{ tempest_workers }} \
         {{ tempest_test_args[tempest_test_type] }} \
 {% if use_blacklist %}
-        --blacklist-file /etc/tempest/test-blacklist
+        --blacklist-file /etc/tempest/test-blacklist \
+{% endif %}
+        || true
```
Itxaka: Im wondering, if we use the blacklist to skip those failed tests, we dont need to exit 0 because the tempest run will work as expected so can we make this a if-else? either we are using the blacklist and we expect the exit code to be the real code or we dont and then we exit 0 as to not block the depoloyment.

I think this is a bad idea still, we should make sure that we blacklist those failing tests that we cannot do anything about *always* and let tempest exit with the real status because if not, we are hiding any problems. We cannot consider deploying tempest if we are gonna ignore the actual state of the cloud tests.

matthew-fuller: Maybe this is me butting heads with the design strategy then. To me, a failed tempest test should not equal a failed pod / failed release. The tempest test pod has (essentially) a single job to do: run all the tempest tests defined for that run, and that's it. If it encounters problems that prevent it from running tests (can't connect to host, can't locate config file, can't write to file system, etc.) then that's a legitimate error and the pod should be moved to 'Error' state because it couldn't complete its job and the tests didn't run. But if all tests were run to completion, then the pod has accomplished its job and it makes sense to me to move it to 'Completed' state. 

If individual tempest tests fail, that shouldn't mean the deployment of the tempest release has failed. A failed test here seems like expected, or at least reasonable, behavior. This might be getting too abstract, but if I think of a group of kids taking an exam, no one would claim the exam didn't happen or the examination process had failed if a couple of kids failed their tests. Maybe if the exam room was locked, the proctor didn't show up, there was a fire drill, etc., then sure, but a couple kids failing their tests is an expected outcome that's independent of the success or failure of the exam process itself.

Okay, I'll shut up now. I definitely see your point, but I wish the state of the pod was totally separate from the state of the individual tests.

Itxaka: Ummm I see, that is a good point. Maybe Im just using the other pods as example, like when launching neutron if neutron fails, pod fails, but indeed this may be a different case as technically tempest has run as expected....

Ok, seems good, Im convinced :+1: , I guess we should also tackle this upstream somewhow...

PR #476, opened 2019-06-11
Catch more failure messages

jesusaurus:
Fail if the shipyard output begins with "Error" or "Traceback"

### DIFF
--- a/playbooks/roles/airship-deploy-osh/tasks/main.yml
+++ b/playbooks/roles/airship-deploy-osh/tasks/main.yml
@@ -183,8 +183,13 @@
         SHIPYARD_IMAGE: "{{ shipyard_image }}"
         OS_PASSWORD: "{{ lookup('password', secrets_location + '/ucp_shipyard_keystone_password ' + password_opts) }}"
       register: shipyard_desc_action
-      failed_when: "'failed' in shipyard_desc_action.stdout"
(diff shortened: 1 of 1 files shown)


playbooks/roles/airship-deploy-osh/tasks/main.yml
```diff
@@ -177,7 +177,10 @@
     SHIPYARD_IMAGE: "{{ shipyard_image }}"
     OS_PASSWORD: "{{ lookup('password', secrets_location + '/ucp_shipyard_keystone_password ' + password_opts) }}"
   register: shipyard_desc_action
-  until: shipyard_desc_action.stdout.find('Processing') < 0 and shipyard_desc_action.stdout.find('running') < 0
+  until:
+    - shipyard_desc_action.stdout.find('Processing') < 0
+    - shipyard_desc_action.stdout.find('running') < 0
+    - shipyard_desc_action.stdout.find('Error') != 0
```
jgu17: why has to be Error?

jesusaurus: There may be "Errors: 0" near the end of the output, so we need to limit this to only when "Error" is at the very beginning of the output.

Itxaka:
Very nice, this will fix some of the fake pass that we are encountering at the gates. Can you rebase this @jesusaurus ?

PR #477, opened 2019-06-11
Apply merged patches as well

jsuchome:
If a patch has been merged upstream, specific revision should still be applied to the code, because git repos we are using should be frozen.

This code should download the patch whether it is merged or not and try to apply it. If merging fails now, the reason is we're using different version of repo and should fix it. Otherwise it will apply even the merged patch because the repo should be frozen so it should not hurt.

### DIFF
--- a/playbooks/roles/dev-patcher/tasks/patch-repos.yml
+++ b/playbooks/roles/dev-patcher/tasks/patch-repos.yml
@@ -9,7 +9,7 @@
 
 - name: Get patch {{ patchnumber }} details
   uri:
-    url: "{{ dev_patcher_baseurl }}/?q=is:open+{{ patchnumber }}&o=ALL_REVISIONS"
+    url: "{{ dev_patcher_baseurl }}/?q={{ patchnumber }}&o=ALL_REVISIONS"
     return_content: yes
     body_format: json
   register: thispatch
@@ -17,21 +17,19 @@
 # Gerrit starts its http response with ")]}'". Strip that to have a valid json.
 - name: Parse patch details
   set_fact:
-    jsoncontent: "{{ thispatch.content.replace(pattern_to_replace,'') | from_json }}"
+    jsoncontent: "{{ thispatch.content.replace(pattern_to_replace,'') | from_json | first }}"
   vars:
     pattern_to_replace: ")]}'"
 
-- name: Apply {{ patchnumber }} - {{ jsoncontent[0]['subject'] }}
+- name: Apply {{ patchnumber }} - {{ jsoncontent['subject'] }}
   shell: |
     set -o errexit
-    git fetch {{ jsoncontent[0]['revisions'][revision]['fetch']['anonymous http']['url'] }} {{ jsoncontent[0]['revisions'][revision]['fetch']['anonymous http']['ref'] }}
+    git fetch {{ jsoncontent['revisions'][revision]['fetch']['anonymous http']['url'] }} {{ jsoncontent['revisions'][revision]['fetch']['anonymous http']['ref'] }}
     git cherry-pick FETCH_HEAD
   args:
     chdir: "{{ upstream_repos_clone_folder }}/{{ project }}"
   vars:
     # Use latest revision of no revision was specified
-    revision: "{{ (patchrevision != 'current') | ternary(patchrevision, jsoncontent[0]['current_revision']) }}"
-    project: "{{ jsoncontent[0]['project'] }}"
-  when:
-    # Do not apply patches if they are already merged
-    - "jsoncontent | length > 0"
+    revision: "{{ (patchrevision != 'current') | ternary(patchrevision, jsoncontent['current_revision']) }}"
+    project: "{{ jsoncontent['project'] }}"
+  changed_when: true


playbooks/roles/dev-patcher/tasks/patch-repos.yml
```diff
@@ -9,29 +9,26 @@
...
     chdir: "{{ upstream_repos_clone_folder }}/{{ project }}"
   vars:
     # Use latest revision of no revision was specified
-    revision: "{{ (patchrevision != 'current') | ternary(patchrevision, jsoncontent[0]['current_revision']) }}"
-    project: "{{ jsoncontent[0]['project'] }}"
-  when:
-    # Do not apply patches if they are already merged
-    - "jsoncontent | length > 0"
+    revision: "{{ (patchrevision != 'current') | ternary(patchrevision, jsoncontent['current_revision']) }}"
+    project: "{{ jsoncontent['project'] }}"
```
Itxaka: either we check if it has been applied and we skip it in that case (when) or you need to add the tag to skip the linter

jsuchome: Yeah. I'm currently more worried that i can't get it work :-)

playbooks/roles/dev-patcher/tasks/patch-repos.yml
```diff
@@ -9,29 +9,27 @@
 
 - name: Get patch {{ patchnumber }} details
   uri:
-    url: "{{ dev_patcher_baseurl }}/?q=is:open+{{ patchnumber }}&o=ALL_REVISIONS"
+    url: "{{ dev_patcher_baseurl }}/?q=is:open+{{ patchnumber }}&q=is:merged+{{ patchnumber }}&o=ALL_REVISIONS"
```
jhesketh: (nit) could drop both these query parts
`url: "{{ dev_patcher_baseurl }}/?q={{ patchnumber }}&o=ALL_REVISIONS"

jsuchome: Oh. And I thought I tested it first :-)

playbooks/roles/dev-patcher/tasks/patch-repos.yml
```diff
@@ -9,29 +9,27 @@
...
   vars:
     # Use latest revision of no revision was specified
-    revision: "{{ (patchrevision != 'current') | ternary(patchrevision, jsoncontent[0]['current_revision']) }}"
-    project: "{{ jsoncontent[0]['project'] }}"
-  when:
-    # Do not apply patches if they are already merged
-    - "jsoncontent | length > 0"
+    revision: "{{ (patchrevision != 'current') | ternary(patchrevision, jsoncontent['current_revision']) }}"
+    project: "{{ jsoncontent['project'] }}"
+  changed_when: false
```
evrardjp: Can you remove ``changed_when: false``? We will actively patch the code, and saying this hasn't changed is wronging user expectations.

playbooks/roles/dev-patcher/tasks/patch-repos.yml
```diff
@@ -9,29 +9,27 @@
 
 - name: Get patch {{ patchnumber }} details
   uri:
-    url: "{{ dev_patcher_baseurl }}/?q=is:open+{{ patchnumber }}&o=ALL_REVISIONS"
+    url: "{{ dev_patcher_baseurl }}/?q={{ patchnumber }}&o=ALL_REVISIONS"
```
jesusaurus: I think it would be better to change ALL_REVISIONS to CURRENT_REVISION here so that we don't need the "first" filter below.

jsuchome: ALL_REVISIONS is there for a reason, so that we can track same revision of a patch although it changs upstream

playbooks/roles/dev-patcher/tasks/patch-repos.yml
jsuchome: OK, I will change it to `true`

playbooks/roles/dev-patcher/tasks/patch-repos.yml
jsuchome: See also https://github.com/SUSE-Cloud/socok8s/pull/395

jesusaurus: Thanks for the context

playbooks/roles/dev-patcher/tasks/patch-repos.yml
```diff
@@ -9,29 +9,27 @@
...
   vars:
     # Use latest revision of no revision was specified
-    revision: "{{ (patchrevision != 'current') | ternary(patchrevision, jsoncontent[0]['current_revision']) }}"
-    project: "{{ jsoncontent[0]['project'] }}"
-  when:
-    # Do not apply patches if they are already merged
-    - "jsoncontent | length > 0"
+    revision: "{{ (patchrevision != 'current') | ternary(patchrevision, jsoncontent['current_revision']) }}"
+    project: "{{ jsoncontent['project'] }}"
+  changed_when: true
```
kantarun: Why we need to highlight that its **always** changed? Any specific reason ?

If repo content remains same as before this task (possible when there is no change in patch from what is present in local version), why we need to highlight that.

jsuchome: - Without `changed_when`, lint complains
- I had `changed_when: false` before, but @evrardjp complained that it's wrong to say nothing has changed when we are doing changes
- And I probably agree that `changed_when: true` has more sense, because we are **always** patching

I'm not sure I see the situation when _repo content remains same as before this task_ . The reason of this whole change is, "if there's a patch that should be applied, apply it. If it fails something is wrong with the setup". So we always want to apply the patch that is in the list of patches

PR #478, opened 2019-06-11
Switch CaaSP workers to a larger flavor

rhafer:
With the current default (m1.large, i.e. 8GB of RAM) deployments run out
of memory from time to time, with the OOM killer starting to kill
processes. Let's switch to a bigger flavor (m1.xlarge, i.e. 16GB of RAM)
to avoid that.

### DIFF
--- a/files/caasp-stack.yaml
+++ b/files/caasp-stack.yaml
@@ -81,7 +81,7 @@ parameters:
   worker_flavor:
     type: string
     description: Worker Flavor
-    default: m1.large
+    default: m1.medlarge
     constraints:
       - custom_constraint: nova.flavor
   worker_count:


rhafer:
> Do we have enough capacity to handle this? 

I can't really answer that. I guess we have to try an see what happens.

> Otherwise we may have to limit the number of concurrent jobs.

We already limit the number of CI jobs (to 7 currently)

jhesketh:
> > Do we have enough capacity to handle this?
> 
> I can't really answer that. I guess we have to try an see what happens.
> 
> > Otherwise we may have to limit the number of concurrent jobs.
> 
> We already limit the number of CI jobs (to 7 currently)

Right, I mean limit further ;-)

rhafer:
He have "m1.medlarge" now. Switching to that. It should provide enough RAM for our deployments for now.

rhafer:
> Right, I mean limit further ;-)

I don't think we should do that for now. Let's try with the new flavor for a while and adjust the queue if really needed.

jhesketh:
medlarge does sound smaller than large fwiw

nicolasbock:
> medlarge does sound smaller than large fwiw

I can name it `medxlarge` :wink:

PR #480, opened 2019-06-11
Remove dvr from api_extensions list in tempest.conf

matthew-fuller:
This change removes 'dvr' from the list of supported api_extensions in tempest.conf.


toabctl:
@matthew-fuller can you describe the problem you want to solve?

PR #482, opened 2019-06-11
Fix the ssh issue from compute nodes to deployer during remove comput…

rajdharwadkar:
Issue: QE hit this issue in remove compute. when compute host is trying to ssh to deployer.
    Error: compute node trying to ssh to deployer
    "msg": "Failed to connect to the host via ssh:

### DIFF
--- a/playbooks/generic-clean_airship.yml
+++ b/playbooks/generic-clean_airship.yml
@@ -13,12 +13,17 @@
 # WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 # License for the specific language governing permissions and limitations
 # under the License.
-#
-- hosts: airship-openstack-compute-workers
+
+- hosts: soc-deployer
   gather_facts: no
   any_errors_fatal: true
-  roles:
-    - { role: openstack-remove-compute, compute_node_name: "{{ inventory_hostname }}" }
+  tasks:
+    - name: Remove compute nodes.
+      include_role:
+         name: openstack-remove-compute
+      loop: "{{ groups['airship-openstack-compute-workers'] }}"
+      loop_control:
+         loop_var: compute_node_name
 
 - hosts: soc-deployer
   gather_facts: no
--- a/playbooks/roles/openstack-remove-compute/tasks/main.yml
+++ b/playbooks/roles/openstack-remove-compute/tasks/main.yml
@@ -6,47 +6,42 @@
   environment:
     OS_CLOUD: openstack
   register: compute_id
-  delegate_to: "{{ groups['soc-deployer'][0] }}"
   failed_when: false
   changed_when: false
 
 - name: Validate compute node id.
   debug:
     msg: "Compute node {{ compute_node_name }} does not exists in openstack."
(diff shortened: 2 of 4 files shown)


playbooks/roles/openstack-remove-compute/tasks/main.yml
```diff
@@ -56,24 +55,22 @@
...
-  when: compute_id.stdout | length | int != 0
+  when: inventory_hostname == groups['soc-deployer'][0] and compute_id is defined and compute_id.stdout | length | int != 0
   failed_when: false
 
 - name: Delete subvolumes from compute node.
   import_role:
     name: airship-configure-worker
     tasks_from: subvolumes-setup.yml
-  delegate_to: '{{ compute_node_name }}'
+  when: inventory_hostname == compute_node_name
```
jgu17: just occurred to me that since we are counting on the compute host name in openstack and in ansible inventory to be identical, we should call it out in the deployment guide inventory file section. can be different doc commit though,

rajdharwadkar: Ok will update the doc.

playbooks/generic-clean_airship.yml
```diff
@@ -14,11 +14,11 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 #
-- hosts: airship-openstack-compute-workers
+- hosts: airship-openstack-compute-workers:soc-deployer
   gather_facts: no
   any_errors_fatal: true
   roles:
-    - { role: openstack-remove-compute, compute_node_name: "{{ inventory_hostname }}" }
+    - { role: openstack-remove-compute, compute_node_name: "{{ groups['airship-openstack-compute-workers'] }}" }
```
kantarun: How this will work ? 

groups is an array and it seems like compute_node_name is expected to be single host value

`               "airship-openstack-compute-workers": [
                    "linux-1pm5",
                    "linux-9nya",
                    "linux-xj84",
                    "linux-vrxc"
                ],
`

playbooks/roles/openstack-remove-compute/tasks/main.yml
kantarun: May be add this in same PR so that it can be reviewed in same place.

kantarun:
> …e node.
> 
> Issue: QE hit this issue in remove compute. when compute host is trying to ssh to deployer.
> {"changed": false, "msg": "Failed to connect to the host via ssh: ", "unreachable": true}

Not sure with above message if someone else can follow what's the issue and what's being done in this change.

playbooks/roles/openstack-remove-compute/tasks/main.yml
rajdharwadkar: done

rajdharwadkar:
> > …e node.
> > Issue: QE hit this issue in remove compute. when compute host is trying to ssh to deployer.
> > {"changed": false, "msg": "Failed to connect to the host via ssh: ", "unreachable": true}
> 
> Not sure with above message if someone else can follow what's the issue and what's being done in this change.

updated message

PR #484, opened 2019-06-12
Revert "Add 4 known smoke test failures to tempest_blacklist"

toabctl:
Reverts SUSE-Cloud/socok8s#479


Itxaka:
TP has been cut already so is the label proper?

Im ok with reverting this now that we have the TP "released" so we can investigate further on those tests

toabctl:
> TP has been cut already so is the label proper?
> 
> Im ok with reverting this now that we have the TP "released" so we can investigate further on those tests

@then add your approval :)

PR #485, opened 2019-06-12
Airship deployer: install python packages as rpm, not from pip [ SOC-9494 ]

jsuchome:
At this point deployer has repositories set up, no need for pip.


jsuchome:
The CI job ended only in the very last step, not even because of error. I assume the problem we see there is not related to the change

PR #487, opened 2019-06-12
Remove the requirement of SUSE CA repo

jsuchome:
There should be no need for usage of SUSE internal repositories.


jsuchome:
Hm, SES setup seems to require `ca-certificates-suse`, let's see if it is really needed

jsuchome:
Seems that the CI job was able to deploy SES...

PR #488, opened 2019-06-12
Use the workspace directory for storing ansible retry files

jsuchome:
Without this, packaged version of socok8s would try to create retry
files under /usr/share/socok8s which is only writable for root.

See SOC-9467

### DIFF
--- a/script_library/run-ansible.sh
+++ b/script_library/run-ansible.sh
@@ -31,7 +31,9 @@ function run_ansible(){
     fi
 
     pushd ${socok8s_absolute_dir}
-        ansible-playbook ${extra_vars:-} -i ${inventorydir} $@ -v
+        # ANSIBLE_RETRY_FILES_SAVE_PATH sets a location where ansible retry files should be located
+        # We are setting it explicitely to the socok8s_workspace directory which is user-writable
+        ANSIBLE_RETRY_FILES_SAVE_PATH=${socok8s_workspace} ansible-playbook ${extra_vars:-} -i ${inventorydir} $@ -v
     popd
     set +x
 }


script_library/run-ansible.sh
```diff
@@ -31,7 +31,7 @@ function run_ansible(){
     fi
 
     pushd ${socok8s_absolute_dir}
-        ansible-playbook ${extra_vars:-} -i ${inventorydir} $@ -v
+        ANSIBLE_RETRY_FILES_SAVE_PATH=${socok8s_workspace} ansible-playbook ${extra_vars:-} -i ${inventorydir} $@ -v
```
Itxaka: should we add a note for this? It could happen...maybe...somehow...that a deployer has already an ansible.cfg file in their deployer machine which contains the same config and we are overriding.

Its a really small chance of that but at least maybe we should comment this somehow somewhere in the docs?

jsuchome: I could add a comment here, and maybe we can add a comment to the docs, i agree.

jsuchome: @csymons-suse could you add a note to the documentation?

toabctl: or add `retry_files_save_path =` to the ansible.cfg file? Do we already ship a default in the workspace dir?

jsuchome: No, we don't. 
That would be the other option, yes, but we thought just using this var should be enough (see linked Jira entry)

If we would have default ansible cfg in the workspace dir, we would still need to export some var to point to this cfg file

Itxaka: not only that but once an ansible.cfg file is loaded, ansible doesnt look for any other files, so in the case of an existing ansible.cfg file already existing, we wont be able to load our own file.

toabctl: we could deliver an `${socok8s_workspace}/.ansible.cfg` file which can be overwritten when setting `ANSIBLE_CONFIG=$somepath` . See https://docs.ansible.com/ansible/2.4/intro_configuration.html#configuration-file

jsuchome: And what's the advantage compared to approach presented here?

(BTW I think we should check ANSIBLE_RETRY_FILES_SAVE_PATH exists and only if it does not, provide the default)

PR #489, opened 2019-06-12
Adding doc for providing master api node info for DNS

kantarun:
Adding instruction in case DNS is not used in CaaSP cluster and
is not availble on deployer for its usage. We need to explictly
set master node DNS resoltion (as used in kube .config file)

Ideally it would have been better to add this automatically via
ansible as we already know master node ip from inventory file. Its
too risky to make this change (Used to think its already there).

### DIFF
--- a/doc/source/deployment/configure.rst
+++ b/doc/source/deployment/configure.rst
@@ -156,7 +156,19 @@ Configure for Kubernetes
 SUSE Containerized OpenStack relies on kubectl and Helm commands to configure
 your OpenStack deployment. You need to provide a `kubeconfig` file on the
 `deployer` node, in your workspace. You can fetch this file from the Velum UI
-on your SUSE CaaS Platform cluster.
+on your SUSE CaaS Platform (CaaSP) cluster.
+
+In case, if DNS is not used for SUSE CaaSP cluster and is not available on
+deployer, add entries for DNS resolution in deployer's /etc/hosts file
+
+.. code-block:: yaml
+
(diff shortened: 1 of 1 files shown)


doc/source/deployment/configure.rst
```diff
@@ -156,7 +156,19 @@ Configure for Kubernetes
...
+In case, if DNS is not used for SUSE CaaSP cluster and is not available on
+deployer, add entries for DNS resolution in deployer's /etc/hosts file
+
+.. code-block:: yaml
+
+  # For example, with following master node, add respective entries
+  # caasp_master_node_ip: 192.168.7.231
+  # caasp_master_host_name: pcloud003master
+
+  192.168.7.231 api.infra.caasp.local
```
jgu17: I don't think we need api.infra.caasp.local. I don't use it in /etc/hosts on deployer.

kantarun: I checked with what Grace T was using and added here as per her usage. 

I recall seeing api.infra.caasp.local available in kube config when I used to do workaround for 'kubectl cluster-info' authorization issue (which comes when we don't use IP for master and admin/dashboard in velum UI).
As I see same her in config, I kept it here as don't see harm in having it here.

jgu17: maybe we are setting differently in the velum. I used only ip in the dashboard and api url. 

@gracetai5 could you check quickly if you have to have the api.infra.caasp.local  in the /etc/hosts for you to run kubectll cluster-info?

PR #490, opened 2019-06-12
doc: correct run.sh commands; add required command lines

csymons-suse:


doc/source/deployment/configure.rst
```diff
@@ -66,6 +66,9 @@ using. For example:
      hosts:
        caasp-worker-001:
          ansible_host: 10.86.1.144
+         airship-openstack-control-workers:
```
jesusaurus: This looks like it was added on accident, is it related to the change below?

csymons-suse: @jesusaurus my error, fix on the way

PR #496, opened 2019-06-12
[1.0] doc: DVR not supported

csymons-suse:
requested by Product Management & Project Management

### DIFF
--- a/doc/source/deployment/requirements.rst
+++ b/doc/source/deployment/requirements.rst
@@ -48,19 +48,15 @@ Minimum Node Specification
 * Memory: 16GB
 * Storage: 80GB
 
-  If the work node is used as Compute node, sizing shall be determined by
+  If the work node is used as Compute node, sizing should be determined by
   the target workloads on the compute node.
 
-:term:`SES` All-In-One (AIO) node (Experimental only)
-+++++++++++++++++++++++++++++++++++++++++++++++++++++
+:term:`SES` node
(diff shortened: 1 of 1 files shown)


doc/source/deployment/requirements.rst
```diff
@@ -105,5 +101,7 @@ Network Requirements
     the Deployer with dnsmasq to supply DNS functionality, but this is beyond
     the scope of this preview.
 
+Distributed Virtual Routing (DVR) is not supported in this Technology Preview.
```
jgu17: @csymons-suse T.T. asked for a section to state what openstack services are supported, and what features are not supported including dvr. so can we add a section in the overview?

jgu17: "Technical Preview"

jgu17: Also we need to state flat network is not supported. could be a different commit.

gmingsuse: I think the official SUSE term is Technology Preview.

jgu17: addressed in separate PR. so I am good here

jgu17: I added to the home/discaimer page. So all good here

dirkmueller:
unfortunately this was already merged via #518 so this one is no longer relevant (although it has a working commit message)

PR #497, opened 2019-06-12
doc: add airship-openstack-l3-agent-workers to host entries

csymons-suse:


jgu17:
CI is failing, but this is a simple doc change and can't cause CI failure.

PR #501, opened 2019-06-13
[1.0] doc: add content describing OpenStack services. Minor proofing.

csymons-suse:


doc/source/user/select-openstack-services.rst
```diff
@@ -1,6 +1,4 @@
-====================================
-Select OpenStack services to install
-====================================
+=========================
```
jgu17: missing the section title?

csymons-suse: blame the tools

PR #502, opened 2019-06-13
CI: build docs when they change

Itxaka:
We currently do not build the docs so we are not sure if they are
properly done.

This enables the build of them if the dir has changed or the tox.ini
file has changed

Also, we take the opportunity to archive the docs if they were built
so they are accessible to download

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -80,6 +80,33 @@ pipeline {
             }
         }
 
+        stage('Build docs') {
+             when {
+                anyOf {
+                    expression { return DocsChanged }
+                    expression { return ToxChanged }
+                }
+             }
+             options {
+                timeout(time: 5, unit: 'MINUTES', activity: true)
+             }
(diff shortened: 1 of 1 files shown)


Itxaka:
@evrardjp I had a previous change (which is now force pushed) with some of the docs changed to see if it was built and published. You can see the jenkins run and the docs here: https://ci.suse.de/job/cloud-socok8s-jobs/job/socok8s-integration/job/PR-502/5/

Itxaka:
> I am fine with current approach, but I still think we should have a different way to trigger things. That can be done in a later patch.

I do agree that having a jenkinsfilechange would be awesome and it should trigger all the jobs, lets merge this and I will send a new patch for that!

Or are you thinking of a different approach?

> LGTM. What is the retention policy of the built doc?

Same as the job, so once the job is deleted, the docs are deleted. I think its 30 days and/or 30 runs per job?

PR #505, opened 2019-06-13
[1.0] jenkins: Sanitize the socok8s enviroment name

rhafer:
SOCOK8S_ENVNAME is used to generate the names for some of the heat
stacks we create. So we need to make sure it doesn't contain any
problematic characters for that. (This should fix the currenty build
failures for the stable/1.0 branch)

(cherry picked from commit 461775c236f07f96149905fb33d9495d90ced1e1)

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -28,8 +28,10 @@ pipeline {
     }
 
     environment {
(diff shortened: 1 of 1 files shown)


toabctl:
Merging. Otherwise we won't have CI for the stable branch....

PR #509, opened 2019-06-13
doc: add content about retry file path (SOC-9467)

csymons-suse:


doc/source/deployment/configure.rst
```diff
@@ -190,6 +190,16 @@ For example:
 
    socok8s_ext_vip: "10.10.10.10"
 
+Set Up Retry Files Save Path
+----------------------------
+
+Before beginning deployment, a path must be specified where Ansible retry files
```
jsuchome: It's actually not a _must_. Those retry files are not necessary and if ansible fails to create them (after a failure during execution), it's just one more error, I would not say a critical one.

doc/source/deployment/configure.rst
```diff
@@ -190,6 +190,16 @@ For example:
...
+can be saved in order to avoid potential errors. The path must point to a
+user-writable directory. Set the path in either of the following ways:
+
+- ``export ANSIBLE_RETRY_FILES_SAVE_PATH=<PATH_TO_FILE>`` before deploying with
```
jsuchome: It's actually path to directory, not to the file

doc/source/deployment/configure.rst
```diff
@@ -190,6 +190,16 @@ For example:
...
+  ``run.sh`` commands.
+- Set the value of `retry_files_save_path` in your Ansible configuration file.
```
jsuchome: There's actually also an option to disable creating those retry files. Use `retry_files_enabled = False` setting in ansible conf file for that

jsuchome:
Hm, actually we should have this in 1.0 only and not in master. In master branch we set  ANSIBLE_RETRY_FILES_SAVE_PATH to point to workdir...

toabctl:
@jsuchome can you create a revert then?

PR #513, opened 2019-06-13
Support multiple PR target branches

jesusaurus:
Switch from a hardcoded 'master' branch to the TARGET_BRANCH environment variable in order to support PRs against stable branches as well as the master branch.

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -48,17 +48,17 @@ pipeline {
             }
         }
         stage('Check for updated files') {
-            when { expression { env.BRANCH_NAME != 'master' } }
+            when { expression { env.TARGET_BRANCH != null } } /* This is only set when testing a PR */
             steps {
                 script {
                     /* When doing PRs, make sure we don't test everything by default */
                     TestDocs = false
                     TestFunctional = false
 
-                    /* Need to fetch master to check against it for the proper diff */
-                    sh "git config --add remote.origin.fetch +refs/heads/master:refs/remotes/origin/master"
+                    /* Need to fetch the target branch to check against it for the proper diff */
+                    sh "git config --add remote.origin.fetch +refs/heads/${env.TARGET_BRANCH}:refs/remotes/origin/${env.TARGET_BRANCH}"
                     sh "git fetch --no-tags"
-                    List<String> sourceChanged = sh(returnStdout: true, script: "git diff --name-only origin/master...origin/${env.BRANCH_NAME}").split()
+                    List<String> sourceChanged = sh(returnStdout: true, script: "git diff --name-only origin/${env.TARGET_BRANCH}...origin/${env.BRANCH_NAME}").split()
                     echo "Files changed for this PR:\n${sourceChanged.join('\n')}"
                     TestDocs = sourceChanged.any{it.contains("doc/")} or sourceChanged.any{it.contains("tox.ini")} or sourceChanged.any{it.contains("Jenkinsfile.integration")}
                     /* In the future we could add a conditional to auto TestFunctional if tox.ini or Jenkinsfile is changed */


Jenkinsfile.integration
```diff
@@ -87,6 +87,7 @@ pipeline {
                 anyOf {
                     expression { return DocsChanged }
                     expression { return ToxChanged }
+                    expression { env.BRANCH_NAME == 'master' }
```
jsuchome: I admit I do not understand the syntax, but this looks like an OR to me. Should not there be rather "check for master AND doc checks" ?

jhesketh:
This isn't a periodic job, so I'm confused by this and would like to see some more info in the commit please.

Jenkinsfile.integration
Itxaka: no, this is correct.

So for the master branch, the DocsChange/ToxChanged will always be false, because you cannot compare between master and master and get what has changed :)

This is just for the periodic job which is run off master.

Itxaka:
> This isn't a periodic job, so I'm confused by this and would like to see some more info in the commit please.

Dont we use the same jenkinsfile for the periodic job? This should trigger on those rigth?


In any case, yes, a better explanation on the commit would be better I guess.

toabctl:
> Dont we use the same jenkinsfile for the periodic job? This should trigger on those rigth?

Yes. That's also how I understand it. I think this change is correct (but the commit msg is not)

jhesketh:
> > Dont we use the same jenkinsfile for the periodic job? This should trigger on those rigth?
> 
> Yes. That's also how I understand it. I think this change is correct (but the commit msg is not)

Unless I'm missing something, the  socok8s-integration job only runs against changes in github. The RPM job (yet to merge https://github.com/SUSE-Cloud/automation/pull/3478) has its own jenkinsfile for a number of reasons.

Itxaka:
> Unless I'm missing something, the socok8s-integration job only runs against changes in github. The RPM job (yet to merge SUSE-Cloud/automation#3478) has its own jenkinsfile for a number of reasons.

Yeah,  I think that the PR job also buids a periodic job from master as well as shown on: https://ci.suse.de/blue/organizations/jenkins/cloud-socok8s-jobs%2Fsocok8s-integration/detail/master/150/pipeline althought Im not sure when or how does it do (after a merge into master Im guessing)

So its not really periodic, but a branch job

jhesketh:
> > Unless I'm missing something, the socok8s-integration job only runs against changes in github. The RPM job (yet to merge [SUSE-Cloud/automation#3478](https://github.com/SUSE-Cloud/automation/pull/3478)) has its own jenkinsfile for a number of reasons.
> 
> Yeah, I think that the PR job also buids a periodic job from master as well as shown on: https://ci.suse.de/blue/organizations/jenkins/cloud-socok8s-jobs%2Fsocok8s-integration/detail/master/150/pipeline althought Im not sure when or how does it do (after a merge into master Im guessing)
> 
> So its not really periodic, but a branch job

Right, yes, I believe it's only on merge/push to a branch.

Jenkinsfile.integration
jhesketh: Right, so this will build Docs on any push to master or stable/1.0 (regardless of what's changed)?
We should add a comment as to why this is.

jesusaurus:
> > > Unless I'm missing something, the socok8s-integration job only runs against changes in github. The RPM job (yet to merge [SUSE-Cloud/automation#3478](https://github.com/SUSE-Cloud/automation/pull/3478)) has its own jenkinsfile for a number of reasons.
> > 
> > 
> > Yeah, I think that the PR job also buids a periodic job from master as well as shown on: https://ci.suse.de/blue/organizations/jenkins/cloud-socok8s-jobs%2Fsocok8s-integration/detail/master/150/pipeline althought Im not sure when or how does it do (after a merge into master Im guessing)
> > So its not really periodic, but a branch job
> 
> Right, yes, I believe it's only on merge/push to a branch.

OK, so I grossly misunderstood when we are running this job on the master and stable branches. I thought we were running the job to know that there are no environmental issues causing failures (such as changes in deep dependencies).

What is the reasoning for running the integration test post-merge? Are there cases where that could fail despite the pr-merge job succeeding?

jhesketh:
> OK, so I grossly misunderstood when we are running this job on the master and stable branches. I thought we were running the job to know that there are no environmental issues causing failures (such as changes in deep dependencies).

There is a periodic job being created for that still. Although it more specifically tests the RPM https://github.com/SUSE-Cloud/automation/pull/3478 (but that of course is to find those deep dependency failures).

> What is the reasoning for running the integration test post-merge? Are there cases where that could fail despite the pr-merge job succeeding?

I believe it's just how the multibranch-pipeline works in Jenkins. But more importantly it is useful as the original run against the PR may not represent the state that the code is in once it merges into master (since we aren't doing predictive gating like zuul).

toabctl:
> I believe it's just how the multibranch-pipeline works in Jenkins. But more importantly it is useful as the original run against the PR may not represent the state that the code is in once it merges into master (since we aren't doing predictive gating like zuul).

We could test changes step-by-step but with the multibranch plugin that would me, whenever something got merged into master, all the open PRs would be retested against current master. And that would create a huge CI load whenever something gets merged

toabctl:
@jesusaurus needs a rebase

PR #514, opened 2019-06-13
doc: Add documentation for updating OpenStack service configuration

matthew-fuller:
This change updates the operations guide to include a section covering
updates to OpenStack service configuration.

### DIFF
--- a/doc/source/operations.rst
+++ b/doc/source/operations.rst
@@ -192,6 +192,46 @@ to
 
    heat_api: "{{ suse_osh_registry_location }}/openstackhelm/heat:stein"
 
+Updating OpenStack Service Configuration
+----------------------------------------
+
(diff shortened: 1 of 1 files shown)


doc/source/operations.rst
```diff
@@ -192,6 +192,39 @@ to
...
+   conf:
+     logging:
+       logger_root:
+         level: DEBUG
+       logger_keystone:
+         level: DEBUG
+
+.. note::
+
+   Please consult the documentation for the appropriate OpenStack service for
```
jgu17: could we add an example what to consult, in this case, the openstack-helm/keystone/values.yaml.

csymons-suse: maybe add a link to https://docs.openstack.org/rocky/configuration/index.html

PR #515, opened 2019-06-13
doc: change install methods, remove install from repository

csymons-suse:
indicate git clone intended for developers


csymons-suse:
> It now has conflict with the master version from another PR. Would you like to rebase/merge first?
done. Looks good now and content is correct. thanks for working on the clone repository option.

PR #519, opened 2019-06-13
1.0 Corrected requirements

jgu17:
Corrected requirements


jgu17:
> Do we need to specify some env variable for clone option (like SOCOK8S_DEVELOPER_MODE , SOOK8S_USE_VIRTUALENV)

good question. dev mode is covered in advanced user options.  I am always confused about the virtual env. I think the favored opinion at the moment is that it is not needed, and if it's not working, it's a bug,

PR #529, opened 2019-06-14
Add instruction to select openstack service to install

jgu17:
Added instructions for advanced users to configure which openstack
services to install.

### DIFF
--- a/doc/source/user/select-openstack-services.rst
+++ b/doc/source/user/select-openstack-services.rst
@@ -1,10 +1,26 @@
-================================
-Select OpenStack services to run
-================================
+====================================
+Select OpenStack services to install
+====================================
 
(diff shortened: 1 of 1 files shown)


doc/source/user/select-openstack-services.rst
```diff
@@ -1,10 +1,26 @@
...
+     - openstack-ingress-controller-soc
+     - openstack-mariadb-soc
+     - openstack-memcached-soc
+     - openstack-keystone-soc
+     - openstack-ceph-config-soc
+     - openstack-glance-soc
+     - openstack-cinder-soc
+     - openstack-compute-kit-soc
+     - openstack-heat-soc
+     - openstack-horizon-soc
```
jsuchome: Hm, the file actually also mentions `openstack-tempest-soc` and deploys it conditionally

PR #534, opened 2019-06-14
Turn the developer mode off for primary CI job (SOC-9363)

jsuchome:
Developer mode is used for building the images which we are not
testing in CI anyway.
And we want the job to be close to end-user scenario.


Itxaka:
acording to this: https://jira.suse.com/browse/SOC-9470 it should fail....

PR #535, opened 2019-06-14
Jenkins: Skip full socok8s run on doc only changes

rhafer:
Rerunning the full deployment for PRs that only touch the documenation
is not needed. Skipping that will save us quite some time.

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -4,6 +4,7 @@
 DocsChanged = false
 PlaybooksChanged = false
 ToxChanged = false
+FullDeploymentNeeded = true
 
 
 pipeline {
@@ -56,12 +57,17 @@ pipeline {
                     sh "git fetch --no-tags"
                     List<String> sourceChanged = sh(returnStdout: true, script: "git diff --name-only origin/master..origin/${env.BRANCH_NAME}").split()
                     echo "Files changed for this PR:\n${sourceChanged.join('\n')}"
+                    FullDeploymentNeeded = false
                     for (int i = 0; i < sourceChanged.size(); i++) {
                         if (sourceChanged[i].contains("playbooks/")) {
                             PlaybooksChanged = true
                         }
                         if (sourceChanged[i].contains("doc/")) {
                             DocsChanged = true
+                        } else {
+                            /* if anything outside the doc dir change we need to
+                               run the socok8s deployment */
+                            FullDeploymentNeeded = true
                         }
                         if (sourceChanged[i].contains("tox.ini")) {
                             ToxChanged = true
@@ -110,6 +116,8 @@ pipeline {
         }
 
         stage('Create network') {
+            when { expression { return FullDeploymentNeeded } }
+
             options {
                 timeout(time: 10, unit: 'MINUTES', activity: true)
             }
@@ -118,6 +126,8 @@ pipeline {
             }
         }
         stage('Create VMs') {
+            when { expression { return FullDeploymentNeeded } }
+
             options {
                 timeout(time: 45, unit: 'MINUTES', activity: true)
             }
@@ -141,6 +151,8 @@ pipeline {
         }
 
         stage('Configure CaaSP workers') {
+            when { expression { return FullDeploymentNeeded } }
+
             options {
                 timeout(time: 10, unit: 'MINUTES', activity: true)
             }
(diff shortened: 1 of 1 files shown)


Jenkinsfile.integration
```diff
@@ -165,38 +181,48 @@ pipeline {
...
-        failure {
-            script {
-                if (env.hold_instance_for_debug == 'true') {
-                    echo "You can reach this node by connecting to its floating IP as root user, with the default password of your image."
-                    timeout(time: 3, unit: 'HOURS') {
-                        input(message: "Waiting for input before deleting  env ${SOCOK8S_ENVNAME}.")
+    if (FullDeploymentNeeded) {
+        post {
+            failure {
+                when { expression { return FullDeploymentNeeded } }
```
Itxaka: is this superflous as we are already inside the if?

Jenkinsfile.integration
```diff
@@ -165,38 +181,48 @@ pipeline {
...
+                script {
+                    sh './run.sh gather_logs'
+                }
+                zip archive: true, dir: 'logs/', zipFile: 'logs.zip'
+                archiveArtifacts artifacts: 'logs.zip'
             }
-            script {
-                sh './run.sh gather_logs'
+            cleanup {
+                when { expression { return FullDeploymentNeeded } }
```
Itxaka: same here

Jenkinsfile.integration
```diff
@@ -56,12 +57,17 @@ pipeline {
                     sh "git fetch --no-tags"
                     List<String> sourceChanged = sh(returnStdout: true, script: "git diff --name-only origin/master..origin/${env.BRANCH_NAME}").split()
                     echo "Files changed for this PR:\n${sourceChanged.join('\n')}"
+                    FullDeploymentNeeded = false
```
Itxaka: this will make it not run when playbooks+docs have changed or tox+docs have changed

Jenkinsfile.integration
rhafer: Argh. The "if" was supposed to go away. As it does not work. (See the failing tests)

Jenkinsfile.integration
rhafer: Why not. It will be set to `true` on line 70 in that case.

Itxaka: but if the docs AND the playbooks have changed we want a full CI job no?

Itxaka: > Why not. It will be set to `true` on line 70 in that case.

it wont. If the docs and playbooks have changed, it wont enter the `else` and it will skip a full deployment.

We want `FullDeploymentNeeded` when only docs have changed, any other combination should result in running the full job.

rhafer: > it wont. If the docs and playbooks have changed, it wont enter the else and it will skip a full deployment.

It's inside the loop  interating of the list of changed files. So a if there is a single file that is not inside the `doc/` directory, it will get to that else branch. Even if it also matches one of the other conditions.

> We want FullDeploymentNeeded when only docs have changed, any other combination should result in running the full job.

We want `FullDeplyomentNeeded = false` when all changes are limited to the doc dir. We want `FullDeploymentNeeded = true`, when either we're builidng from the master branch (that's why the value defaults to `true` on line#7 and is set to `false` before the loop on line#60) or anything outside the `doc/` dir was changed (that's what the `else` on line#70 should catch).

If you have any better suggestion how to express that in groovy I am all ears :).

Itxaka: You are correct!

Jenkinsfile.integration
```diff
@@ -56,12 +57,17 @@ pipeline {
...
                     for (int i = 0; i < sourceChanged.size(); i++) {
                         if (sourceChanged[i].contains("playbooks/")) {
                             PlaybooksChanged = true
                         }
                         if (sourceChanged[i].contains("doc/")) {
                             DocsChanged = true
+                        } else {
```
evrardjp: is the else only triggering when doesn't contain docs? (#me, noob at jenkins...).
If that's the case, what about a change that contain both docs and code? Wouldn't that be a problem?

rhafer: `sourceChanged` is a list of strings. It contains all name of files that have been touched by the current PR (generated by the git diff command on line 58).

As we are inside a for loop (line 61) iterating through that list. The else branch will trigger as soon as there is a single filename in that list that does not contain `doc/`. (So if there is a change that touches doc and code it would trigger)

evrardjp: Basic logic doesn't apply on Friday afternoon, everyone knows that.

I suppose the reno would be considered as part of docs so I don't see a reason to not do it this way.

PR #536, opened 2019-06-14
CI: track changes to jenkinsfile

Itxaka:
Track changes to jenkinsfile file in case we want to run the
full pipeline when something changes tot est that all steps are
correct

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -4,6 +4,7 @@
 DocsChanged = false
 PlaybooksChanged = false
 ToxChanged = false
+JenkinsfileChanged = false
 
 
 pipeline {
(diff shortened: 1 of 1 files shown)


Itxaka:
will wait until https://github.com/SUSE-Cloud/socok8s/pull/535 is merged to rework this as we might be able to reuse parts of those to set the full run

evrardjp:
> will wait until #535 is merged to rework this as we might be able to reuse parts of those to set the full run

do we? These are two different things IMO.

Jenkinsfile.integration
```diff
@@ -66,13 +67,21 @@ pipeline {
                         if (sourceChanged[i].contains("tox.ini")) {
                             ToxChanged = true
                         }
+                        if (sourceChanged[i].contains("Jenkinsfile.integration")) {
+                            JenkinsfileChanged = true
```
evrardjp: A cleaner way would be that if Jenkinsfile.integration has changed, then set ToxChanged/DocsChanged/PlaybooksChanged to true. It would not need the extra conditionals.

Itxaka:
> do we? These are two different things IMO.

If we have a FullDeploymentNeeded var that sets all steps to run, then we can probably leverage that to set it if the jenkinsfile is changed, so we dont have to add much

evrardjp:
See also:  https://github.com/SUSE-Cloud/socok8s/pull/559

jhesketh:
This needs a rebase now 559 is merged.

evrardjp:
I think it can even be abandonned.

Itxaka:
abandoned in favour of #559 !

PR #537, opened 2019-06-14
Stop using which

dirkmueller:
We can easily avoid this dependency, so use a bash builtin
instead of the (deprecated) which command.


jsuchome:
TODO: drop which from package Requires

PR #539, opened 2019-06-14
Prevent adding socok8s repository when the content is already there

jsuchome:
zypper_repository is not enough, as it only checks for same name or URL.
However users might add the iso image manually. Image has same content
as repository and installs the same product, but different name/URI.

### DIFF
--- a/playbooks/roles/airship-setup-deployer/tasks/main.yml
+++ b/playbooks/roles/airship-setup-deployer/tasks/main.yml
@@ -5,6 +5,16 @@
 - name: Load vip from user variables
   include_vars: "{{ socok8s_extravars }}"
 
+# In some cases, people might add iso manually.
+# Let's prevent adding the same product twice.
+- name: Check if socok8s product already present
+  shell: |
+    set -o pipefail
+    zypper products | grep -q openSUSE-Addon-socok8s
+  register: product_present
+  changed_when: "product_present.rc == 1"
+  failed_when: "product_present.rc not in [0,1]"
+
 - name: Add socok8s repository
   become: yes
   zypper_repository:
@@ -13,6 +23,7 @@
     autorefresh: True
     auto_import_keys: yes
     state: present
+  when: product_present is changed
 
 - name: Install system packages
   become: yes


playbooks/roles/airship-setup-deployer/tasks/main.yml
```diff
@@ -5,6 +5,12 @@
 - name: Load vip from user variables
   include_vars: "{{ socok8s_extravars }}"
 
+# In some cases, people might add iso manually.
+# Let's prevent adding the same product twice.
+- name: Check if socok8s product already present
+  command: zypper products | grep -q openSUSE-Addon-socok8s
+  register: product_present
+
```
toabctl: `changed_when: False` needed

playbooks/roles/airship-setup-deployer/tasks/main.yml
```diff
@@ -13,6 +20,7 @@
     autorefresh: True
     auto_import_keys: yes
     state: present
+  when: not product_present
```
Itxaka: cant we use the command here directly?

jsuchome: I don't know, can we? I still feel new to ansible...

Itxaka: @evrardjp might be the person to ask this :D

playbooks/roles/airship-setup-deployer/tasks/main.yml
```diff
@@ -5,6 +5,15 @@
 - name: Load vip from user variables
   include_vars: "{{ socok8s_extravars }}"
 
+# In some cases, people might add iso manually.
+# Let's prevent adding the same product twice.
+- name: Check if socok8s product already present
+  shell: |
+    set -o pipefail
+    zypper products | grep -q openSUSE-Addon-socok8s
```
jhesketh: Should we base it off the URL or something else in case it was added manually under a different name?

jsuchome: I'm explicitly checking for a product name which is defined on our side and cannot be changed by end user.

playbooks/roles/airship-setup-deployer/tasks/main.yml
jsuchome: I think one can't execute code in `when`

playbooks/roles/airship-setup-deployer/tasks/main.yml
evrardjp: That is a good check then.

I would change that task though:
```
shell: zypper products | grep -q openSUSE-Addon-socok8s
register: product_present
changed_when: "product_present.rc == 1"
failed_when: "product_present.rc not in [0,1]"
```

This would allow to know if things need changing in an ansible way (if this task is changed, trigger another task).

In that case, the addition of the other task's conditional is:
```
when: product_present is changed
```

jsuchome:
The repository code was reworked recently with https://github.com/SUSE-Cloud/socok8s/pull/567

I'm not yet sure how to react. This PR used to handle specific repository only, while the merged PR makes the code more generic. That sounds reasonable in general, but does not really help our case here...

dirkmueller:
rebase anyone?

jsuchome:
It's not simple rebase, I'd need to look if my idea could still be applied to current code.

PR #548, opened 2019-06-14
Update to freeze osh images

jgu17:
Updated to use frozen osh and osh-infra images with date tag on 2019/06/13.
Changed RabbitMQ to also use opensuse image for db init.
Fixed libvirt chart overrides for image tag to use the freezing osh infra image tag.


Itxaka:
Do we need this on mater actually? or we want to keep using the latest unfreeze (unfrozen?) images?

jgu17:
> Do we need this on master actually? or we want to keep using the latest unfreeze (unfrozen?) images?

excellent question. would make sense to use latest now that we have 1.0 branch. Live dangerously :-)

PR #556, opened 2019-06-16
[1.0] Stable minor fixups

csymons-suse:
remove duplicate content in deployment/requirements.rst


jgu17:
It is just doc change. overriding the wait for CI to put doc change on a fast trck for the doc release today

PR #558, opened 2019-06-18
Switch CI job to provo mirror for Leap

rhafer:
This commit changes the CI job to use the local openSUSE Leap mirrors in
the datacenter, which should give a bit of a speed up for the zypper dup
call. (And also help avoid issues when the external mirror
infrastructure is slow)

Additionally we disable the use of delta RPMs in that case. Which
doesn't give any benefit when using the local mirror.

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -32,6 +32,7 @@ pipeline {
            as the names of the heat stacks will be derived from this. Also
            the CaaSP Velum automation has issues with mixed case hostnames. */
         SOCOK8S_ENVNAME = "cloud-socok8s-${env.BRANCH_NAME.replaceAll("[^a-zA-Z0-9-]+", "-").toLowerCase()}-${env.BUILD_NUMBER}"
+        SOCOK8S_OPENSUSE_MIRROR="https://provo-mirror.opensuse.org"
         OS_CLOUD = "engcloud-cloud-ci"
         KEYNAME = "engcloud-cloud-ci"
         DELETE_ANYWAY = "YES"
--- a/playbooks/openstack-osh_hostconfig.yml
+++ b/playbooks/openstack-osh_hostconfig.yml
@@ -8,6 +8,25 @@
     - name: Ensure all the openstack mechanism vars are loaded
       include_vars: "{{ playbook_dir }}/../vars/deploy-on-openstack.yml"
 
+    - name: Collect repo files for removal
+      find:
+        paths: "/etc/zypp/repos.d/"
+        patterns: "*.repo"
+      register: repo_files
+
+    - name: Clear existing repo config
+      file:
+        path: "{{ item['path'] }}"
+        state: absent
+      with_items: "{{ repo_files['files'] }}"
+
+    - name: Disable delta rpms
+      ini_file:
+        path: /etc/zypp/zypp.conf
+        section: main
+        option: download.use_deltarpm
+        value: "false"
+
     - name: Configure host repositories
       zypper_repository:
         name: "{{ item }}"
--- a/vars/common-vars.yml
+++ b/vars/common-vars.yml
@@ -18,15 +18,21 @@ suse_osh_registry_location: "docker.io"
 suse_openstack_image_version: "rocky-opensuse_15"
 suse_infra_image_version: "latest-opensuse_15"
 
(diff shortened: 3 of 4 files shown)


Jenkinsfile.integration
```diff
@@ -32,6 +32,7 @@ pipeline {
            as the names of the heat stacks will be derived from this. Also
            the CaaSP Velum automation has issues with mixed case hostnames. */
         SOCOK8S_ENVNAME = "cloud-socok8s-${env.BRANCH_NAME.replaceAll("[^a-zA-Z0-9-]+", "-").toLowerCase()}-${env.BUILD_NUMBER}"
+        SOCOK8S_USE_PRV_MIRROR="true"
```
toabctl: Could we allow a custom mirror url instead of this bool var?

vars/common-vars.yml
```diff
@@ -14,19 +14,30 @@ socok8s_registry_cert: "{{ socok8s_workspace }}/certs/domain.crt"
 socok8s_registry_certkey: "{{ socok8s_workspace }}/certs/domain.key"
 socok8s_caasp_environment_details: "{{ socok8s_workspace }}/environment.json"
 socok8s_libvirtuuid: "{{ socok8s_workspace }}/libvirt.uuid"
+socok8s_use_prv_mirror: "{{ lookup('env','SOCOK8S_USE_PRV_MIRROR') | default('False', true) | bool }}"
 suse_osh_registry_location: "docker.io"
 suse_openstack_image_version: "rocky-opensuse_15"
 suse_infra_image_version: "latest-opensuse_15"
 
+obs_mirror: "{%- if socok8s_use_prv_mirror -%}
+               https://provo-mirror.opensuse.org
```
jhesketh: Will the whitespace matter here?

Jenkinsfile.integration
rhafer: Yeah, let's try that ...

vars/common-vars.yml
rhafer: I reworked that part. It's gone now.

rhafer:
When using the local mirror, "Deploy CCP Deployer" is now down to ~8.5 minutes (from 17min) in the CI. So it's no longer the slowest task in that "Create VM substep".

playbooks/openstack-osh_hostconfig.yml
```diff
@@ -8,6 +8,25 @@
     - name: Ensure all the openstack mechanism vars are loaded
       include_vars: "{{ playbook_dir }}/../vars/deploy-on-openstack.yml"
 
+    - name: Collect repo files for removal
+      find:
+        paths: "/etc/zypp/repos.d/"
+        patterns: "*.repo"
+      register: repo_files
+
+    - name: Clear existing repo config
```
Itxaka: Could we disable them instead? Removing them may bring issues with existing repos on the deployer.

Or maybe move them somewhere?

rhafer: There are no other existing repos on the deployer than what's part of the image. And those are the ones we want to replace. What other repos are there that we'd want to keep?

Itxaka: Im just thinking of someone bringing its own deployer, instead of letting us create it, but that is not supported, nor this is logical as there may as well be other changes to do in that case, so ignore me!

playbooks/openstack-osh_hostconfig.yml
```diff
@@ -8,6 +8,25 @@
...
+      file:
+        path: "{{ item['path'] }}"
+        state: absent
+      with_items: "{{ repo_files['files'] }}"
```
toabctl: Always interesting to see how many lines of ansible you need to do simple tasks. wow

vars/common-vars.yml
```diff
@@ -18,15 +18,21 @@ suse_osh_registry_location: "docker.io"
 suse_openstack_image_version: "rocky-opensuse_15"
 suse_infra_image_version: "latest-opensuse_15"
 
+obs_mirror: "{{ lookup('env','SOCOK8S_OPENSUSE_MIRROR') | default('https://download.opensuse.org', true) }}"
```
toabctl: nit: should that be opensuse_mirror instead of obs_mirror ?

PR #559, opened 2019-06-18
Ensure everything is properly tested


evrardjp:
With current code, there is no change triggered if site/ is changed,
Jenkinsfile, or vars. They should all be tested.

We should by default test everything, and skip what's not
to test, when doing a PR.

This fixes the issue by:
* skipping functional builds if the PR contains only docs changes
* build docs if only docs, tox, or Jenkinsfile is changed
  (We can't assume tox.ini won't change lint, nor that Jenkins file
  has no impact on testing, so a change on those should trigger
  everything).

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -1,9 +1,8 @@
 #!groovy
 
 // Global vars so they can be used in different steps
-DocsChanged = false
-PlaybooksChanged = false
-ToxChanged = false
+TestDocs = true
+TestFunctional = true
 
 
 pipeline {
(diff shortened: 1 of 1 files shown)


Jenkinsfile.integration
```diff
@@ -50,29 +49,25 @@ pipeline {
...
-                            PlaybooksChanged = true
-                        }
-                        if (sourceChanged[i].contains("doc/")) {
-                            DocsChanged = true
-                        }
-                        if (sourceChanged[i].contains("tox.ini")) {
-                            ToxChanged = true
-                        }
-                    }
+                    TestDocs = sourceChanged.any{it.contains("doc/")} or sourceChanged.any{it.contains("tox.ini")} or sourceChange.any{it.contains("Jenkinsfile.integration")}
```
Itxaka: missing a d!

```suggestion
                    TestDocs = sourceChanged.any{it.contains("doc/")} or sourceChanged.any{it.contains("tox.ini")} or sourceChanged.any{it.contains("Jenkinsfile.integration")}
```

evrardjp: Good catch!

evrardjp:
It has properly run all the jobs. Woot!

PR #562, opened 2019-06-19
CI: Publish ARA reports

Itxaka:
Same as the docs, publish ARA reports so its easier
to troubleshoot what went wrong and get timing info
on playbooks

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -39,6 +39,9 @@ pipeline {
         ANSIBLE_VERBOSITY = 2
         ANSIBLE_STDOUT_CALLBACK = "yaml"
         USER = "jenkins" /* Why isn't this set in the jenkins environment? */
+        USE_ARA = "True"
+        SOCOK8S_USE_VIRTUALENV = "True"
+        ARA_DIR = "$WORKSPACE"
     }
 
     stages {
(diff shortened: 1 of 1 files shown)


Jenkinsfile.integration
```diff
@@ -176,6 +177,19 @@ pipeline {
     }
 
     post {
+        always {
+            script {
+                sh 'ara generate html ara_report'
```
evrardjp: I guess we need to make sure this command runs inside the appropriate venv?

Itxaka: yes, this is something that I wanted to test with the jenkins run, as it may need tweaking (and also it may need ara install? Ara is probably not installed in the nodes and we are not using venv currently IIRC)

Itxaka:
as expected this fails because ara is not installed.... 

I guess we should install it manually in the node

toabctl:
> I guess we should install it manually in the node

let me try to create a package for it..

Itxaka:
ara >=1.x does not support generating static content anymore. For this to work we should create a venv at the start to install an older ara version and use that or install ara with `--user` and use the full path to it.

Both options require too many changes for what ara gives us, so not moving forward with this.

Itxaka:
ara 0.16.3 is installed and available under jenkins so this should work...maybe?

Itxaka:
so this works: https://ci.suse.de/job/cloud-socok8s-jobs/job/socok8s-integration/job/PR-562/6/ARA_20Report/

PR #564, opened 2019-06-19
CI: Drop travis file

Itxaka:
There is no point anymore on running travis as we are running
everything from jenkins. Lets just drop travis and make the
jenkins check mandatory


toabctl:
So linting now also runs from Jenkins? I still like travis because it reports faster than our jenkins (which is currently limited to 7 slots and does not report that a job is in the queue)

evrardjp:
Agreed on the intent. Let's define everything in Jenkinsfile, and if really necessary, run things in parallel.

Itxaka:
This is here for the future, until we force the jenkins job to be required this cannot be merged, or we split the jenkins job in two, althougth rigth now it seems like a really nice way of doing it, linting first which stops the rest of the job if it fails.

> Agreed on the intent. Let's define everything in Jenkinsfile, and if really necessary, run things in parallel.

Im not sure if I agree on running in parallel, I really like that we do not consume resources if the linting is wrong, having a job for linting in parallel with the full run seems a waste of resources.
Or maybe Im not understanding the comment correctly, Im a bit reading-handicapped today.

Itxaka:
CI is now enforced so we should drop travis altogether as the linting is covered under jenkins *and* will make the job fail fast if it fails

PR #565, opened 2019-06-19
Do not deploy glance with the values intended for Newton

jsuchome:
glance deployment script (100-glance.sh) sets OSH_OPENSTACK_RELEASE:="newton"

### DIFF
--- a/playbooks/roles/deploy-osh/tasks/main.yml
+++ b/playbooks/roles/deploy-osh/tasks/main.yml
@@ -295,6 +295,7 @@
   vars:
     _component_runcommand: "./tools/deployment/multinode/100-glance.sh"
     _component_run_env:
+      OSH_OPENSTACK_RELEASE: "{{ suse_osh_openstack_release }}"
       OSH_EXTRA_HELM_ARGS_GLANCE: >
         --values=./glance/values_overrides/rocky-opensuse_15.yaml
         --values=/tmp/socok8s-susedefaults-glance.yaml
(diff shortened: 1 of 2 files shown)


playbooks/roles/deploy-osh/tasks/main.yml
```diff
@@ -295,6 +295,7 @@
   vars:
     _component_runcommand: "./tools/deployment/multinode/100-glance.sh"
     _component_run_env:
+      OSH_OPENSTACK_RELEASE: "rocky"
```
Itxaka: should we use a generic var on commo-vars.yml to track this in case we need it somewhere else? I dont like to hardcode things.

jsuchome: Good point

toabctl:
@jsuchome I guess we need to backport this, right? Or don't we care?

Itxaka:
@jsuchome still hardcoded and no way to change it, can we move it into a var instead?

evrardjp:
I think it would be neat to backport it, but it's not per se necessary.

jsuchome:
@Itxaka Ouch. I changed to the var yesterday but forgot to push the change... now it's updated

vars/common-vars.yml
```diff
@@ -17,6 +17,7 @@ socok8s_libvirtuuid: "{{ socok8s_workspace }}/libvirt.uuid"
 suse_osh_registry_location: "docker.io"
 suse_openstack_image_version: "rocky-opensuse_15"
 suse_infra_image_version: "latest-opensuse_15"
+suse_osh_openstack_release: "rocky"
```
toabctl: why `suse`? isn't that osh specific?

jsuchome: Initially, just to make it similar as other variables above.

And I think there might be a time it could be used by airship as well

toabctl: I dont' like the same but I don't have a better idea :)

jsuchome: I can change it to `osh_openstack_release` if it would make you happy

jsuchome:
> If you still feel like changing the variable name, please to so. But I am also fine with this as is.

When I have green CI and 2 approvals? You must be joking :-)

PR #566, opened 2019-06-19
bump automation submodule

rhafer:
Commmit 670a76de728add5a77bf6149638bbaa3d894bf21 accidently reverted to
an older checkout of automation, which lacks recent fixes for the SES
deployment. This commits move us to the current master state.


toabctl:
Merging given that everything else in the CI is failing currenty due to this..

PR #567, opened 2019-06-19
Decouple airship-setup-deployer from common socok8s vars

toabctl:
Try to keep roles as separated as possible. That makes the role
reusable also for non socok8s cases and the role has an interface (the
variables) to interact with.
Note: airship-setup-deployer still depends on the common-deployer
role.
x

### DIFF
--- a/playbooks/generic-deploy_airship.yml
+++ b/playbooks/generic-deploy_airship.yml
@@ -2,8 +2,16 @@
 - hosts: soc-deployer
   gather_facts: yes
   any_errors_fatal: true
+  vars_files:
+    - "{{ playbook_dir }}/../vars/common-vars.yml"
+    - "{{ playbook_dir }}/../vars/deploy-on-openstack.yml"
   roles:
     - role: airship-setup-deployer
+      vars:
+        dcm_vip: "{{ socok8s_dcm_vip }}"
+        ext_vip: "{{ socok8s_ext_vip }}"
+        zypper_extra_repos:
+          socok8s: "{{ socok8s_repos_to_configure['socok8s'] }}"
       when: redeploy_osh_only is not defined or not redeploy_osh_only
       tags:
         - install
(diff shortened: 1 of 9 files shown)


evrardjp:
I like the idea to decouple the role from the common vars, and inject those. :+1: 
on the principle.

I didnt' get the chance to dig into the code fully.

playbooks/roles/airship-setup-deployer/defaults/main.yml
```diff
@@ -1,9 +1,19 @@
 ---
 # defaults file for suse-airship-deploy
 
+# IP on the Data Center Management (DCM) network
+dcm_vip: null
+
+# IP on the external network
+ext_vip: null
```
Itxaka: do we need to add these? isnt that covered by the change above?

toabctl: It is covered. But having it here is imo good to have the available variables documented.

toabctl: @Itxaka is that answer ok? Or do you want me to change something here?

playbooks/generic-deploy_airship.yml
```diff
@@ -2,8 +2,19 @@
 - hosts: soc-deployer
   gather_facts: yes
   any_errors_fatal: true
+  pre_tasks:
+    - name: Load generic vars
+      include_vars: "{{ item }}"
+      loop:
+        - "{{ playbook_dir }}/../vars/common-vars.yml"
+        - "{{ playbook_dir }}/../vars/deploy-on-openstack.yml"
```
kantarun: Not clear if there is any specific purpose of using vars as pre_tasks. If its just vars availability across roles, may be  vars_files will be more concise ?  https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#defining-variables-in-files

toabctl:
> Just searching of this vip variable in socok8s repo, there are some more lingering references which needs to be updated as well

@kantarun no. I do not globally change the variable names. I'm only changing the names in the role `airship-setup-deployer` and make the role `airship-setup-deployer` independent from our global variable definitions .

playbooks/generic-deploy_airship.yml
toabctl: Fixed. Thanks!

PR #569, opened 2019-06-19
[1.0] backport some fixes from master

rhafer:
This contains the cherry-picked backports for 
https://github.com/SUSE-Cloud/socok8s/pull/487
and
https://github.com/SUSE-Cloud/socok8s/pull/558


rhafer:
> Are they absolutely required for stable/1.0?

Nothing is absolutely stable :)

But maybe it helps you do know, that both commits only touch places that are either exclusively used in CI or for ECP based deploylments. So the risk for breaking the TP is very low. Additionally they should stabilize and speed up the ci a bit.

PR #571, opened 2019-06-20
Enhance ucp pod status check

jgu17:
Added check for keystpne and shipyard api readyness check and wait for
all pods so deployment won't continue to osh in the case of ucp charts
are still being updated.

Remove the wait flag in the armada apply cmd in order not to force everything into sequential deployment and let the armada chart wait property to take into effect.


playbooks/roles/airship-deploy-ucp/tasks/main.yml
```diff
@@ -244,40 +244,80 @@
...
+      delay: 10
+      changed_when: False
+      
     - name: Wait until Armada api pod is deployed
       command: 'kubectl get pod -l application=armada,component=api -n {{ ucp_namespace_name }} -o jsonpath="{.items[0].metadata.name}"'
       register: armada_results
       until: armada_results.stdout.find('armada-api-') == 0
-      retries: 180
+      retries: 240
       delay: 10
```
rajdharwadkar: Can we make it configurable ?

PR #572, opened 2019-06-20
Remove custom modifications of rootwrap filters

jsuchome:
The change was merged upstream: https://review.opendev.org/#/c/664243/


jsuchome:
Oh, I think we need to update the version of neutron chart we are using in shipyard ....

PR #570, opened 2019-06-19
Switch to a smaller flavor for SES the instance

rhafer:
We don't seem to need 16GB instances there currently. We originally
switch to d4.xlarge because of the ephemeral disk. But that is provided
by "cloud-ses" as well. Switching to a smaller flavor should save some
resources in the Cloud.

PR #573, opened 2019-06-20
[1.0] Switch to a smaller flavor for SES the instance

rhafer:
We don't seem to need 16GB instances there currently. We originally
switch to d4.xlarge because of the ephemeral disk. But that is provided
by "cloud-ses" as well. Switching to a smaller flavor should save some
resources in the Cloud.

(cherry picked from commit 488d4760033fb21cebdf9bbf81d3a31dfced0785)

Backport of https://github.com/SUSE-Cloud/socok8s/pull/570


rhafer:
Again something that shouldn't affect stability of the TP, but improve CI.

PR #575, opened 2019-06-20
Jenkinsfile: Do CaaSP worker setup in parallel to SES setup

toabctl:
We can reduce the CI run time a bit (ca. 5 min looking at previous
logs) when we configure the CaaSP workers after the deployer and CaaSP
are done (not waiting for SES). The it will look like:

    |---Deployer-|
    |            |---CaaSP worker--|
    |            |                 |
    |---CaaSP----|                 |---
----|                              |
    |                              |
    |                              |
    |---SES------------------------|

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -114,9 +114,28 @@ pipeline {
                 timeout(time: 45, unit: 'MINUTES', activity: true)
             }
             parallel {
-                stage('Deploy CaaSP') {
-                    steps {
-                        sh "./run.sh deploy_caasp"
+                stage('Deploy CCP deployer and CaaSP') {
+                    parallel {
+                        stage('Deploy CaaSP') {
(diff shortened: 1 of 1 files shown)


rhafer:
@jsuchome Did you try something like this already? IIRC it's not possible to nest parallel stages.

jsuchome:
Yeah, but I do not remember how it ended. Either it was not possible, or it kind of worked but UI was confused, or we found out that we have the dependency wrong and it actually cannot be paralelized.

I think it was third one, but maybe the code has changed meanwhile and dependencies weakened.

toabctl:
> IIRC it's not possible to nest parallel stages.

@rhafer hm. According to https://jenkins.io/doc/book/pipeline/syntax/#parallel it's not possible. I thought I read somewhere that it is possible. Closing for now. Sorry for the noise.

PR #577, opened 2019-06-20
doc: add resolution for RabbitMQ timing error (SOC-9536)

csymons-suse:


csymons-suse:
> Assuming this was tested..

Grace Tai reported this based on her testing.

PR #578, opened 2019-06-20
Enhance ucp pod status check

jgu17:
Added check for keystpne and shipyard api readyness check and wait for
all pods so deployment won't continue to osh in the case of ucp charts
are still being updated.

(cherry picked from commit 9c7e4e67e338808b4859135a94ae40eb53671698)


jgu17:
will redo it

PR #579, opened 2019-06-20
Enhance UCP status check for the ucp service update use case

jgu17:
Added check for keystone and shipyard api readiness and wait for all
pods. Otherwise, the update workflow may continue to overcloud even when
some ucp charts are still being updated. In the update use, armada api
will always be ready if no update for the armada pod.

Removed the wait flag in the armada apply cmd in order not to force
everything into sequential deployment from the cli and let the armada
chart wait property to take into effect.

### DIFF
--- a/playbooks/roles/airship-deploy-ucp/tasks/main.yml
+++ b/playbooks/roles/airship-deploy-ucp/tasks/main.yml
@@ -244,40 +244,80 @@
 
 # TODO(aagate): Add a changed_when: to help idempotency
 - name: Deploy Airship UCP ... quick coffee break, maybe?
-  command: 'kubectl exec {{ armada_pod_name }} -n ucp -- armada apply /armada/{{ socok8s_site_name }}-ucp.yaml --wait --target-manifest ucp-bootstrap'
+  command: 'kubectl exec {{ armada_pod_name }} -n ucp -- armada apply /armada/{{ socok8s_site_name }}-ucp.yaml --target-manifest ucp-bootstrap'
   tags:
     - install
     - skip_ansible_lint
 
 - block:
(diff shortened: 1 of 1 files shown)


playbooks/roles/airship-deploy-ucp/tasks/main.yml
```diff
@@ -244,40 +244,80 @@
...
-        msg: "armada-api pod found: {{ armada_api_pod_name }}"
-
-    - name: Wait until Airship api becomes ready
+    - name: Wait until Armada api becomes ready
       command: "kubectl get pod {{ armada_api_pod_name }} -n {{ ucp_namespace_name }} -o jsonpath='{.status.containerStatuses[].ready}'"
       register: armada_api_pod_status
       until: armada_api_pod_status.stdout == "true"
-      retries: 60
+      retries: 120
       delay: 10
```
rajdharwadkar: Can we make retries configurable ?

playbooks/roles/airship-deploy-ucp/tasks/main.yml
```diff
@@ -244,40 +244,80 @@
...
+      delay: 10
+      changed_when: False
+
     - name: Wait until Armada api pod is deployed
       command: 'kubectl get pod -l application=armada,component=api -n {{ ucp_namespace_name }} -o jsonpath="{.items[0].metadata.name}"'
       register: armada_results
       until: armada_results.stdout.find('armada-api-') == 0
-      retries: 180
+      retries: 240
       delay: 10
```
rajdharwadkar: Can we make retries configurable ?

jesusaurus: We can do that in a follow-up change

PR #580, opened 2019-06-20
Enhance UCP status check for the ucp service update use case

jgu17:
Added check for keystone and shipyard api readiness and wait for all
pods. Otherwise, the update workflow may continue to overcloud even when
some ucp charts are still being updated. In the update use, armada api
will always be ready if no update for the armada pod.

Removed the wait flag in the armada apply cmd in order not to force
everything into sequential deployment from the cli and let the armada
chart wait property to take into effect.

(cherry picked from commit 87af0b17ad2b05ad8d77fe5eb662700e5e785dde)


jgu17:
redo the cherry pick

PR #584, opened 2019-06-20
[1.0] Update openstack helm libvirt image tag to 20190620

jgu17:
Update to the new date tag 20190620 which contains the libvirt fix that
includes the missing block rbd pakcage.


jgu17:
travis is stuck somewhere. any idea how to re-trigger the travis ci?

PR #587, opened 2019-06-21
Jenkinsfile: Do ansible lint and doc build in parallel

toabctl:
We already have conditionals to not always lint or build the docs. But
in case both is needed, do it in parallel.


toabctl:
I wonder if we should also create the network in parallel (even if we might not need it in case linting fails). That would save us another minute or so...

PR #595, opened 2019-06-26
Fix CI against official JeOS images

dirkmueller:
which is not installed by default in the openSUSE official
JeOS images as well as it currently has the "no DNS" bug.
which we can workaround for now.


Itxaka:
CI passed but its stuck at the end, not reporting and not moving to done...

PR #600, opened 2019-06-27
Build against Leap 15.1 deployer image

dirkmueller:
This switches to the official JeOS image for openSUSE Leap 15.1.
Unfortunately this one doesn't have python 2.x embedded so
we need to be ultra correct in specifying the ansible python interpreter
everywhere. On the bright side that actually fixes the reality to
match the documentation (documenation didn't say that python 2.x
is required while it was actually required).

### DIFF
--- a/playbooks/generic-deploy_airship.yml
+++ b/playbooks/generic-deploy_airship.yml
@@ -19,12 +19,27 @@
   hosts: ses_nodes[0]
   remote_user: root
   gather_facts: true
+  vars:
+    ansible_python_interpreter: /usr/bin/python2
   roles:
     - common-ses
 
-- hosts: soc-deployer:caasp-workers
+- hosts: soc-deployer
   gather_facts: false
   any_errors_fatal: true
+  vars:
+    ansible_python_interpreter: /usr/bin/python3
+  roles:
+    - role: registry-server-setup
+      tags:
+        - install
+        - imagebuilder
+
+- hosts: caasp-workers
+  gather_facts: false
+  any_errors_fatal: true
+  vars:
+    ansible_python_interpreter: /usr/bin/python2
   roles:
     - role: registry-server-setup
       tags:
@@ -44,6 +59,8 @@
 - hosts: caasp-workers
   gather_facts: yes
   any_errors_fatal: true
+  vars:
+    ansible_python_interpreter: /usr/bin/python2
   roles:
     - role: airship-configure-worker
       when: redeploy_osh_only is not defined or not redeploy_osh_only
--- a/playbooks/generic-clean_airship.yml
+++ b/playbooks/generic-clean_airship.yml
@@ -17,6 +17,8 @@
 - hosts: soc-deployer
   gather_facts: no
   any_errors_fatal: true
+  vars:
+    ansible_python_interpreter: /usr/bin/python3
   tasks:
     - name: Remove compute nodes.
       include_role:
@@ -27,6 +29,8 @@
 
 - hosts: soc-deployer
   gather_facts: no
+  vars:
+    ansible_python_interpreter: /usr/bin/python3
   pre_tasks:
     - name: Load generic vars
       include_vars: "{{ playbook_dir }}/../vars/common-vars.yml"
@@ -37,6 +41,8 @@
 - hosts: caasp-workers
   gather_facts: no
   any_errors_fatal: true
+  vars:
+    ansible_python_interpreter: /usr/bin/python2
   tasks:
     - name: Clean airship related existing images on worker node
       shell: |
--- a/playbooks/generic-clean_k8s.yml
+++ b/playbooks/generic-clean_k8s.yml
@@ -17,6 +17,8 @@
 
 - hosts: soc-deployer
   gather_facts: no
+  vars:
+    ansible_python_interpreter: /usr/bin/python3
   tasks:
     - name: Running cleanup script
       script: "{{ playbook_dir }}/../script_library/cleanup-k8s.sh"
--- a/playbooks/generic-collect-logs.yml
+++ b/playbooks/generic-collect-logs.yml
@@ -18,6 +18,7 @@
 - hosts: soc-deployer
   gather_facts: false
   vars:
+    ansible_python_interpreter: /usr/bin/python3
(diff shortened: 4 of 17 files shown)


examples/workdir/inventory/hosts.yml
```diff
@@ -1,4 +1,8 @@
 ---
+localhost:
+  vars:
+   ansible_python_interpreter: /usr/bin/python3
```
Itxaka: cant we set this in the inventory directly instead?

dirkmueller: @Itxaka there is no inventory file for "localhost" as far as I can see.and I couldn't get this to work in all cases. is there a problem with listing this in the tasks? the problem is that if you set it in the inventory it didn't have an "effect" in all cases, leaving a weird mixture of python2 and python3 being executed, which fails if python2 is not there.

Itxaka: not a problem, but the inventory seems like the logical place for this IMO.

It not working in all cases kind of worries me :(

jesusaurus: That worries me too. I thought we found all the places we confused targeting localhost vs targeting the deployer while getting the job green in jenkins, but maybe there are still some mistakes?

dirkmueller: @jesusaurus yeah, but thats a matter of "our CI sucks" and we have an obvious solution here: extending CI coverage. If we agree the change is a good one and ci passes then we should merge it and deal with the fallout, including upgrading the ci coverage.

Itxaka:
mmmh, CI seems to be failing when adding repos, but I can reach them no problem....

dirkmueller:
@Itxaka every once in a while the mirroring is broken, unfortunately :(

examples/workdir/inventory/hosts.yml
evrardjp: This, behind the scenes, will create an explicit localhost entry in the ansible inventory, instead of relying on the host named "localhost" implicitly created by ansible.

For that implicit inventory host, ansible is using its own python version interpreter, which should be okay, assuming you install ansible under python3. This whole thing shouldn't therefore be used.

examples/workdir/inventory/hosts.yml
```diff
@@ -1,4 +1,8 @@
...
+
 caasp-admin:
   vars:
     ansible_user: root
```
evrardjp: The target nodes however will be tested against <python version of the deployer>. Assuming a venv is used for example, this will fail, and then cause a fallback to known python, which might be symlinked to python2.

For that, we should have ``ansible_python_interpreter: /usr/bin/python3`` for those groups.
While this seems to be done when generating the real inventory, I think we should also do it with the example.

Because there seems to be a lot of repetitions, I would do the following for caasp-admin:
```
caasp-admin:
  vars: &connectionsettings
    ansible_user: "root"
    ansible_python_interpreter: "/usr/bin/python3"
```

and then, for all the other groups, use:
```
vars: *connectionsettings
```
It seems it's only wanted for some groups, but you get the idea.

playbooks/generic-deploy_airship.yml
```diff
@@ -20,12 +20,27 @@
   hosts: ses_nodes[0]
   remote_user: root
   gather_facts: true
+  vars:
+    ansible_python_interpreter: /usr/bin/python2
```
evrardjp: what's the problem with python3 here? Anyway, assuming you have the inventory with python3, you can still override it here.

playbooks/openstack-update_and_reboot_caasp.yml
```diff
@@ -63,5 +63,5 @@
         validate_certs: no
       register: result
       until: result.status == 200
-      retries: 60
+      retries: 100
```
evrardjp: nit: probably shouldn't be on the same commit

playbooks/roles/handle-nodes-on-openstack/tasks/create_on_openstack.yml
```diff
@@ -37,6 +37,7 @@
         hosts:
           {{ servername }}:
             ansible_host: {{ floating_ip }}
+            ansible_python_interpreter: {{ ansible_python_interpreter }}
```
evrardjp: That could instead be the right interpreter from the start, instead of relying on the interpretation of said variable for the host. It would be clearer to debug.

examples/workdir/inventory/hosts.yml
evrardjp: sed caasp-admin with the appropriate host who needs python3.

playbooks/openstack-update_and_reboot_caasp.yml
dirkmueller: removed

playbooks/generic-deploy_airship.yml
dirkmueller: this runs on the ses nodes, which are SLE_12, which are python 2.x only

evrardjp: I expected this. What I meant is that we don't expect to need to override this on a play basis because something would not work without it set, but more generally about the nodes themselves having no python3. Which point me to the inventory thing.

Itxaka:
mmmh, pretty sure the submodule was a mistake unless its required by some reason

evrardjp:
> mmmh, pretty sure the submodule was a mistake unless its required by some reason

probably, but don't want to vote positively until this is settled :)

examples/workdir/inventory/hosts.yml
dirkmueller: @evrardjp I have removed the explicit localhost entry and moved the variable setting to other places.

playbooks/roles/handle-nodes-on-openstack/tasks/create_on_openstack.yml
dirkmueller: I removed this, was from a previous attempt that didn't work.

dirkmueller:
@evrardjp I removed the submodule change from the PR.

dirkmueller:
no longer needed, we switched to SLE15SP1

PR #601, opened 2019-07-01
Clarify socok8s install methods and ssh key prep

jgu17:
Clarification in the 3rd party software package install and ssh key
preperation.


jgu17:
redo it

PR #602, opened 2019-07-01
Clarify 3rd party packge installation, ssh key prep and file locations.

jgu17:
Clarifications on the 3rd party package install, ssh key prep to include
deployer node, and example file location.


doc/source/deployment/setup-deployer.rst
```diff
@@ -91,7 +91,7 @@ There are several ways to install the SUSE Containerized OpenStack software.
    After the required packages are installed, clone
```
sgrasley: Can we fix step 1 and 2 above this change?  It seems a bit redundant.

PR #604, opened 2019-07-02
docs: add ARA install

AbelNavarro:
In the docs for localhost install instructions it is recommended to use
ARA for Ansible. Added install instructions for it.

### DIFF
--- a/doc/source/suse-ecp/ose-localhost.rst
+++ b/doc/source/suse-ecp/ose-localhost.rst
@@ -69,6 +69,7 @@ Optionally, `localhost` can be preinstalled with the following software:
   * python3-jmespath
   * python3-openstacksdk
   * python3-netaddr
+  * python3-ara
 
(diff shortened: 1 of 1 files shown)


doc/source/suse-ecp/ose-localhost.rst
```diff
@@ -69,6 +69,7 @@ Optionally, `localhost` can be preinstalled with the following software:
   * python3-jmespath
   * python3-openstacksdk
   * python3-netaddr
+  * python3-ara
```
jsuchome: I don't think we have `python3-ara` in regular Leap 15 repos. AFAIK if you want ara, you need to install it into virtualenv

AbelNavarro: ah, that may be true since I'm using tumbleweed. Let's dismiss this PR then.

PR #605, opened 2019-07-02
Workspace dir needed for ARA config

AbelNavarro:
The directory needed for ARA config is created either when the virtual
environment is created or in the first call of run_ansible. In case you
are not using a virtual environment and you are using ARA that directory
will be missing and the pre-fligh-checks would fail. This patch adds that
directory in such case.

### DIFF
--- a/script_library/pre-flight-checks.sh
+++ b/script_library/pre-flight-checks.sh
@@ -11,6 +11,12 @@ check_common_env_vars_set (){
     fi
 
     echo "Using ${SOCOK8S_WORKSPACE_BASEDIR}/${SOCOK8S_ENVNAME}-workspace as workspace directory"
+
+    # Needed for the Ansible ARA check step
+    if [ ! -d ${SOCOK8S_WORKSPACE_BASEDIR}/${SOCOK8S_ENVNAME}-workspace ]; then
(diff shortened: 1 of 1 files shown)


Itxaka:
I am unable to hit this issue, after cleaning up my workspace directory and exporting `USE_ARA=yes` everything works when launching the steps.

Any more info regarding this? Is in a specific step?

AbelNavarro:
I was trying to run this without a virtual environment and it failed when I first run `./run.sh deploy_ses`.

AbelNavarro:
Notice this was failing at: https://github.com/SUSE-Cloud/socok8s/blob/master/script_library/pre-flight-checks.sh#L86 because the workspace dir didn't exist.

AbelNavarro:
> The change itself looks good, but the commit message is inaccurate and needs updating. The underlying issue is that the workspace directory is never explicitly created, it is implicitly created either during creation of the virtual environment (if using a virtual environment) or the first call to run_ansible while creating the `env/` subdirectory for the `extravars` file. In the second case, if `USE_ARA=True` then the workspace directory will not have been created before we try to put `ara.rc` in that directory.

rephrased the commit message. Hopefully looks better now.

PR #606, opened 2019-07-02
caasp4 cleanup: fix corner case

Itxaka:
There is an issue with the cleanup of caasp 4 that will try to
remove the terraform plan even if the terraform workspace
does not exists, which leads to a failure

This patch adds a check for the workspace existing to trigger the
task to remove the terraform plan

### DIFF
--- a/playbooks/roles/caasp4/tasks/cleanup_nodes.yml
+++ b/playbooks/roles/caasp4/tasks/cleanup_nodes.yml
@@ -19,6 +19,11 @@
 # stack in setup_nodes.yml so we need to remove the resources again from
 # terraform state and let heat handle the reource cleanup
 
+- name: Does the terraform workspace exist
+  stat:
+    path: skuba_ci_terraform_workspace
+  register: _terraform_workspace
+
 - name: Does the terraform state exist
   stat:
     path: "{{ skuba_ci_terraform_workspace }}/terraform.tfstate"
@@ -32,6 +37,7 @@
   terraform:
     project_path: "{{ skuba_ci_terraform_workspace }}"
     state: absent
+  when: _terraform_workspace.stat.exists
 
 - name: Drop skuba openstack terraform files
   file:


playbooks/roles/caasp4/tasks/cleanup_nodes.yml
```diff
@@ -19,6 +19,11 @@
...
 # terraform state and let heat handle the reource cleanup
 
+- name: Does the terraform workspace exist
+  stat:
+    path: skuba_ci_terraform_workspace
+  register: _terraform_workspace
+
 - name: Does the terraform state exist
   stat:
     path: "{{ skuba_ci_terraform_workspace }}/terraform.tfstate"
```
toabctl: Could we use this check instead of adding a new one?

Itxaka: is it possible that the tfstate file doesnt exist but the dir does? In that case, the workspace wont be totally cleaned up no?

evrardjp:
Not so sure about the problem, so not really sure we need this.

playbooks/roles/caasp4/tasks/cleanup_nodes.yml
toabctl: It's possible that the dir exists but tfstate not.

Itxaka:
The problem seems to be that the terraform module doesn't provide for a nice fallback when using absent if the dir doesn't exists (like wut?) which turns into a failure.

This can be tested by just calling the caasp4 cleanup with no skuba_ci_terraform_workspace dir

playbooks/roles/caasp4/tasks/cleanup_nodes.yml
Itxaka: in that case, we cannot use the tfstate for that as it would leave the dir around. Unless we dont care about that and only care about the tfstate file, in which case the dir cleanup can be dropped

toabctl: ok

PR #607, opened 2019-07-03
doc: deploy_network prerequisite for manual steps

AbelNavarro:
In the doc section to deploy SES and CaaSP for SUSE developers, there's
a step-by-step section to deploy them with run.sh, but it needs the
'deploy_network' step before. That step is already included in the
aggregated 'setup_hosts' step.


Itxaka:
cc @csymons-suse

PR #608, opened 2019-07-03
bump osh repo for deploy_osh

Itxaka:
All repos are unfrozen when deploying just osh but openstack-helm-infra
This is currently resulting in an issue where the neutron chart from
master requires a change in osh-infra but infra is frozen to an older
version.

This bumps osh-infra to master like the other repos. This config is
only applied when using deploy_osh so it should not affect airship
deployment as that uses a different repo tracking
(site/soc/software/config/versions.yml)

### DIFF
--- a/vars/manifest.yml
+++ b/vars/manifest.yml
@@ -2,7 +2,7 @@
 upstream_repos:
   openstack-helm-infra:
     src: https://opendev.org/openstack/openstack-helm-infra
-    version: 565fb4606b54444f4ea126a598ec0519c8aec75c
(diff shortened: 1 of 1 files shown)


Itxaka:
This fails to apply @sgrasley multiple VIP patch because its on merge conflict. I asked him to rebase it if possible because currently deploy_osh is broken which makes it difficult to work in the charts.

Itxaka:
requires #612

Itxaka:
rebased now that  #612 is merged

PR #610, opened 2019-07-03
Update playbook to remove become attribute.

sgrasley:
OpenSUSE Leap 15.0 updated ansible to version 2.8. New version of
ansible produces this error.

ERROR! 'become' is not a valid attribute for a IncludeRole

Become was not necessary for this task because already running as root.

### DIFF
--- a/playbooks/roles/common-ses/tasks/main.yml
+++ b/playbooks/roles/common-ses/tasks/main.yml
@@ -3,7 +3,6 @@
   include_vars: "{{ playbook_dir }}/../vars/common-vars.yml"
 
 - name: Create pools in SES for socok8s usage
-  become: yes
   include_role:
     name: ses
     tasks_from: openstack_config


toabctl:
This is a regression in openSUSE. @sgrasley Did you report it there, too?

sgrasley:
> This is a regression in openSUSE. @sgrasley Did you report it there, too?

I have not reported this.  Not sure how it is an openSUSE regression. Can you explain?

toabctl:
> I have not reported this. Not sure how it is an openSUSE regression. Can you explain?

@sgrasley according to your commit msg, openSUSE 15.0 updated ansible (which was done via https://build.opensuse.org/request/show/711325 ). Due to this update, we get now `ERROR! 'become' is not a valid attribute for a IncludeRole`. That's a regression if you use ansible on openSUSE 15.0

PR #612, opened 2019-07-05
Drop VIP patch

Itxaka:
Looks like the patch is not gonna be merged upstream and its
currently on merge conflict. Talked with Scott and he seems to
agree with this so the recomendation is to drop it and look
for an alternative solution

Meanwhile having the pathc in there block us from updating osh-infra
to the latest commit, which is necessary as neutron deployment
is currently broken due to a missing feature from htk which is
only available on the latest git versions, hence the need to
bump osh-infra to master for now.

### DIFF
--- a/doc/source/deployment/configure.rst
+++ b/doc/source/deployment/configure.rst
@@ -205,20 +205,6 @@ For example:
    socok8s_ext_vip: "10.10.10.10"
 
 
-Configure the VIP that will be used for Airship UCP service endpoints
---------------------------------------------------------------------------
-
-Add `socok8s_dcm_vip:` with its appropriate value for your
-environment in your `env/extravars`. This should be an available IP
-on the Data Center Management (DCM) network (in development environment, it
-can be the same as CaaSP cluster network).
-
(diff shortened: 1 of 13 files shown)


Itxaka:
without the patch it seems like airship is unable to deploy keystone. We need a solution for this ASAP because currently everything is broken. osh-infra is frozen because of this patch, we need osh-infra to be bumped to a newest version because osh neutron needs the latest osh-infra, but the patch in in conflict with the latest osh-infra.

Itxaka:
> in the playbooks/roles/airship-setup-deployer/tasks/main.yml, change socok8s_dcm_vip to socok8s_ext_vip for now until we integrate with metallb or external lb to have proper multiple vips

THANK YOU!

I was looking for this but could not find where it was the change needed to make it work. :+1: :+1: :+1:

Itxaka:
CC @csymons-suse: Change in docs to drop the dcm_vip mentions/config/seyup

Itxaka:
CI passing!

PR #614, opened 2019-07-05
Use SLE15 for the deployer

Itxaka:
Uses SLE15 for the deployer.

 - Removes 2 lines from the velum Gemfile.lock for bundle 2.0 compatibility
 - Adds an extra vendor from the devel:languages:ruby for zypper
 - Drops rng-tools from the required packages as it doesnt seem to be used
 - Adds the needed repos for SLES15-JeOS
 - Changes the openstack image used from Leap 15.0 to SLES15-JeOS

### DIFF
--- a/vars/common-vars.yml
+++ b/vars/common-vars.yml
@@ -32,6 +32,20 @@ socok8s_repos_to_configure:
   openSUSE-Leap-15.0-Cloud: "{{ obs_mirror }}/repositories/Cloud:/Tools/openSUSE_Leap_15.0/"
   Leap15develtools: "{{ obs_mirror }}/repositories/devel:/tools/openSUSE_Leap_15.0/"
   socok8s: "{{ obs_mirror }}/repositories/Cloud:/socok8s:/master/openSUSE_Leap_15.0/"
+  SLES15-JeOS-OpenStack-Cloud-GM-Pool: http://download.suse.de/ibs/SUSE:/SLE-15:/GA/standard/
+  SLES15-JeOS-OpenStack-Cloud-GM-Update: http://download.suse.de/ibs/SUSE:/SLE-15:/Update/standard/
+  # needed for python-barbicanclient
+  SLES15-JeOS-OpenStack-Cloud-GM-backports: "{{ obs_mirror }}/repositories/openSUSE:/Backports:/SLE-15/standard/"
+  # may not be needed?
+  SLES15-JeOS-OpenStack-Cloud-GM-Cloud:  "{{ obs_mirror }}/repositories/Cloud:/Tools/SLE_15/"
+  # needed for gcc
+  SLES15-JeOS-OpenStack-Cloud-GM-develtools: "{{ ibs_mirror }}/ibs/SUSE/Products/SLE-Module-Development-Tools/15/x86_64/product/"
+  # needed for phantomjs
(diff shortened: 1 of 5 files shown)


Itxaka:
SLE15-SP1 support may come later on monday

vars/common-vars.yml
```diff
@@ -32,6 +32,20 @@ socok8s_repos_to_configure:
   openSUSE-Leap-15.0-Cloud: "{{ obs_mirror }}/repositories/Cloud:/Tools/openSUSE_Leap_15.0/"
   Leap15develtools: "{{ obs_mirror }}/repositories/devel:/tools/openSUSE_Leap_15.0/"
   socok8s: "{{ obs_mirror }}/repositories/Cloud:/socok8s:/master/openSUSE_Leap_15.0/"
+  SLES15-JeOS-OpenStack-Cloud-GM-Pool: http://download.suse.de/ibs/SUSE:/SLE-15:/GA/standard/
```
toabctl: Should we start to split this? Eg. introduce `socok8s_deployer_repos` or something like this?

Itxaka: it may make sense if we want to support multiple OS for the deployer

toabctl: I was more thinking that the variable `socok8s_repos_to_configure` might be used in multiple places (on different hosts). But if we anyway use it only on the deployer, than fine for me to just add more repos..

dirkmueller:
@Itxaka to some extend this conflicts with my leap 15.1 enablement work. can we decide in which order we move forward?

Itxaka:
> @Itxaka to some extend this conflicts with my leap 15.1 enablement work. can we decide in which order we move forward?

I was hoping we could support both leap && SLE, but as far as I understood, the product will only support SLE once released no?

To be fair this is not clear enough yet, what do we need to support in the final Cloud10 release? Cant we work on leap && SLE in parallel?

toabctl:
> I was hoping we could support both leap && SLE, but as far as I understood, the product will only support SLE once released no?

Yes. The product will only support SLES.

> To be fair this is not clear enough yet, what do we need to support in the final Cloud10 release? Cant we work on leap && SLE in parallel?

We need to support SLES, but having openSUSE working has a couple of advantages (like being able to use it for CI stuff)

Itxaka:
> Yes. The product will only support SLES.

Then its clear, we need to jump on the SLE train asap :)

> We need to support SLES, but having openSUSE working has a couple of advantages (like being able to use it for CI stuff)

But we can use SLE for CI with no problems, which is also better because we test what customers will use.

Itxaka:
So I just checked, and we could work with this in parallel no issues, @dirkmueller patch will only conflict with this in the `deploy_on_openstack_oshnode_image` var. The rest is not only compatible but some of it needed (ansible_python_interpreter for once)

evrardjp:
I thought the move to sle 15 was only expected for caasp4, won't this be repeated work?

Itxaka:
> I thought the move to sle 15 was only expected for caasp4, won't this be repeated work?

This is for the deployer in preparation for cloud10 release. In any case, this PR is a testbed, a new one will come with 15-sp1 support based off this

evrardjp:
Else it looks okay to me.

Itxaka:
Moving to SP1 so this can be closed: https://github.com/SUSE-Cloud/socok8s/pull/628

PR #615, opened 2019-07-08
Caasp4 more providers

toabctl:


playbooks/kvm-deploy_caasp4.yml
```diff
@@ -0,0 +1,45 @@
...
+        tasks_from: setup_terraform_workspace_libvirt
+      vars:
+        skuba_ci_terraform_worker_count: "{{ deploy_on_kvm_caasp_workers }}"
+        skuba_ci_terraform_workspace: "{{ socok8s_workspace }}/skuba-terraform/"
+
+    - import_role:
+        name: caasp4
+        tasks_from: setup_nodes
+      vars:
+        _import_terraform_resources: false
```
Itxaka: This seems like it should be a normal var. The `_` seems to indicate that is temporary, butr no other declared var from playbooks (other that commmand outputs) is using this, so its a bit confusing to track it down.

toabctl: Removed the `_`. Marking as solved conversation for now.

toabctl:
@Itxaka could you review again please?

playbooks/kvm-delete_caasp4.yml
```diff
@@ -0,0 +1,30 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+
+- hosts: localhost
+  gather_facts: yes
+  any_errors_fatal: true
```
evrardjp: Do we need this? If true, can you explain why in a comment?

toabctl: I don't think we need it. Copy&paste artifact.

PR #622, opened 2019-07-12
Don't try to remove a non-existent caaspv4 cluster

jesusaurus:
Attempting to run a teardown action on a default deployment fails due to
a missing terraform binary. Remove the caaspv4 cleanup from the teardown
action since a default deployment will deploy caaspv3. The caaspv3
cleanup can be replaced with the caaspv4 cleanup when we update the
setup_hosts action to deploy caaspv4 instead of caaspv3.

### DIFF
--- a/script_library/deployment-actions-openstack.sh
+++ b/script_library/deployment-actions-openstack.sh
@@ -63,7 +63,6 @@ function clean_openstack(){
     run_ansible ${socok8s_absolute_dir}/playbooks/openstack-osh_instance.yml -e osh_node_delete=True
     echo "Delete CaaSP 3 nodes"
     run_ansible ${socok8s_absolute_dir}/playbooks/openstack-delete_caasp.yml
-    clean_caasp4
     echo "Delete SES node"
     run_ansible ${socok8s_absolute_dir}/playbooks/openstack-ses_aio_instance.yml -e ses_node_delete=True
     echo "Delete network stack"


script_library/deployment-actions-openstack.sh
```diff
@@ -63,7 +63,6 @@ function clean_openstack(){
     run_ansible ${socok8s_absolute_dir}/playbooks/openstack-osh_instance.yml -e osh_node_delete=True
     echo "Delete CaaSP 3 nodes"
     run_ansible ${socok8s_absolute_dir}/playbooks/openstack-delete_caasp.yml
-    clean_caasp4
```
toabctl: Not sure how you did run into the problem you describe in the commit message. There is already a check that the terraform binary is available (see https://github.com/SUSE-Cloud/socok8s/blob/master/script_library/deployment-actions-openstack.sh#L55 ). Also I would really like to keep the caasp4 cleanup for people who deploy caasp4. Otherwise we waste resources in case the deployment is on the engcloud

jgu17: so how do people remove caasp v4 if this line is deleted?

jesusaurus: by manually running `run.sh clean_caasp4`, which makes sense since deploying caasp4 requires a manual `run.sh deploy_caasp4`.

jesusaurus: The check for the terraform binary only occurs during a caasp4 deployment. If we have deployed a caasp3 stack, we may not have terraform installed.

toabctl: I still don't get it - the `clean_caasp4` function does the check if the terraform binary is available. So we have 2 cases:

- If the binary is not available, the function `clean_caasp4` does noting. 
- I the binary is available, `playbooks/openstack-delete_caasp4.yml` is called which triggers `playbooks/roles/caasp4/tasks/cleanup_nodes.yml` which should do nothing if no terraform workspace is found.

Am I missing something?

jesusaurus: If the binary is not available, then the failed check will exit the script before we delete the SES node and the network stack.

jesusaurus:
I'm closing this because a workaround is to install the binaries needed for caasp4 deployments, and we are moving in that direction anyway.

PR #623, opened 2019-07-12
Delegate cert copy to the host with the cert

jesusaurus:
We generate the docker registry certificate on the deployer host, in the
event that the ansible runner host is not the deployer host, the copy
task will fail because there is no certificate on the ansible localhost.
Delegate the task to the first host in the host group we generated the
certificate on.

### DIFF
--- a/playbooks/roles/registry-server-setup/tasks/apply-certs.yml
+++ b/playbooks/roles/registry-server-setup/tasks/apply-certs.yml
@@ -4,6 +4,7 @@
   copy:
     src: "{{ suse_registry_certs_dir }}/domain.crt"
     dest: "/etc/pki/trust/anchors"
+  delegate_to: "{{ groups['soc-deployer'][0] }}"
   become: yes
   register: _copy_ssc_result
 


playbooks/roles/registry-server-setup/tasks/apply-certs.yml
```diff
@@ -4,6 +4,7 @@
   copy:
     src: "{{ suse_registry_certs_dir }}/domain.crt"
     dest: "/etc/pki/trust/anchors"
+  delegate_to: "{{ groups['soc-deployer'][0] }}"
```
evrardjp: isn't the generation of the cert on the localhost anymore?

jesusaurus: no, the cert generation happens in the registry-server-setup role which runs on the soc-deployer hostgroup

evrardjp: mmm. https://github.com/SUSE-Cloud/socok8s/blob/master/playbooks/generic-generate_certs_for_local_registry.yml that was what to reuse. Not sure how we managed to make that on a different node.

PR #624, opened 2019-07-15
Changes made to replace airship components base image from leap15.0 to le…

rajdharwadkar:
…ap15.1

* update base image

### DIFF
--- a/site/soc/software/config/versions.yaml
+++ b/site/soc/software/config/versions.yaml
@@ -83,6 +83,12 @@ data:
         reference: 3811eedefe92f1df6c1d31a366773b9a6dce0b7b
         subpath: tempest
         type: git
+    ucp:
+      shipyard:
+        location: https://opendev.org/airship/shipyard
+        reference: 6a54c628281eaad8526ca1a4dd28f79c3ad23405
+        subpath: charts/shipyard
+        type: git
   images:
     calico:
       calico: {}
--- a/vars/common-vars.yml
+++ b/vars/common-vars.yml
@@ -64,7 +64,7 @@ ucp_namespace_name: "{{ ucp_namespace | default('ucp') }}"
 
 # Variables used for building airship images locally
 suse_distro_identifier: opensuse_15
-suse_airship_components_base_image: "opensuse/leap:15.0"
+suse_airship_components_base_image: "opensuse/leap:15.1"
 
 # check if argument is set in commandline for airship_local_images_flag value
 # otherwise use corresponding environment variable to determine it
@@ -106,11 +106,11 @@ suse_airship_components_image_tag: |
                                    %}
                                    {% else %}
                                    {% set _img_tag = {
-                                     'armada': '1110c23cd192ea5eff77629d10924b290140d6e2-opensuse_15',
-                                     'shipyard': 'cd6f154da56079ba408b922ed748a80b622c3592-opensuse_15',
-                                     'airflow': 'cd6f154da56079ba408b922ed748a80b622c3592-opensuse_15',
-                                     'deckhand': 'd8c8e15074d484c356c03460c0acbaea96705150-opensuse_15',
-                                     'pegleg': '652c3abefd0f647cfc45096aa76cb76bfaf8a805-opensuse_15'
+                                     'armada': '46bb9bd8b35905294920eb4fdf30593cb4899780-opensuse_15',
+                                     'shipyard': 'a9ff859df980fbb95a912fc49312ef3c51f65b3f-opensuse_15',
+                                     'airflow': 'a9ff859df980fbb95a912fc49312ef3c51f65b3f-opensuse_15',
+                                     'deckhand': '971c901c4da36e3e31be58d4be8b0e194e715d8f-opensuse_15',
+                                     'pegleg': '70f2db465277b60d80e4bcbf347be565b6f81a05-opensuse_15'
                                    }
                                    %}
                                    {% endif %}


rajdharwadkar:
> I'm running into this error during the "Create config docs in Shipyard" task:
> 
> ```
>   stdout: |-
>     Traceback (most recent call last):
>       File "/usr/bin/shipyard", line 10, in <module>
>         sys.exit(shipyard())
>       File "/usr/lib/python3.6/site-packages/click/core.py", line 722, in __call__
>         return self.main(*args, **kwargs)
>       File "/usr/lib/python3.6/site-packages/click/core.py", line 697, in main
>         rv = self.invoke(ctx)
>       File "/usr/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
>         return _process_result(sub_ctx.command.invoke(sub_ctx))
>       File "/usr/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
>         return _process_result(sub_ctx.command.invoke(sub_ctx))
>       File "/usr/lib/python3.6/site-packages/click/core.py", line 895, in invoke
>         return ctx.invoke(self.callback, **ctx.params)
>       File "/usr/lib/python3.6/site-packages/click/core.py", line 535, in invoke
>         return callback(*args, **kwargs)
>       File "/usr/lib/python3.6/site-packages/click/decorators.py", line 17, in new_func
>         return f(get_current_context(), *args, **kwargs)
>       File "/usr/lib/python3.6/site-packages/shipyard_client/cli/create/commands.py", line 182, in create_configdocs
>         with open(_file, 'r') as stream:
>     PermissionError: [Errno 13] Permission denied: '/target/peggles/treasuremap.yaml'
> ```

That error is not related with my changes.  I remember similar permission issue was fixed in stable branch. may be we need to merge that into master.

vars/common-vars.yml
```diff
@@ -89,11 +89,11 @@ suse_airship_components_image_tag: |
...
-                                     'armada': '1110c23cd192ea5eff77629d10924b290140d6e2-opensuse_15',
-                                     'shipyard': 'cd6f154da56079ba408b922ed748a80b622c3592-opensuse_15',
-                                     'airflow': 'cd6f154da56079ba408b922ed748a80b622c3592-opensuse_15',
-                                     'deckhand': 'd8c8e15074d484c356c03460c0acbaea96705150-opensuse_15',
-                                     'pegleg': '652c3abefd0f647cfc45096aa76cb76bfaf8a805-opensuse_15'
+                                     'armada': '46bb9bd8b35905294920eb4fdf30593cb4899780-opensuse_15',
+                                     'shipyard': 'a9ff859df980fbb95a912fc49312ef3c51f65b3f-opensuse_15',
+                                     'airflow': 'a9ff859df980fbb95a912fc49312ef3c51f65b3f-opensuse_15',
+                                     'deckhand': '971c901c4da36e3e31be58d4be8b0e194e715d8f-opensuse_15',
+                                     'pegleg': '70f2db465277b60d80e4bcbf347be565b6f81a05-opensuse_15'
```
kantarun: Confirmed the recent images on quay.io matches those tags.

vars/common-vars.yml
```diff
@@ -47,21 +47,21 @@ ucp_namespace_name: "{{ ucp_namespace | default('ucp') }}"
...
-# check if argument is set in commandline for airship_local_images_flag value
+# Flag to be used when just want to use previously built local airship component images
+# without need of re-building them.
+suse_airship_use_local_images: "{{ lookup('env','AIRSHIP_USE_LOCAL_IMAGES') | default('False', true) | bool }}"
+
+# check if argument is set in commandline for suse_airship_local_images_flag value
 # otherwise use corresponding environment variable to determine it
-suse_airship_build_local_images: "{%- if airship_local_images_flag is defined -%}
-                                    {{ airship_local_images_flag | bool }}
+suse_airship_build_local_images: "{%- if suse_airship_use_local_images_flag is defined -%}
```
kantarun: Why we are tying build local images flag to use local images flag. With this change, if use local images is enabled, then we are going to build local images everytime which is not a desired and unnecessary behavior. 
What's the reason for changing the logic now ?

rajdharwadkar: > Why we are tying build local images flag to use local images flag. With this change, if use local images is enabled, then we are going to build local images everytime which is not a desired and unnecessary behavior.
> What's the reason for changing the logic now ?

I tested it locally and it seems to be pulling image if export AIRSHIP_USE_LOCAL_IMAGES=true. 
so it not building image everytime.

rajdharwadkar: Looks like for building image both flags needs to be set. So reverting those changes.

vars/common-vars.yml
```diff
@@ -104,11 +104,11 @@ suse_airship_components_image_tag: |
...
-                                     'armada': '1110c23cd192ea5eff77629d10924b290140d6e2-opensuse_15',
-                                     'shipyard': 'cd6f154da56079ba408b922ed748a80b622c3592-opensuse_15',
-                                     'airflow': 'cd6f154da56079ba408b922ed748a80b622c3592-opensuse_15',
-                                     'deckhand': 'd8c8e15074d484c356c03460c0acbaea96705150-opensuse_15',
-                                     'pegleg': '652c3abefd0f647cfc45096aa76cb76bfaf8a805-opensuse_15'
+                                     'armada': '46bb9bd8b35905294920eb4fdf30593cb4899780-opensuse_15',
+                                     'shipyard': 'a9ff859df980fbb95a912fc49312ef3c51f65b3f-opensuse_15',
+                                     'airflow': 'a9ff859df980fbb95a912fc49312ef3c51f65b3f-opensuse_15',
+                                     'deckhand': '971c901c4da36e3e31be58d4be8b0e194e715d8f-opensuse_15',
+                                     'pegleg': '70f2db465277b60d80e4bcbf347be565b6f81a05-opensuse_15'
```
Itxaka: `opensuse_15` shouldnt it be 15.1? does the upstream airship images only point to 15 with no point release? seems like a problem waiting to happen :(

site/soc/software/config/versions.yaml
```diff
@@ -83,6 +83,12 @@ data:
         reference: 3811eedefe92f1df6c1d31a366773b9a6dce0b7b
         subpath: tempest
         type: git
+    ucp:
+      shipyard:
+        location: https://opendev.org/airship/shipyard
+        reference: 6a54c628281eaad8526ca1a4dd28f79c3ad23405
+        subpath: charts/shipyard
+        type: git
```
kantarun: I am assuming, this will be revisited when we sync site information maintained in socok8s repo with upstream version (originally derived from upstream treasuremap airskiff site).
Hopefully, we would not need to carry this override locally after sync with latest version of upstream.

vars/common-vars.yml
kantarun: In upstream, we wanted to keep this specific to a major release from maintenance perspective. The reason being, we want to maintain one version of Dockerfile per major release of distro. There is not much changed in (terms of package dependency, executable paths/ env etc) within minor releases of a major version.

PR #625, opened 2019-07-15
[1.0] Fix neutron dnsmasq failure

jgu17:
With latest caasp, dnsmasq fails in the Neutron dhcp pod without read
permission to the dnsmasq conf file. The root cause is missing
appropriate apparmor configuration for the dnsmasq.

(cherry picked from commit 1ef53fd337c8b0c8c1c1543e7be0814a913ba7a8)


jgu17:
> So we missed to port this from osh to airship :( that sucks :(

Yes :-(.  Also odd is that we didn't see the issue in GM testing at all.  May be related to some recent caasp updates.

PR #627, opened 2019-07-15
Make sure shipyard can read the pegleg output

jesusaurus:
The treasuremap.yaml file created by pegleg may not be readable by
shipyard, make sure it is world-readable.

### DIFF
--- a/playbooks/roles/airship-deploy-osh/tasks/main.yml
+++ b/playbooks/roles/airship-deploy-osh/tasks/main.yml
@@ -86,6 +86,13 @@
     - install
     - always
 
+- name: Make sure shipyard can read the pegleg output
+  become: yes
+  file:
+    path: '{{ upstream_repos_clone_folder }}/airship/shipyard/{{ pegleg_output }}'
+    recurse: yes
+    mode: '+r'
+
 # TODO(aagate): Add a changed_when: to help idempotency
 - name: Create config docs in Shipyard
   command: '{{ shipyard }} create configdocs {{ socok8s_site_name }}-design --replace --directory=/target/{{ pegleg_output }}'
@@ -97,6 +104,7 @@
   register: shipyard_create_config
   failed_when:
     - shipyard_create_config.stdout.find('Error') == 0
+    - shipyard_create_config.stdout.find('Traceback') == 0
     - "no_revision_error_msg not in shipyard_create_config.stdout"
   tags:
     - install


jesusaurus:
@kantarun the permission issue is happening inside the shipyard container, so running the ansible task as root doesn't make a difference.

playbooks/roles/airship-deploy-osh/tasks/main.yml
```diff
@@ -97,6 +104,7 @@
   register: shipyard_create_config
   failed_when:
     - shipyard_create_config.stdout.find('Error') == 0
+    - shipyard_create_config.stdout.find('Traceback') == 0
```
jgu17: good catch here... in what cases will shipyard use traceback?

playbooks/roles/airship-deploy-osh/tasks/main.yml
```diff
@@ -86,6 +86,13 @@
     - install
     - always
 
+- name: Make sure shipyard can read the pegleg output
+  become: yes
```
jgu17: what permission was this set to when shipyard can't read it?

rajdharwadkar:
We need this PR because I see CI failed with this permission error.
16:51:45      PermissionError: [Errno 13] Permission denied: '/target/peggles/treasuremap.yaml'

https://ci.suse.de/job/cloud-socok8s-jobs/job/socok8s-integration/view/change-requests/job/PR-624/5/execution/node/128/log/

playbooks/roles/airship-deploy-osh/tasks/main.yml
jesusaurus: In this particular case, the Traceback was a PermissionError trying to read treasuremap.yml

playbooks/roles/airship-deploy-osh/tasks/main.yml
jesusaurus: owned by root:root and mode 0640

kantarun:
Just clarifying the need of this change:

Upstream pegleg code umask is changed (https://github.com/airshipit/pegleg/commit/a8620cfd8d1fd32db546c3e4abbb26a215369649) , that's why generated output file has different permissions with newer image. We did not see this issue with leap 15.0 as that image was built earlier and we will see same issue if image is built now. 
So seems like a valid change on our side to make it readable for others (shipyard container user)

PR #628, opened 2019-07-16
 Use SLE15SP1 for the deployer

Itxaka:
Removes 2 lines from the velum Gemfile.lock for bundle 2.0 compatibility
  Adds an extra vendor from the devel:languages:ruby for zypper
  Drops rng-tools from the required packages as it doesnt seem to be used
  Adds the needed repos for SLES15SP1-JeOS
  Changes the openstack image used from Leap 15.0 to SLES15SP1-JeOS

### DIFF
--- a/playbooks/openstack-enroll_caasp_workers.yml
+++ b/playbooks/openstack-enroll_caasp_workers.yml
@@ -47,6 +47,20 @@
         dest: "{{ upstream_repos_clone_folder }}/ccp/"
 
 
+    # In SLE15SP1 the default bundler is 2.x while the velum Gemfile.lock has been bundled with 1.15.0 so we need to
+    # manually remove the BUNDLED WITH+1.15.0 text of the Gemfile.lock for it to install the dependencies
+    # This is only needed for CAASP3, can be removed once we move to CAASP4
+    - name: Remove "BUNDLED WITH" text from Gemfile.lock
+      lineinfile:
+        path: "{{ upstream_repos_clone_folder }}/ccp/velum-bootstrap/Gemfile.lock"
+        state: absent
+        regexp: "{{ item }}"
+      with_items:
+        - "^BUNDLED WITH$"
+        - "1.15.0$"
+
+
+
     # Ansible velum needs to get given a specific shell, bash, therefore
     # the shell module has to be used (command doesn't accept executable)
     # Skip ansible lint is therefore used, to avoid the warning that this task
--- a/playbooks/openstack-osh_hostconfig.yml
+++ b/playbooks/openstack-osh_hostconfig.yml
@@ -28,10 +28,12 @@
         value: "false"
 
     - name: Workaround stupid JeOS image bugs (bsc#1138727)
-      shell: "test -L /etc/resolv.conf || netconfig update -f"
+      file:
       args:
-        executable: /bin/sh
-        creates: /var/run/netconfig/resolv.conf
+        src: /var/run/netconfig/resolv.conf
+        dest: /etc/resolv.conf
+        state: link
+        force: yes
 
     - name: Configure host repositories
       zypper_repository:
@@ -40,6 +42,7 @@
         autorefresh: True
         auto_import_keys: yes
         state: present
+        disable_gpg_check: yes
       loop: "{{ deploy_on_openstack_osh_repos_per_imagename[deploy_on_openstack_oshnode_image] }}"
 
     - name: Add known vendors
@@ -47,7 +50,7 @@
         path: /etc/zypp/vendors.d/opensuse
         block: |
           [main]
-          vendors = suse,opensuse,obs://build.opensuse.org/Cloud:OpenStack,obs://build.opensuse.org/devel:languages:python
+          vendors = suse,opensuse,obs://build.opensuse.org/Cloud:OpenStack,obs://build.opensuse.org/devel:languages:python,obs://build.opensuse.org/Cloud,obs://build.opensuse.org/devel:languages:ruby
         create: yes
 
     - name: Zypper dist-upgrade
--- a/playbooks/roles/handle-nodes-on-openstack/tasks/create_on_openstack.yml
+++ b/playbooks/roles/handle-nodes-on-openstack/tasks/create_on_openstack.yml
@@ -37,6 +37,7 @@
         hosts:
           {{ servername }}:
             ansible_host: {{ floating_ip }}
+            {% if python_interpreter|default(False) %}ansible_python_interpreter: {{ python_interpreter }}{% endif %}
 
 - meta: refresh_inventory
 
--- a/vars/common-vars.yml
+++ b/vars/common-vars.yml
@@ -31,7 +31,18 @@ socok8s_repos_to_configure:
   openSUSE-Leap-15.0-Update: "{{ obs_mirror }}/update/leap/15.0/oss/"
   openSUSE-Leap-15.0-Cloud: "{{ obs_mirror }}/repositories/Cloud:/Tools/openSUSE_Leap_15.0/"
   Leap15develtools: "{{ obs_mirror }}/repositories/devel:/tools/openSUSE_Leap_15.0/"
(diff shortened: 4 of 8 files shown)


Itxaka:
Wont work because gcc doesnt seem to be available in any of the 30 repos we add ....somehow....

playbooks/openstack-enroll_caasp_workers.yml
```diff
@@ -46,6 +46,19 @@
         dest: "{{ upstream_repos_clone_folder }}/ccp/"
 
 
+    # In SLE15SP1 the default bundler is 2.x while the velum Gemfile.lock has been bundled with 1.15.0 so we need to
```
jsuchome: Could you please add a comment that this change is only needed for CaaSP3 and can be removed once we switch to CaaSP4?

Itxaka: Good point!

playbooks/openstack-osh_hostconfig.yml
```diff
@@ -28,10 +30,12 @@
         value: "false"
 
     - name: Workaround stupid JeOS image bugs (bsc#1138727)
```
jsuchome: BTW, wasn't this problem already fixed in latest JeOS image?

vars/common-vars.yml
```diff
@@ -31,7 +31,21 @@ socok8s_repos_to_configure:
...
+  SLE-15-SP1-JeOS-GM-backports: "{{ obs_mirror }}/repositories/openSUSE:/Backports:/SLE-15-SP1/standard/"
+  # not sure why its needed?
+  SLE-15-SP1-JeOS-GM-Cloud:  "{{ obs_mirror }}/repositories/Cloud:/Tools/SLE_15_SP1/"
+  # needed for phantomjs
+  SLE-15-SP1-JeOS-GM-develtools: "{{ ibs_mirror }}/ibs/SUSE/Products/SLE-Module-Development-Tools/15-SP1/x86_64/product/"
+  # needed for ruby2.5-devel
+  SLE-15-SP1-JeOS-GM-ruby: "https://download.opensuse.org/repositories/devel:/languages:/ruby/SLE_15_SP1/"
+  # needed for ruby2.5-rubygem-bundler
+  SLE-15-SP1-JeOS-GM-rubyextensions: "{{ obs_mirror}}/repositories/devel:languages:ruby:extensions/SLE_15_SP1/"
+  # This is where the repo for gcc would go IF I could find ONE!
```
jsuchome: Should be in develtools repo actually

playbooks/openstack-osh_hostconfig.yml
Itxaka: Yes, but it seems that it wasnt correctly fixed, new fix is from https://github.com/SUSE-Cloud/socok8s/pull/600

vars/common-vars.yml
Itxaka: :+1:

Itxaka:
current missing packages:

```
    <message type="error">No provider of make found.</message>
    <message type="error">No provider of libxml2-devel found.</message>
    <message type="error">No provider of which found.</message>
    <message type="error">No provider of libxslt-devel found.</message>
    <message type="error">No provider of jq found.</message>
    <message type="error">No provider of phantomjs found.</message>
```

vars/common-vars.yml
```diff
@@ -31,7 +31,18 @@ socok8s_repos_to_configure:
   openSUSE-Leap-15.0-Update: "{{ obs_mirror }}/update/leap/15.0/oss/"
   openSUSE-Leap-15.0-Cloud: "{{ obs_mirror }}/repositories/Cloud:/Tools/openSUSE_Leap_15.0/"
   Leap15develtools: "{{ obs_mirror }}/repositories/devel:/tools/openSUSE_Leap_15.0/"
+  SLE-15-SP1-JeOS-GM-Full: http://download.suse.de/full/full-sle15-sp1-x86_64/
```
AbelNavarro: `download.suse.de` only works inside our VPN. Is that the expected behavior?

Itxaka: yep. For the moment we dont support using SCC for updates so this will have to work until we iterate over that and enable it

Itxaka:
somehow ses is picking up the var and setting its interpreter to python3, debugging for now...

vars/common-vars.yml
jsuchome: This repo is not ideal, but if we make it work, we at least know it's possible. And in another PR's we can work on better setup.

playbooks/openstack-osh_hostconfig.yml
```diff
@@ -40,14 +42,15 @@
         autorefresh: True
         auto_import_keys: yes
         state: present
+        disable_gpg_check: yes
```
ilausuch: Why is this disabled?

Itxaka: Because the full image repo for SLE15-SP1 is not signed
No idea why, this should be temporary as we should move to use SCC for adding SLE repos instead of manually adding them

Itxaka:
forcing the repos to point to https://download.opensuse.org becuase the provo mirror doesnt seem to be properly synced so we get failures on adding repos (ruby and rubyextensions for example)

Itxaka:
should be green now

Itxaka:
Fixed airship adding the socok8s repo to point to the proper one

playbooks/openstack-osh_hostconfig.yml
evrardjp: Could you please add a comment there, to explain why we are doing this?

Something like
```#TODO(itxaka): Remove `disable_gpg_check` when SLE15SP1 image is signed```

File: playbooks/roles/handle-nodes-on-openstack/tasks/create_on_openstack.yml [hunk 6]
```diff
@@ -37,6 +37,7 @@
         hosts:
           {{ servername }}:
             ansible_host: {{ floating_ip }}
+            {% if python_interpreter|default(False) %}ansible_python_interpreter: {{ python_interpreter }}{% endif %}
```
evrardjp: Do we need that? Can we make that less conditional?

vars/common-vars.yml
```diff
@@ -31,7 +31,18 @@ socok8s_repos_to_configure:
...
+  SLE-15-SP1-JeOS-GM-Update: http://download.suse.de/ibs/SUSE:/SLE-15-SP1:/Update/standard/
+  # needed for python-barbicanclient
+  SLE-15-SP1-JeOS-GM-backports: "https://download.opensuse.org/repositories/openSUSE:/Backports:/SLE-15-SP1/standard/"
+  # needed for ruby2.5-devel
+  SLE-15-SP1-JeOS-GM-ruby: "https://download.opensuse.org/repositories/devel:/languages:/ruby/SLE_15_SP1/"
+  # needed for ruby2.5-rubygem-bundler
+  SLE-15-SP1-JeOS-GM-rubyextensions: "https://download.opensuse.org/repositories/devel:languages:ruby:extensions/SLE_15_SP1/"
+  # phantomjs
+  SLE-15-SP1-JeOS-GM-openstack-master: "https://download.opensuse.org/repositories/Cloud:/OpenStack:/Master/SLE_15_SP1/"
```
dirkmueller: ugh, that doesn't sound like a stable solution..

evrardjp: Agreed.

vars/common-vars.yml
dirkmueller: we really should use download.suse.de or download.o.org in the socok8s jobs but use the provo local mirrors (as indicated by {{ obs_mirror }} for example

jsuchome: AFAIK this was really a last resort when we were not able to find out all the right repositories for SP1 packages

jsuchome:
This should be obsoleted by https://github.com/SUSE-Cloud/socok8s/pull/637

evrardjp:
Correct. Let's close.

dirkmueller:
closing, the patches from itxaka are in the other PR..

PR #629, opened 2019-07-16
Make airship control plane node count configurable

jesusaurus:
Add an ansible variable for the number of caasp workers to use as
airship control plane workers instead of hard-coding to 2.

### DIFF
--- a/playbooks/openstack-caasp_instance.yml
+++ b/playbooks/openstack-caasp_instance.yml
@@ -119,8 +119,8 @@
               {% endfor %}
 
               airship-openstack-control-workers:
-                hosts: &firsttwoworkers
-              {% for node in workers %}{% if loop.index <= 2 %}
+                hosts: &first_workers
+              {% for node in workers %}{% if loop.index <= airship_ucp_control_workers %}
                   {{ node.Name }}:
                     ansible_host: {{ node.Networks.split(',')[-1].replace(' ','') }}
                     ansible_user: root
@@ -131,12 +131,12 @@
               {% endfor %}
 
               airship-ucp-workers:
-                hosts: *firsttwoworkers
+                hosts: *first_workers
 
               airship-openstack-compute-workers:
                 hosts:
               {% for node in workers %}
-              {% if loop.index == 3 %}
+              {% if loop.index > airship_ucp_control_workers %}
                   {{ node.Name }}:
                     ansible_host: {{ node.Networks.split(',')[-1].replace(' ','') }}
                     ansible_user: root
@@ -147,7 +147,7 @@
               {% endfor %}
 
               airship-kube-system-workers:
-                hosts: *firsttwoworkers
+                hosts: *first_workers
 
         - meta: refresh_inventory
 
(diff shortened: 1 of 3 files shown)


jsuchome:
1. Could you also adapt playbooks/roles/caasp4/templates/inventory-caasp4.yml.j2 ?
2. What if we are not deploying on openstack?

jesusaurus:
@jsuchome
1. done
2. I'm not very familiar with that use case, I thought inventory files were manually created in that case. I've renamed the variable and moved it to common-vars

vars/common-vars.yml
```diff
@@ -125,3 +125,5 @@ scale_profile: "{{ lookup('env','scale_profile') | default('minimal', true) }}"
 
 #Set timeout for helm cleanup.
 helm_delete_timeout: 900
+
+airship_ucp_workers: 2
```
jsuchome: The meaning is actually "control workers", right? As opposed to "compute workers", if I get it right. So it is a subset of "caasp workers" and current variable name does not make it clear IMHO

playbooks/openstack-caasp_instance.yml
```diff
@@ -116,8 +116,8 @@
               {% endfor %}
 
               airship-openstack-control-workers:
-                hosts: &firsttwoworkers
-              {% for node in workers %}{% if loop.index <= 2 %}
+                hosts: &first_workers
```
jsuchome: What is actually the meaning of this? I do not understand the syntax...

jesusaurus: This is a yaml "anchor" 
https://confluence.atlassian.com/bitbucket/yaml-anchors-960154027.html

jsuchome: Thanks!

vars/common-vars.yml
```diff
@@ -143,6 +143,10 @@ scale_profile: "{{ lookup('env','scale_profile') | default('minimal', true) }}"
 #Set timeout for helm cleanup.
 helm_delete_timeout: 900
 
+# Number of caasp workers to use for the control plane,
+# the remainder will be used as computer hosts
+airship_ucp_control_workers: 2
```
toabctl: Why common-vars? This is airship specific...

playbooks/roles/caasp4/templates/inventory-caasp4.yml.j2
```diff
@@ -32,15 +32,15 @@ airship-openstack-control-workers:
...
+  hosts: *first_workers
   vars:
     ansible_user: sles
     ansible_python_interpreter: /usr/bin/python3
 
 airship-openstack-compute-workers:
   hosts:
 {% for worker_ip in _terraform_json_output.ip_workers.value %}
-{% if loop.index0 > 1 %}
+{% if loop.index0 >= airship_ucp_control_workers %}
```
toabctl: urgs. so we (again) include $random variables from everywhere in specific roles. But not your fault..

vars/common-vars.yml
jesusaurus: I put it under the airship-only section of common-vars https://github.com/SUSE-Cloud/socok8s/blob/master/vars/common-vars.yml#L56; we can move that entire section to a new file, but I believe that is a separate task from this PR.

PR #630, opened 2019-07-17
Support for CaaSP4 setup, cont. (Support CaaSPv4 in socok8s SOC-9926)

jsuchome:
Various fixes for CaaSP4 deployment

Deployment tested tested with skuba-0.7.1 package. I suggest to use SLE15SP1 node as "localhost".

Also, not that we need newer kubectl package on the deployer node. SLE15SP1 support for deployer should hopefully solve it


playbooks/roles/caasp4/templates/inventory-caasp4.yml.j2
```diff
@@ -2,33 +2,55 @@
 caasp-masters:
   hosts:
 {% for master_ip in _terraform_json_output.ip_masters.value %}
-    {{ master_ip }}
+    caasp-master-{{ loop.index }}:
+      ansible_host: {{ master_ip }}
 {% endfor %}
+  vars:
+    ansible_user: sles
```
evrardjp: this look very image dependent, but I guess we are only carying about the CI use cases for now, so we can improve later.

I expect we could do a ci inventory, whose content contains CI only vars.

jsuchome:
> I am pretty confused, why do we have to carry all those libvirt files?

The files are currently present in skuba repo, and I'm copying them to ours (because when I started this PR, skuba was still private repository)
Once they get packaged (as the openstack files are) we should use the packaged ones.

evrardjp:
> > I am pretty confused, why do we have to carry all those libvirt files?
> 
> The files are currently present in skuba repo, and I'm copying them to ours (because when I started this PR, skuba was still private repository)
> Once they get packaged (as the openstack files are) we should use the packaged ones. At least that would be my preferred solution. Or we can actually use skuba as submodule for now.

How long would it take to help them package those? I think it would be good to avoid the submodule, but it's better to carry a submodule than a fork :)

jsuchome:
This is what I got from CaaSP developer: _we currently don't really test it with libvirt, i'm not sure in which state it is at the moment. and libvirt is very low on the prioritized list of the officially supported platforms_

jsuchome:
@evrardjp so actually, I removed those terraform files again. CaaSP team does not seem to care much about libvirt and openstack ones we take from skuba package

evrardjp:
> This lgtm but I haven't tested it.
> 
> After CaaSPv4 is released we should remove all of the v3 codepaths and pay down a lot of this technical debt.

That's indeed the goal.

PR #634, opened 2019-07-18
Airship: Removed not used rm privileged-cluster-role-binding.yml

ilausuch:
In role airship-configure-caasp the task privileged-cluster-role-binding.yml is already integrated in the main.yml task.


ilausuch:
It is a very small task, In my opinion, dividing it in two tasks doesn't make this more clear

PR #635, opened 2019-07-19
Sets character restrictions on SOCOK8S_ENVNAME

scottwulf:
Disallows SOCOK8S_ENVNAME from containing characters which cause
issues with this project's scripts and playbooks


scottwulf:
Oops, I somehow dismissed "stale" reviews.

PR #637, opened 2019-07-23
Switch to SLE15SP1 for deployer node (SOC-9680)

dirkmueller:
Based on #628 however including the official repositories instead.

### DIFF
--- a/vars/deploy-on-openstack.yml
+++ b/vars/deploy-on-openstack.yml
@@ -30,6 +30,14 @@ deploy_on_openstack_osh_repos_per_imagename:
     - openSUSE-Leap-15.0-Cloud
     - Leap15develtools
     - socok8s
+  SLE-15-SP1-JeOS-GM:
+    - SLE-15-SP1-JeOS-GM-Full
+    - SLE-15-SP1-JeOS-GM-Update
+    - SLE-15-SP1-JeOS-GM-backports
+    - SLE-15-SP1-JeOS-GM-openstack-master
+    - SLE-15-SP1-JeOS-GM-ruby
+    - SLE-15-SP1-JeOS-GM-rubyextensions
+    - socok8s_SLE15SP1
 
 deploy_on_openstack_caasp_repos_per_imagename:
   caasp-3.0.0-GM-OpenStack-qcow:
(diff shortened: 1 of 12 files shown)


vars/deploy-on-openstack.yml
```diff
@@ -30,6 +30,14 @@ deploy_on_openstack_osh_repos_per_imagename:
     - openSUSE-Leap-15.0-Cloud
     - Leap15develtools
     - socok8s
+  SLE-15-SP1-JeOS-GM:
```
evrardjp: Could you add a ```#TODO: remove SLES-15-SP1-openstack-master when caasp3 compat is removed```, please?

dirkmueller: @evrardjp I don't see it in the list of repositories, why should I add a comment about a repository that isn't used? ;) I guess I should squad the commits at some point..

evrardjp: That was just before you pushed a patch I think or because I am blind. Just ignore this. :)

evrardjp:
Approved the concept, waiting for ci.

jsuchome:
`Problem: nothing provides libcrypto.so.1.0.0()(64bit) needed by phantomjs-2.1.1.0-1.1.x86_64`

evrardjp:
> `Problem: nothing provides libcrypto.so.1.0.0()(64bit) needed by phantomjs-2.1.1.0-1.1.x86_64`

phantomjs will go away in caasp4, it makes me very sad to have to fix this just before ripping this out.

dirkmueller:
@evrardjp legacy module, added now.

evrardjp:
That looks better. Respinning CI in my own PR.

evrardjp:
Looks okay to me, if it passes CI.

evrardjp:
Should we remove python3-cmd2? Not sure it brings something here.

jsuchome:
`Package '+python-pyOpenSSL' not found`

evrardjp:
This CI job was stuck for an hour and a half without doing anything, so I stopped it.

evrardjp:
Ok the restart seems better, I can see logs now. Weird.

PR #638, opened 2019-07-23
Remove references to inventory_hostname


evrardjp:
Airship role is using inventory_hostname for hostnames.

This is a problem, as it represents the inventory
name, not the real node name. Kubernetes is using node name.

This fixes it by removing all the references in the airship side
of inventory_hostname to the ansible_hostname, which will be
gathered on the hosts by ansible, during the fact gathering step.

### DIFF
--- a/playbooks/roles/airship-configure-worker/tasks/main.yml
+++ b/playbooks/roles/airship-configure-worker/tasks/main.yml
@@ -7,7 +7,7 @@
     - add_compute_node
 
 - name: Enroll node for Airship kube-system ingress
-  shell: "kubectl label nodes {{ inventory_hostname }} kube-ingress=enabled"
+  command: "kubectl label nodes {{ ansible_hostname }} kube-ingress=enabled"
   when: ('airship-kube-system-workers' in group_names)
   register: _enrollforucp
   changed_when:
@@ -19,7 +19,7 @@
     - skip_ansible_lint
 
 - name: Enroll node for Airship UCP
-  shell: "kubectl label nodes {{ inventory_hostname }} ucp-control-plane=enabled"
+  command: "kubectl label nodes {{ ansible_hostname }} ucp-control-plane=enabled"
   when: ('airship-ucp-workers' in group_names)
   register: _enrollforucp
   changed_when:
@@ -31,9 +31,7 @@
     - skip_ansible_lint
 
 - name: Enroll node for Airship Openstack control plane
-  shell: |
-    kubectl label nodes {{ inventory_hostname }} openstack-control-plane=enabled
-    kubectl label nodes {{ inventory_hostname }} openvswitch=enabled
+  command: "kubectl label nodes {{ ansible_hostname }} openstack-control-plane=enabled openvswitch=enabled"
   when: ('airship-openstack-control-workers' in group_names)
   register: _enrollforopenstackcp
   changed_when:
@@ -45,8 +43,8 @@
     - skip_ansible_lint
 
 - name: Enroll node for Airship Openstack L3 agent
-  shell: |
-    kubectl label nodes {{ inventory_hostname }} openstack-l3-agent=enabled
+  command: |
+    kubectl label nodes {{ ansible_hostname }} openstack-l3-agent=enabled
   when: ('airship-openstack-l3-agent-workers' in group_names)
   register: _enrollforopenstackl3agent
   changed_when:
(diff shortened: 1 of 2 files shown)


evrardjp:
> CI is failing. The command module does not accept multiple commands like the shell module does. Instead you need to use with_items, for example:
> 
>     * name: command v shell
>       command: "{{ item }}"
>       register: c1
>       with_items:
>       
>       * ls /home
>       * ls /root
> 
> 
> This change modifies the value of the registered value (it is no longer an aggregated value) and the concern then becomes whether the code checking the registered variable still behaves as intended. Please check that. Thanks.

FYI ```with_items``` is deprecated (and so is every with_ construct). ```loop:``` should be used instead.
Anyway, this whole series of tasks need to be refactor with a loop, to avoid this long series of tasks, which is hard to read and doesn't bring a lot of value. I am not here to refactor this work, but make CaaSP 4 deployable.

We don't need looping though, as what we are looking for already works in kubectl.

playbooks/roles/airship-configure-worker/tasks/main.yml
```diff
@@ -58,9 +56,9 @@
     - skip_ansible_lint
 
 - name: Enroll node for Airship Openstack compute plane
-  shell: |
-    kubectl label nodes {{ inventory_hostname }} openstack-compute-node=enabled
-    kubectl label nodes {{ inventory_hostname }} openvswitch=enabled
+  command: |
```
scottwulf: Can you merge the calls here like line 34 to fix break?

evrardjp: Good catch!

evrardjp:
The symlink doesn't seem to work.

playbooks/roles/airship-deploy-osh/tasks/main.yml
```diff
@@ -69,10 +69,11 @@
     - skip_ansible_lint
     - update_airship_osh_site
 
-- name: Copy site to Shipyard
+- name: Link shipyard to expected location
```
scottwulf: CI is failing with "refusing to convert from directory to symlink for /opt/airship/shipyard". I am not familiar with the situation or the code's intention but maybe a "when: "{{ treasuremap_dir }}/{{ pegleg_output }} != {{ upstream_repos_clone_folder }}/airship/shipyard"" condition is a sufficient fix.

evrardjp: I am not too familiar either. I will convert this to a recursive copy instead, until this is clarified.

kantarun: Its better to copy this file in shipyard folder as this file is made available to shipyard CLI container via mount. And we need read permissions for others to be accessible in the container. See PR #627 . Not sure how symlink is going to play out in all this.

playbooks/roles/airship-deploy-osh/tasks/main.yml
```diff
@@ -69,10 +69,11 @@
     - skip_ansible_lint
     - update_airship_osh_site
 
-- name: Copy site to Shipyard
+- name: Copy shipyard to expected location
```
kantarun: Technically we are copying aggregated and partially validated site document to shipyard for publishing to deckhand.
So previous text is more closer.

playbooks/roles/airship-deploy-osh/tasks/main.yml
```diff
@@ -71,8 +71,9 @@
 
 - name: Copy site to Shipyard
   become: yes
-  delegate_to: '{{ inventory_hostname }}'
-  synchronize:
+  copy:
+    remote_src: yes
+    mode: "preserve"
```
Itxaka: [2019-07-29T09:45:33.816Z]   msg: 'Remote copy does not support recursive copy of directory: /opt/airship/treasuremap/peggles'

Itxaka: maybe due to  `remote_src supports recursive copying as of version 2.8.` didnt we had 2.8 already in the CI?

Itxaka: no we dont, 2.7.8 still on CI:

```
ccp-ci-launcher:~ # ansible --version
ansible 2.7.8
```

evrardjp:
Moved back to symlink, now in "force" mode. Let's see how it goes.

evrardjp:
> Moved back to symlink (Arun told me to test it), now in "force" mode. Let's see how it goes.

Not working, reverting to a copy, this time in shell. Tired of this.

evrardjp:
500: Server Error for url: https://engcloud.prv.suse.net:8774

evrardjp:
Seems to pass now.

PR #639, opened 2019-07-23
More prominently documents the insecure flag

scottwulf:
I happened to skim over the existing, quick mention of the insecure
flag and spent more time than necessary attempting to resolve
SSL certificate verification failures. This fix more prominently
documents the insecure flag to save users' time.

### DIFF
--- a/doc/source/suse-ecp/ose-localhost.rst
+++ b/doc/source/suse-ecp/ose-localhost.rst
@@ -185,6 +185,7 @@ The following is an example if you are running on an "engcloud":
    clouds:
      engcloud:
        region_name: CustomRegion
+       insecure: False
        auth:
          auth_url: https://keystone_url/v3
          username: john # your username here


doc/source/suse-ecp/ose-localhost.rst
```diff
@@ -185,6 +185,7 @@ The following is an example if you are running on an "engcloud":
    clouds:
      engcloud:
        region_name: CustomRegion
+       insecure: False
```
evrardjp: Did you mean True here? Refering to your commit message where you had to go through validation errors... (it's not clear what the cause is, by looking at the commit message).

scottwulf: No, I meant False. False is the default, and more secure, value. I just wanted to add the insecure flag to the example config file in case others, like I, miss the other mention of said flag.

kantarun: So in ECP, this works with insecure: False OR insecure: True 
Looks like from your earlier discussions, you may end up using insecure: True in your environment. What was different and will it be helpful to know (if it seems generic enough) to have that check if this flag needs to be changed in one's environment or not ?

scottwulf: If you run into SSL issues, worse case solution would be to change it to True (and leave yourself vulnerable to man-in-the-middle attacks).

Yes, early validation of the SSL certificate would be a nice-to-have so I have created SOC-9982.

PR #640, opened 2019-07-23
[DNM] Test CaaSP4 with SLE15SP1 deployer


evrardjp:
See commit message.


evrardjp:
This can be abandonned nowadays. It lived it usefulness.

PR #643, opened 2019-07-25
Remove automation submodule

dirkmueller:
Not sure we still use it


.gitmodules
```diff
@@ -1,6 +1,3 @@
-[submodule "submodules/SUSE-Cloud/automation"]
-	path = submodules/SUSE-Cloud/automation
-	url = https://github.com/SUSE-Cloud/automation.git
```
kantarun: I see SES integration still uses this module 
https://github.com/SUSE-Cloud/socok8s/blob/master/playbooks/roles/ses

jsuchome: Yeah, we use it for CaaSPv3, so it's still not possoble to remove it

PR #644, opened 2019-07-25
Adding tempest usage for existing integration job

kantarun:
Adding additional tempest test run step when related flag is set.
Jobs are supposed to set the flag and tempest test type (full or smoke)
as input parameters.

In case of 'openstack' deployment mechanism (deploying in ECP cloud),
tempest network values are derived from variables in deploy-on-opnestack.
For 'kvm' deployment mechanism (deploying on baremetal or non-ECP
cloud), user is expected to provide 'openstack_project_network_cidr'
which is an existing behavior for running tempest test step.

Assuming CIDR value is defined and correct, playbooks will create
network and subnet needed for tempest tests execution as part of tempest
run step.

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -24,6 +24,8 @@ pipeline {
         /* first value in the list is the default */
         choice(choices: ['airship', 'osh'], description: 'Which deployment mechanism?', name: 'deployment')
         choice(choices: ['caasp3', 'caasp4'], description: 'Which caasp version to use?', name: 'caasp_version')
+        booleanParam(name: 'run_tempest', defaultValue: false, description: 'Run tempest tests?')
+        choice(choices: ['smoke', 'all'], description: 'Run smoke or all tempest tests?', name: 'tempest_test_type')
     }
 
     environment {
@@ -206,6 +208,26 @@ pipeline {
             }
             steps {
                 sh "source ${WORKSPACE}/sock_${env.SOCOK8S_ENVNAME} && ./run.sh setup_airship"
+            }
+        }
+
+        stage('Run Tempest') {
+            options {
+                timeout(time: 240, unit: 'MINUTES', activity: true)
+            }
+            when {
+                allOf {
+                    expression { env.run_tempest == 'true' }
+                    expression { params.deployment == 'airship' }
+                    expression { return  TestFunctional }
+                }
+            }
+            steps {
+                sh """
+                  export TEMPEST_TEST_TYPE=${env.tempest_test_type}
+                  export AIRSHIP_TEMPEST_LOG_STDOUT=True
+                  ./run.sh test
+                """
             }
         }
     }
--- a/doc/source/deployment/verify.rst
+++ b/doc/source/deployment/verify.rst
@@ -23,43 +23,68 @@ OpenStack Tempest Testing
 
 After the deployment of SUSE Containerized OpenStack has completed, it is possible to run
 OpenStack Tempest tests against the core services in the deployment using the `run.sh` script.
-Before running Tempest tests, it will be necessary to manually configure OpenStack network
-resources and provide a few configuration parameters in the ${WORKDIR}/env/extravars file.
+As part of Tempest tests execution, there is need to configure OpenStack network resources
+and provide a few configuration parameters in the `${WORKDIR}/env/extravars file`.
 
-Setting Up An External Network And Subnet in OpenStack
-------------------------------------------------------
+Setting Up Network CIDR(s) in OpenStack
+----------------------------------------
 
-To set up an external network and subnet in OpenStack, the following commands can be run from a
-shell on the `Deployer` node.
+Tempest will create a private network (10.0.0.0/8) to use as the default network, and it will
+need to know the CIDR block from which to allocate project IPv4 subnets. This value should be
+specified with the following key in the `${WORKDIR}/env/extravars` file:
+
+.. code-block:: yaml
+
+   openstack_project_network_cidr: "10.0.4.0/24"
+
+By default, same CIDR block is used for creating external network which is needed by tempest
+tests execution. If that needs to be different in your environment, provide value with
+following key in the `${WORKDIR}/env/extravars` file:
+
+.. code-block:: yaml
+
+   openstack_external_network_cidr: "192.168.100.0/24"
+
+Tempest test execution via `run.sh` script will create the necessary networks as long as
+above CIDR value is correctly identified and specified in extravars.
+
+.. note::
+
+   In **openstack** deployment mechanism, defaults are used from `vars/deploy-on-openstack.yml`
+   file. So setup of above CIDR may not be needed if those defaults can be used for tempest
+   test execution.
+
+Optional External Network Setup
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If for some reason, you prefer to create your own external network, the following commands can
+be run from a shell on the `Deployer` node.
 
 .. code-block:: console
 
    export OS_CLOUD=openstack
    openstack network create --provider-network-type flat --provider-physical-network external \
-     --external public
-   openstack subnet create --network public --subnet-range 192.168.100.0/24 --allocation-pool \
-     start=192.168.100.10,end=192.168.100.200 --gateway 192.168.100.1 --no-dhcp public-subnet
+     --external socok8s-tempest-public
+   openstack subnet create --network socok8s-tempest-public --subnet-range 192.168.100.0/24 \
+   --allocation-pool start=192.168.100.10,end=192.168.100.200 --gateway 192.168.100.1 \
+   --no-dhcp socok8s-tempest-public-subnet
+
+If the external network and subnet have been created in OpenStack manually as mentioned above,
+their names will need to be made known to Tempest by adding the following keys in the
+`${WORKDIR}/env/extravars` file:
+
+.. code-block:: yaml
+
+   openstack_external_network_name: "socok8s-tempest-public"
+   openstack_external_subnet_name: "socok8s-tempest-public-subnet"
+
 
 .. note::
 
    The external public network is expected to be able to reach the internet. The above values
    will vary based on your network environment.
 
-Once the public network and subnet have been created in OpenStack, their names will need to be
-made known to Tempest by adding the following keys in the ${WORKDIR}/env/extravars file:
 
-.. code-block:: yaml
-
-   openstack_external_network_name: "public"
-   openstack_external_subnet_name: "public-subnet"
-
-Tempest will create a private network (10.0.0.0/8) to use as the default network, and it will
-need to know the CIDR block from which to allocate project IPv4 subnets. This value should be
-specified with the following key in the extravars file:
-
-.. code-block:: yaml
-
-   openstack_project_network_cidr: "10.0.4.0/24"
 
 Configuring Tempest Test Parameters
 -----------------------------------
@@ -148,8 +173,10 @@ are executed.
 Tempest Test Results
 --------------------
 
-All test results can be viewed by retrieving the logs from the airship-tempest-run-tests pod by running
-the following command:
+By default, tempest test execution pod logs are displayed on ansible stdout during `test` option.
+
+Later, all test results can be viewed by retrieving the logs from the airship-tempest-run-tests pod by
+running the following command:
 
 .. code-block:: console
 
--- a/playbooks/roles/airship-deploy-tempest/tasks/prepare-tempest.yml
+++ b/playbooks/roles/airship-deploy-tempest/tasks/prepare-tempest.yml
@@ -1,0 +1,85 @@
+---
+- name: Get deployment mechanism value
+  set_fact:
+    deployment_mechanism: "{{ lookup('env', 'DEPLOYMENT_MECHANISM') }}"
+  tags:
+    - always
+
+- name: Load openstack mechanism variables
+  include_vars: "{{ playbook_dir }}/../vars/deploy-on-openstack.yml"
+  when:
+    - deployment_mechanism == 'openstack'
+  tags:
+    - always
+
+- name: Get tempest network cidr and test type for its execution
+  set_fact:
+    openstack_project_network_cidr: "{{ deploy_on_openstack_internal_subnet_cidr }}"
+    # env variable is used to pass test type from CI. If that is not set, then use value via extravars
+    # variable or ansible variable default value
+    tempest_test_type: "{{ lookup('env', 'TEMPEST_TEST_TYPE') | default(tempest_test_type, true) }}"
+  when:
+    - deployment_mechanism == 'openstack'
+
+- name: Print tempest project and external network cidr value
+  debug:
+    msg:
+    - "openstack_project_network_cidr = {{ openstack_project_network_cidr }}"
+    - "openstack_external_network_cidr = {{ openstack_external_network_cidr }}"
+
+- name: Make sure that tempest project network cidr is present and correctly defined
+  assert:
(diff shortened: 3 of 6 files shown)


jhesketh:
So if I'm reading this correctly, this is deploying tempest, but not actually running it?

jgu17:
> So if I'm reading this correctly, this is deploying tempest, but not actually running it?

The invocation is in the run.sh.  The tempest runs when the tempest pod is deployed.

playbooks/roles/airship-deploy-tempest/tasks/prepare-tempest.yml
```diff
@@ -0,0 +1,68 @@
+---
+# tasks file for airship-deploy-tempest
+- name: Load generic and openstack mechanism variables
+  include_vars: "{{ item }}"
+  with_items:
+    - "{{ playbook_dir }}/../vars/common-vars.yml"
+    - "{{ playbook_dir }}/../vars/deploy-on-openstack.yml"
```
jgu17: should that only be included when the deploy mechanism is openstack?

jgu17:
Does this PR now create the ext network and subnet if they were not found for both deployment mechanisms? If so, we should update the doc. Nice enhancement.

playbooks/roles/airship-deploy-tempest/tasks/prepare-tempest.yml
kantarun: Agree this is useful by 'openstack' deployment mechanism. When I ran with 'kvm' mechanism, I didn't see any side-effect and deployment/ tempest execution ran without error. It may be because we don't have any corresponding kvm specific variables so there is no conflict per se and impact as well.

jgu17: agree that there is no side effect at the moment, but it could be a time tomb and easy to avoid :-).

kantarun: I have made inclusion conditional with latest update. So hopefully saved world from world war 3 :)

doc/source/deployment/verify.rst
```diff
@@ -26,40 +26,30 @@ OpenStack Tempest tests against the core services in the deployment using the `r
...
-.. note::
+.. code-block:: yaml
 
-   The external public network is expected to be able to reach the internet. The above values
-   will vary based on your network environment.
+   openstack_project_network_cidr: "10.0.4.0/24"
 
-Once the public network and subnet have been created in OpenStack, their names will need to be
-made known to Tempest by adding the following keys in the ${WORKDIR}/env/extravars file:
+By default, same CIDR block is used for creating external network which is needed by tempest
```
jgu17: I thought the code allows the end user (not ECP) to use an existing external network if the user decides to or needs to? should we explain that approach here for end users?

jhesketh:
> > So if I'm reading this correctly, this is deploying tempest, but not actually running it?
> 
> The invocation is in the run.sh. The tempest runs when the tempest pod is deployed.

Hmm okay. This wasn't obvious in the Jenkins output or by the timestamps. Would it be possible to add a step that tails the output of the run, or the summary somehow so that we have something to confirm?

doc/source/deployment/verify.rst
kantarun: Which openstack external network here you are referring to? 

Are you referring to ansible variables _openstack_external_network_name_ and _openstack_external_subnet_name_ ?

I did not mention those variables as then user need to be aware of network/subnets create command in case its created differently (from what's needed for tempest). And may be we end up describing details here as that will come up in when verifying these docs.

I do see your point as it can be useful if compatible existing network can be used. Though not sure what kind of cleanup is done once execution is completed.

Can you clarify above and where it can be useful?

jgu17: I see two scenarios when the user runs tempest. One is that the user has not network created yet (such as in ecp), so convenient/helpful for the playbook to create the network automatically and clean up after the test run is finished. The other case is that the user creates the external network by hand as it is more specific to their specific environment and likely to continue to use it beyond tempest testing. In the latter case, the tempest run will use the already created external network.

jgu17: or if the user reruns the tempest after the cloud used for a while to verify the health/trouble shooting etc., at which time the external network will be pre-existing?

kantarun:
> > > So if I'm reading this correctly, this is deploying tempest, but not actually running it?
> > 
> > 
> > The invocation is in the run.sh. The tempest runs when the tempest pod is deployed.
> 
> Hmm okay. This wasn't obvious in the Jenkins output or by the timestamps. Would it be possible to add a step that tails the output of the run, or the summary somehow so that we have something to confirm? (or was I blind and that's there already)



> > > So if I'm reading this correctly, this is deploying tempest, but not actually running it?
> > 
> > 
> > The invocation is in the run.sh. The tempest runs when the tempest pod is deployed.
> 
> Hmm okay. This wasn't obvious in the Jenkins output or by the timestamps. Would it be possible to add a step that tails the output of the run, or the summary somehow so that we have something to confirm? (or was I blind and that's there already)

There is logic to print pod logs when there is failure. But not in case of success. So let me add logs for tempest run pod after test execution is completed (regardless of success or failure)

doc/source/deployment/verify.rst
kantarun: Okay. I will add back the creation of network step and providing that information in extravars

playbooks/roles/airship-deploy-tempest/tasks/prepare-tempest.yml
kantarun: Requested change was made in earlier patches

Jenkinsfile.integration
```diff
@@ -24,6 +24,10 @@ pipeline {
     parameters {
         /* first value in the list is the default */
         choice(choices: ['airship', 'osh'], description: 'Which deployment mechanism?', name: 'deployment')
+
+        booleanParam(name: 'run_tempest', defaultValue: true, description: 'Run tempest tests?')
+
+        choice(choices: ['full', 'smoke'], description: 'Run full or smoke tempest tests?', name: 'tempest_test_type')
```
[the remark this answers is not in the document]
kantarun: Its added as an option as currently tempest tests execution adds to already a lengthy airship deployment which in turn limits the frequency of running this job. Also a very long job execution adds to infra resource constraints as a single job will take longer time. 
As per my previous smoke tests execution, it adds around 1 hour and full tests execution adds somewhere between 2-3 hours in local deployment. 
I understand that this is not a desirable behavior and we need to improve it. 
I was thinking that this way we can run jobs with tempest less frequently but on periodic basis so that we have visibility of  issues. And in parallel, we can also work on improving airship deployment and tempest test execution time frames.

Jenkinsfile.integration
```diff
@@ -181,6 +185,25 @@ pipeline {
...
         }
+
+        stage('Run Tempest') {
+            options {
+                timeout(time: 180, unit: 'MINUTES', activity: true)
+            }
+            when {
+                allOf {
+                    expression { env.run_tempest == 'true' }
+                    expression { params.deployment == 'airship' }
```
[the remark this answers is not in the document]
kantarun: Please see earlier for response for first part of review.

"(BTW the docs test boolean is not there)" 
Can you clarify this bit ? 

Are you talking about boolean check should not be using 'true' . Based on some examples, I have seen in automation repo, many places its compared as 'true' so followed the same pattern here.

Jenkinsfile.integration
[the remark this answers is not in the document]
kantarun: For other part of having blacklist file, its there for tempest execution but possibly, tempest play logic can be enhanced to be provided as external input file. I will think that will be something needs to be done in separate review when needed.

https://github.com/SUSE-Cloud/socok8s/blob/master/playbooks/roles/airship-deploy-tempest/files/tempest_blacklist

Usage: https://github.com/SUSE-Cloud/socok8s/blob/master/site/soc/software/charts/osh/openstack-tempest/tempest.yaml#L37

I am not sure about exposing this part in tempest job as configurable parameter. May be this can be revisited later when we have better idea about different scenarios and support in tempest charts.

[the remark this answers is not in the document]
evrardjp: > Its added as an option as currently tempest tests execution adds to already a lengthy airship deployment which in turn limits the frequency of running this job. Also a very long job execution adds to infra resource constraints as a single job will take longer time.
> As per my previous smoke tests execution, it adds around 1 hour and full tests execution adds somewhere between 2-3 hours in local deployment.
> I understand that this is not a desirable behavior and we need to improve it.
> I was thinking that this way we can run jobs with tempest less frequently but on periodic basis so that we have visibility of issues. And in parallel, we can also work on improving airship deployment and tempest test execution time frames.

There are two things we need to improve here: quality and effiency.

For me, adding one hour for a tempest smoke test is ... insane. That shouldn't be the case.
You're right, that's too long.

This highlights issues. We may have a problem of performance, which would still apply to customers, so we need to fix this.

In the meantime, what I am proposing helps you. It basically allows a file whitelisting what to do. In some specific testing, we could edit the file to increase the coverage (like a scenario for refstack, adding all the tests necessary).
In other cases, devs will just add their test into the list, and improve coverage over time. We can then decide in the review if this should be in a specific scenario or not.

That sounds like a win win to me: Very limited scope, with ability to extend.

So I see no reason to allow the skip of testing.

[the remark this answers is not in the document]
jgu17: hesitating to add to already long conversation here. but we have both smoke test and full tempest here. so maybe smoke test for the gate, full test for the daily?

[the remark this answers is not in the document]
kantarun: I see there are multiple aspects to this which needs to be figured out. 
1) Currently socok8s don't have mechanism to limit (include) certain set of tests via **external** input file. This might be relatively easy change but we may need to make them in upstream osh charts first.
2) Then there is task of identifying which set of tests are essential and how much time they take.
2 b) And then need to identify scenarios to run selectively based on service/ component change. This more seems like bigger task and may not have priority among many other things we have on our plate.

3) Then we need to improve the workflow around tempest charts and how the are invoked currently in socok8s. I think this is something we can do upstream and will provide more bang for buck here.

But currently we are not there yet and in meantime, imho having a mechanism to run tempest in CI selectively for tests visibility can be useful. And that's why this flag.

For other changes, I will suggest to create stories and address changes in separate set of PRs.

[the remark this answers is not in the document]
jhesketh: I agree with @kantarun's summary here.

At the moment we aren't exercising any tempest tests in the gate (or otherwise?), so we should focus on making that happen. The rest is optimisation that can be a followup in my opinion.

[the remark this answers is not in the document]
evrardjp: I just want to remove the ability to disable tests.

We are testing quality with linting at all times, and so should we run tempest every time we run functional testing.

Do we really want to not check what we deliver?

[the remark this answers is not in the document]
evrardjp: I think we agree then. Let's remove the conditionals, and I am good.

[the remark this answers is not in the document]
kantarun: I do agree with the universal idea that we should be running functional test with PR change. 

But in current state of things, we are not there yet. So I have more pragmatic opinion about running tempest periodically rather than running for every non-doc PR change. This adds to a already long-running job which adds to already-constrained ECP resource consumption and unreliability of gate job deployment.

That being said, I will go along with consensus here so I can keep or remove the flag based on it. 
FYI, run_tempest flag default is true and we do have job definitions in automation repo where we are planing to run tempest job every 8 hrs (it takes 4-5 hrs to run) .

evrardjp:
I am afraid if we postpone the testing, it won't be enabled until very late down the line.
Shouldn't we enable it, and improve it over time, as discussed?

jhesketh:
> I am afraid if we postpone the testing, it won't be enabled until very late down the line.
> Shouldn't we enable it, and improve it over time, as discussed?

I think what @kantarun is suggesting is that we run tempest in nightly builds rather than against every PR to save on resources. Obviously this isn't perfect and is a compromise due to CI resources.

If that goes against what we've previously agreed to or isn't acceptable, then maybe we should try enabling it for every PR and see how painful it is?

evrardjp:
> > I am afraid if we postpone the testing, it won't be enabled until very late down the line.
> > Shouldn't we enable it, and improve it over time, as discussed?
> 
> I think what @kantarun is suggesting is that we run tempest in nightly builds rather than against every PR to save on resources. Obviously this isn't perfect and is a compromise due to CI resources.
> 
> If that goes against what we've previously agreed to or isn't acceptable, then maybe we should try enabling it for every PR and see how painful it is?

Do we have cards to track the periodics running? AFAIK I don't know of anyone looking at periodics.

Jenkinsfile.integration
```diff
@@ -202,6 +204,39 @@ pipeline {
...
+            }
+            steps {
+                sh """
+                  export TEMPEST_TEST_TYPE=${env.tempest_test_type}
+                  ./run.sh test
+                """
+            }
+            post {
+                always {
+                    steps {
```
evrardjp: isn't that "script" instead of steps?

See jenkins log.

kantarun:
> > > I am afraid if we postpone the testing, it won't be enabled until very late down the line.
> > > Shouldn't we enable it, and improve it over time, as discussed?
> > 
> > 
> > I think what @kantarun is suggesting is that we run tempest in nightly builds rather than against every PR to save on resources. Obviously this isn't perfect and is a compromise due to CI resources.
> > If that goes against what we've previously agreed to or isn't acceptable, then maybe we should try enabling it for every PR and see how painful it is?
> 
> Do we have cards to track the periodics running? AFAIK no-one is responsible of taking care of periodics.

imho, all of us are responsible for checking periodic execution time to time. May be we can add email or rocketchat notification on some specific failures.
I like the idea that we first check tempest execution against periodic execution and improve there (upstream as well as local improvements) and then we can switch to PR change when team thinks its ready.

Jenkinsfile.integration
kantarun: Thanks for noticing it. Looks like got confused with wording of syntax https://jenkins.io/doc/book/pipeline/syntax/#post 

Made a change and see if it works.

Jenkinsfile.integration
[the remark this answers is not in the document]
kantarun: All of the comments either have been addressed or follow up stories (tempest improvement) are created. So resolving these conversations

Jenkinsfile.integration
kantarun: Ended up implementing tempest logs in ansible play so this section is no longer needed and removed from jenkins file.

Jenkinsfile.integration
```diff
@@ -24,6 +24,8 @@ pipeline {
         /* first value in the list is the default */
         choice(choices: ['airship', 'osh'], description: 'Which deployment mechanism?', name: 'deployment')
         choice(choices: ['caasp3', 'caasp4'], description: 'Which caasp version to use?', name: 'caasp_version')
+        booleanParam(name: 'run_tempest', defaultValue: true, description: 'Run tempest tests?')
+        choice(choices: ['smoke', 'all'], description: 'Run smoke or all tempest tests?', name: 'tempest_test_type')
```
jgu17: Would you want a default value for the smoke vs all choices?

Itxaka: on `choices` first option is the default. Not very intuitive...

kantarun: Yes, I agree.

I will change default for run_tempest to false as @Itxaka mentioned, we want to avoid running tempest for PR change, even waiting for related automation changes to merge

PR #645, opened 2019-07-26
Jenkinsfile needs to skip teardown if only docs changes

ilausuch:
When only the documentation has changed CI will skip the
creation of the cluster. In this case a cleanup (teardonw)
is not necessary and generates an error. This commit
removes the execution of the cleanup.

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -109,6 +109,7 @@ pipeline {
                 timeout(time: 10, unit: 'MINUTES', activity: true)
             }
             steps {
+                sh 'touch /tmp/${SOCOK8S_ENVNAME}.needcleanup'
                 sh "./run.sh deploy_network"
             }
         }
@@ -199,7 +200,7 @@ pipeline {
         }
         cleanup {
             script {
-                    sh './run.sh teardown'
+                    sh '[ -f /tmp/${SOCOK8S_ENVNAME}.needcleanup ] && ./run.sh teardown && rm /tmp/${SOCOK8S_ENVNAME}.needcleanup'
                 }
         }
     }


Jenkinsfile.integration
```diff
@@ -199,7 +200,7 @@ pipeline {
         }
         cleanup {
             script {
-                    sh './run.sh teardown'
+                    sh '[ -f /tmp/$SOCOK8S_ENVNAME.needcleanup ] && ./run.sh teardown && rm /tmp/$SOCOK8S_ENVNAME.needcleanup'
```
Itxaka: I think we cannot leverage those `TestFunctional` and `TestDocs` vars for this rigth? Like they re not available? :(

Jenkinsfile.integration
```diff
@@ -109,6 +109,7 @@ pipeline {
                 timeout(time: 10, unit: 'MINUTES', activity: true)
             }
             steps {
+                sh 'touch /tmp/$SOCOK8S_ENVNAME.needcleanup'
```
Itxaka: should we use `${SOCOK8S_ENVNAME}` in case there is issues? *

* /me not a bash expert

evrardjp: For safety, I would do it. It's clearer. I think this works, though.

Jenkinsfile.integration
evrardjp: Good question. I think this is very simple, and I would be happy to see our CI gods improving over this, to feel it less hacky.

Itxaka: IIRC @jhesketh tried this and it didnt work? those vars and the `when` keyword are not part of the `post` workflow

Jenkinsfile.integration
```diff
@@ -199,7 +200,7 @@ pipeline {
         }
         cleanup {
             script {
-                    sh './run.sh teardown'
+                    sh '[ -f /tmp/${SOCOK8S_ENVNAME}.needcleanup ] && ./run.sh teardown && rm /tmp/${SOCOK8S_ENVNAME}.needcleanup'
```
ilausuch: I think would be better reorder the teardown and the rm commands in case teardown exits with error and to be sure that the file is removed. Is interesting to create a new patch for that?

evrardjp: not a bad idea.

PR #646, opened 2019-07-26
socok8s ansible didn't show failure status when deploy osh operation fails in Airship

chajain:
when deploy osh operation fails in Airship, the socok8s ansible playbook reports all nodes are successful (no error). The Airship Shipyard error messages are printed right before the final ansible status, just to be clear.

Modified the code so that if the Update software action failed, it prints a failed message and the deployer node status is failed
Also in the case when the no of retries are finished and the Update software action is still running, it prints a message without failing any nodes.

### DIFF
--- a/playbooks/roles/airship-deploy-osh/tasks/main.yml
+++ b/playbooks/roles/airship-deploy-osh/tasks/main.yml
@@ -177,26 +177,37 @@
     - update_airship_osh_site
 
 # TODO(aagate): Add a changed_when: to help idempotency
-- name: "Wait for update software action to complete... it can take up to {{ airship_deploy_openstack_timeout }} minutes"
-  command: '{{ shipyard }} describe {{ shipyard_action_key }}'
-  args:
-    chdir: '{{ upstream_repos_clone_folder }}/airship/shipyard'
-  environment:
-    SHIPYARD_IMAGE: "{{ shipyard_image }}"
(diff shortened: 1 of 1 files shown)


playbooks/roles/airship-deploy-osh/tasks/main.yml
```diff
@@ -177,26 +177,37 @@
...
-  retries: "{{ airship_deploy_openstack_timeout | int * 2 }}"
-  delay: 30
-  tags:
-    - skip_ansible_lint
-    - update_airship_osh_site
-    - add_compute_node
-
-- name: Print Shipyard update software action status
+- block:
+    - name: "Wait for update software action to complete... it can take up to {{ airship_deploy_openstack_timeout }} minutes"
```
Itxaka: why change the whitespace from 2 to 4? please follow up the rest of the file conventions.

playbooks/roles/airship-deploy-osh/tasks/main.yml
```diff
@@ -177,26 +177,37 @@
...
+      delay: 30
+      tags:
+        - skip_ansible_lint
+        - update_airship_osh_site
+        - add_compute_node
+
+  rescue:
+    - name: Print Shipyard update software action status
+      debug:
+        msg: "Update Software action has failed and stopped creating pods. Please check"
```
kantarun: Earlier we used to print actual error happened (Line# 198 in existing code)  which can be useful to triage failure step. Is that no longer useful to have it ?

chajain:
Closing this PR and moving the branch to own fork

PR #647, opened 2019-07-29
socok8s ansible didn't show failure status when deploy osh operation fails

chajain:
Modified the code so that if the no of retries is finished and the Update Software activity is still running, it will not fail any node and prints a message
Also added the code so that if Update Software activity is failed, it fails the deployer node and prints the failed message


Itxaka:
much nicer @chajain thanks for redoing the PR!

Looks good to me, but I guess either @kantarun or @jgu17 should have the last word here as they are the most knowledgeable in this area

PR #648, opened 2019-07-30
Upgrades Ansible to latest version

scottwulf:
Upgrades Ansible from 2.7.8 to 2.8.3

### DIFF
--- a/script_library/requirements.txt
+++ b/script_library/requirements.txt
@@ -1,4 +1,4 @@
-ansible==2.7.12
+ansible==2.8.3
 netaddr==0.7.19
 jmespath==0.9.3
 requests==2.21.0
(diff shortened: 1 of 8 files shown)


script_library/requirements.txt
```diff
@@ -1,4 +1,4 @@
-ansible==2.7.8
+ansible==2.8.3
```
kantarun: Do we know if this ansible version is going to be available as an rpm which is compatible to SLE15 sp1 (or SLE 15 later versions)?

I see 2.8.1 as latest factory version (https://build.suse.de/package/show/SUSE:Factory:Head/ansible )

We may want to make sure that core package are available in SLE versions which we are targeting for C10. Latest and greatest is definitely good but core package like ansible means, we can easily build code dependency with that version. 
That being said, it will work in virtualenv so may be its okay if plan is to use virtualenv for now.

toabctl:
@scottwulf needs a rebase...

script_library/requirements.txt
dirkmueller: ansible is not part of SLES15 so we can ship whatever we like. and newer is better.

PR #650, opened 2019-07-31
Allow choosing caasp version

Itxaka:
To allow for manually testing of caasp4 until we can enable it
by default, we need to have a path on CI to test it.

Adds a choice param that defaults to caasp3 and can manually be
chosen when triggering the job manually to allow for caasp4 testing.


Itxaka:
Seems to work, CI issues are unrelated but you can choose the caasp version and it defaults to caasp3: https://ci.suse.de/job/cloud-socok8s-jobs/job/socok8s-integration/job/PR-650/build?delay=0sec

PR #653, opened 2019-08-01
CaaSP 4 deploy should fail if requirements are not met


evrardjp:
By default we don't return an error code when doing the
pre-flight checks.

This is a problem, as things might look okay, and the deploy
continues, where it should stop.

This should fix it.


evrardjp:
CREATE failed: OverQuotaClient: resources.router: Quota exceeded for resources: [''router''].\nNeutron server returns request_ids: [''req-411c6268-efe0-4b54-992d-c35508bda203'']",

PR #656, opened 2019-08-02
Do not run caasp3 steps on caasp4 testing

Itxaka:
On caasp4 we are not using velum anymore so there is no need to run
the enroll caasp workers step.

This patch separates enroll_caasp_workers and
setup_caasp_workers_for_openstack into two steps so we can skip
uneeded ones on caasp4 CI testing

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -4,7 +4,6 @@
 TestDocs = true
 TestFunctional = true
 
-
 pipeline {
 
     options {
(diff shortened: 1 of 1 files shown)


evrardjp:
Alternatively, we could make sure the enroll of the caasp workers for caasp4 is done correctly, in other words, trigger the skuba calls in that step, instead of doing everything in one go.

Itxaka:
> Alternatively, we could make sure the enroll of the caasp workers for caasp4 is done correctly, in other words, trigger the skuba calls in that step, instead of doing everything in one go.

I think the most urgent thing here is to get to a testing caasp4 CI and then we can fold the different tasks into the caasp4 role if needed, or expand them into different roles.

evrardjp:
> I think the most urgent thing here is to get to a testing caasp4 CI and then we can fold the different tasks into the caasp4 role if needed, or expand them into different roles.

Agreed, this is why I voted positively.

PR #657, opened 2019-08-02
Caasp4 part 2: caasp harder (SOC-9926)

Itxaka:


script_library/deployment-actions-openstack.sh
```diff
@@ -32,6 +32,7 @@ function deploy_caasp4(){
     source ${scripts_absolute_dir}/pre-flight-checks.sh check_openstack_environment_is_ready_for_deploy
     source ${scripts_absolute_dir}/pre-flight-checks.sh check_caasp4_skuba_available
     source ${scripts_absolute_dir}/pre-flight-checks.sh check_caasp4_terraform_available
+    source ${scripts_absolute_dir}/pre-flight-checks.sh check_caasp4_terraform_version
```
jhesketh: (nit) I feel like these caasp4 checks could be rolled into one

evrardjp:
Jenkins complaints sh "eval '$(ssh-agent -s)'" is to hard for him/it to understand.

script_library/pre-flight-checks.sh
```diff
@@ -56,6 +56,15 @@ check_caasp4_terraform_available(){
         exit
     fi
 }
+
+check_caasp4_terraform_version(){
+  echo "Checking for terraform version (should be lower than 0.12)"
+  ${TERRAFORM_BINARY_PATH} version|egrep -qv "^Terraform\sv0\.12\.[0-9]{2}$"
```
toabctl: Can you check for 0.11 so we are good even if there is a new (eg. 0.13) terraform available?

playbooks/roles/caasp4/templates/skuba-terraform-openstack.tfvars.j2
```diff
@@ -29,7 +29,7 @@ packages = [
   "kernel-default",
   "-kernel-default-base",
   "ca-certificates-suse",
-  "patterns-caasp-Node",
+  "patterns-caasp-Node-1.14-1.15",
```
toabctl: Did that package rename happen by intention? Maybe the package misses a Priovides?

Jenkinsfile.integration
```diff
@@ -129,7 +129,10 @@ pipeline {
                 stage('Deploy CaaSP4') {
```
toabctl: This commit (bac2ee2cb93ea68a64c9b2e7c2368af6cd9e8ea0) is not needed given that you switch to the script from skuba in the next commit. Please remove the commit

Jenkinsfile.integration
```diff
@@ -47,6 +47,12 @@ pipeline {
                 sh 'printenv'
             }
         }
+        stage('Set ssh-agent for caasp4 deployment') {
+            when {  expression { return  TestFunctional } }
```
toabctl: nit: 2 spaces instead of 1

Jenkinsfile.integration
```diff
@@ -114,6 +120,7 @@ pipeline {
                 sh "./run.sh deploy_network"
             }
         }
+
```
toabctl: unrelated change

files/setup-ssh.py
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python3
+
```
toabctl: Could you add a note that this file is a copy from skuba? And if somebody wants to update it, it should be done in skuba and then synced

script_library/pre-flight-checks.sh
```diff
@@ -65,13 +65,7 @@ check_caasp4_terraform_version(){
...
   fi
 }
-check_caasp4_ssh_agent_running(){
-    echo "Checking if ssh-agent is running"
-    if ! ssh-add -L >/dev/null ; then
-        echo "ssh agent is not running or does not contain any identities"
-        exit
-    fi
-}
+
```
toabctl: unrelated

playbooks/roles/caasp4/tasks/setup_nodes.yml
```diff
@@ -34,14 +34,6 @@
   include_tasks: _setup_nodes_terraform_import_resources.yml
   when: import_terraform_resources|bool
 
-- debug:
```
toabctl: You added theses lines in the previous commit. Can you just not add the lines there instead of adding/removing it?

playbooks/roles/caasp4/templates/inventory-caasp4.yml.j2
```diff
@@ -2,7 +2,7 @@
 caasp-masters:
   hosts:
 {% for master_ip in _terraform_json_output.ip_masters.value %}
-    caasp-master-{{ loop.index0 }}:
+    master-{{ loop.index0 }}:
```
toabctl: Could describe (in the commit message) what the problem with `caasp-master-X` is? Where is that naming used afterwards? In Airship?

playbooks/roles/handle-nodes-on-openstack/tasks/create_on_openstack.yml
```diff
@@ -37,7 +37,7 @@
         hosts:
           {{ servername }}:
             ansible_host: {{ floating_ip }}
-            {% if ansible_user|default(False) %}ansible_user: {{ ansible_user }}{% endif %}
+            ansible_user: {{ ssh_user|default('root') }}
```
toabctl: Please mention in the commit msg why this is useful/needed

files/setup-ssh.py
Itxaka: tbh is not a copy from skuba directly as its modified by ourselves to write to a file and accept a parameter...I guess I could just pointthat is based off that file?

playbooks/roles/handle-nodes-on-openstack/tasks/create_on_openstack.yml
Itxaka: sorry about that, had to send the patch fast as it was late in the night, my brain was not able to write more than this  :)

playbooks/roles/caasp4/templates/inventory-caasp4.yml.j2
Itxaka: kubectl will report those names as said on the commit msg. It just makes sense to keep the same naming convention that we had before + what kubectl reports as its used afterwards for labeling the nodes for example.

I will try to expand the commit msg to make it more clear.

playbooks/roles/caasp4/tasks/setup_nodes.yml
Itxaka: :+1:

Jenkinsfile.integration
Itxaka: :+1:

script_library/pre-flight-checks.sh
Itxaka: good point

playbooks/roles/caasp4/templates/skuba-terraform-openstack.tfvars.j2
Itxaka: Yes, this was done on purpouse between beta 4 and 5: https://susedoc.github.io/caasp-release-notes/beta/release-notes/single-html/#_updating_from_beta_4_to_beta_5

Seems like its due to architectural changes and the package names are different because you need a manual step to upgrade. In our case we should go with 1.15 directly

playbooks/roles/airship-configure-worker/tasks/main.yml
```diff
@@ -1,95 +1 @@
...
-- name: Copy openstack apparmor profile into the workers
-  copy:
-    src: "{{ item }}"
-    dest: "/etc/apparmor.d/local/{{ item }}"
-  loop: "{{ apparmor_profiles }}"
-  register: apparmor_copy
-  notify: Reload apparmor profile
-  when:
-    - ('airship-openstack-control-workers' in group_names) or ('airship-openstack-compute-workers' in group_names)
+---
```
toabctl: Not sure, but can't we delete this file? Or is a `main.yml` a requirement for a role?

playbooks/roles/caasp4/templates/skuba-terraform-openstack.tfvars.j2
```diff
@@ -29,7 +29,7 @@ packages = [
   "kernel-default",
   "-kernel-default-base",
   "ca-certificates-suse",
-  "patterns-caasp-Node",
+  "patterns-caasp-Node-1.15",
```
evrardjp: You can remove this, skuba takes care of it.

toabctl: I already asked this and the answer from @Itxaka is here: https://github.com/SUSE-Cloud/socok8s/pull/657#discussion_r310976875

files/setup-ssh.py
toabctl: Or modify the file in skuba, too (if that makes sense). But keeping it as it is is fine for me

playbooks/roles/caasp4/tasks/setup_nodes.yml
```diff
@@ -55,3 +55,30 @@
   template:
     src: inventory-caasp4.yml.j2
     dest: "{{ socok8s_workspace }}/inventory/caasp4.yml"
+
+- meta: refresh_inventory
+
+- name: Get caasp ips
+  set_fact:
+    ssh_known_hosts_masters: "{{ groups['caasp-masters'] | map('extract', hostvars, ['ansible_host']) | list }}"
```
toabctl: I wonder if we want to pass a variable instead of using the `groups` variable here. Currently **all** variables that are used in the caasp4 role are defined in `defaults/main.yml` and set in `playbooks/openstack-deploy-caasp4.yml`. That way we have a clean interface (the variables) to interact with the role (instead of accessing $random variables from everywhere).

playbooks/roles/caasp4/tasks/setup_nodes.yml
```diff
@@ -55,3 +55,30 @@
...
+    ssh_known_hosts_workers: "{{ groups['caasp-workers'] | map('extract', hostvars, ['ansible_host']) | list }}"
```
toabctl: same here

playbooks/roles/airship-configure-worker/tasks/main.yml
Itxaka: I...dont know :o

playbooks/roles/caasp4/tasks/setup_nodes.yml
Itxaka: but this is not a random variable, its a dynamic variable that can change in teh future (more masters? more workers?)

We need to calculate it on the fly for this to work.

playbooks/roles/airship-configure-worker/tasks/main.yml
Itxaka: Seems like it:
`Roles expect files to be in certain directory names. Roles must include at least one of these directories, however it is perfectly fine to exclude any which are not being used. When in use, each directory must contain a main.yml file, which contains the relevant content`

from https://docs.ansible.com/ansible/latest/user_guide/playbooks_reuse_roles.html

playbooks/roles/caasp4/tasks/setup_nodes.yml
toabctl: Yeah - but it assumes that somebody created a group variable with the specific name . You need to
- a) document it somewhere that a user who wants to reuse this role needs to set it

or

b) Add it to defaults/main.yml (as empty list) and then overwrite it when calling the tasks that need the variable (in `playbooks/openstack-deploy-caasp4.yml`).

Itxaka: but that group var comes form the inventory and we are in control of the inventory names no? They should not change at all if we want everything to work, otherwise everything else that refers to caasp-workers/masters will fail to deploy

Itxaka:
CI reported the wrong job btw, the job reported to the PR is the caasp4 one while the caasp3 its still running correctly and will probably tun green: https://ci.suse.de/blue/organizations/jenkins/cloud-socok8s-jobs%2Fsocok8s-integration/detail/PR-657/56/pipeline

Itxaka:
> Looks good to me. We should perhaps generate a new private key for each CI run, but this could be done as a follow up.

Thats a really great idea!

playbooks/roles/caasp4/tasks/setup_nodes.yml
toabctl: ah. True. the variables are filled from the caasp4 role. sorry for the noise.

playbooks/roles/airship-configure-worker/tasks/main.yml
toabctl: thx

script_library/deployment-actions-openstack.sh
```diff
@@ -32,7 +32,7 @@ function deploy_caasp4(){
     source ${scripts_absolute_dir}/pre-flight-checks.sh check_openstack_environment_is_ready_for_deploy
     source ${scripts_absolute_dir}/pre-flight-checks.sh check_caasp4_skuba_available
     source ${scripts_absolute_dir}/pre-flight-checks.sh check_caasp4_terraform_available
```
ilausuch: Also the terraform needs that the variable TERRAFORM_BINARY_PATH exists and not empty. If we don't check it this check stucks forever here

This is a possible solution
check_caasp4_terraform_available(){
    echo "Checking for CaaSP 4 that terraform is available"
    if [ -z ${TERRAFORM_BINARY_PATH+x} ]; then
        echo "Terraform is not installed. Can not deploy CaaSP 4"
        exit
    fi
    ... Following lines
}

I added it to the PR I am preparing for the card SOC-9927 or we can include it here

Itxaka: I did not touch that, its coming from master so not touching it further.

Please send the PR into master if thats an issue :)

toabctl: TERRAFORM_BINARY_PATH is always set when you use run.sh . Or how do you run into the problem that the var does not exist?

toabctl: In any case, it is imo unrelated to this PR...

Itxaka:
Updated with a fix for the terraform resources laying around

playbooks/roles/caasp4/templates/skuba-terraform-openstack.tfvars.j2
```diff
@@ -29,7 +29,6 @@ packages = [
   "kernel-default",
   "-kernel-default-base",
   "ca-certificates-suse",
-  "patterns-caasp-Node",
```
toabctl: Commit msg does not reflect what is done here. Also I think we don't want to remove the pattern - or do we?

Itxaka: we do. Its skuba the one that will install the needed packages on node join, so there is no point in installing those packages already. We should leave it to skuba instead so it can manage the dependencies, versions and whatnot without our intervention :)

files/id_shared
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
toabctl: Applying the patch locally shows, that the file permissions for the private key are `0644` which leads to:

```
$ ssh -i files/id_shared sles@10.17.1.0
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
@         WARNING: UNPROTECTED PRIVATE KEY FILE!          @
@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
Permissions 0644 for 'files/id_shared' are too open.
It is required that your private key files are NOT accessible by others.
This private key will be ignored.
Load key "files/id_shared": bad permissions
Password: 
```

Can you change the permissions to 440 please?

Itxaka: ah sh*t, I swear I fixed this...I probably lost it in the rebase, sorry about that

files/setup-ssh.py
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python3
```
jsuchome: Could you rename it from generic `setup-ssh.py` to something like `setup-ssh-for-caasp`? 
We only need it for caasp and it's not deductible from file location so let's make it clear from the name...

Itxaka: good point, I changed it to `setup-ssh-caasp-ci.sh`

playbooks/roles/registry-server-setup/tasks/main.yml
```diff
@@ -18,6 +18,7 @@
   tags:
     - install
 
+# TODO(itxaka): Check if this is still valid on caasp4 as its using crio
```
jsuchome: What about creating Jira story instead? IMHO it is not lost so easily as random TODOs...

playbooks/roles/caasp4/templates/inventory-caasp4.yml.j2
jsuchome: Hm, but doesn't caasp3 deployment actually use the hostnames of the nodes? (I admit `caasp-master-X` etc. are not hostnames though)

playbooks/generic-deploy_airship.yml
```diff
@@ -42,10 +42,32 @@
...
+      tags:
+        - preinstall
+        - add_compute_node
+
+- hosts: caasp-workers
+  tasks:
+    - import_role:
+        name: airship-configure-worker
+        tasks_from: apparmor
+      when: (redeploy_osh_only is not defined or not redeploy_osh_only) and not caasp4_deployed
```
jsuchome: Why different parenthesis style than in previous task?

Itxaka: I just forgot, no other reason :)

playbooks/roles/registry-server-setup/tasks/main.yml
Itxaka: that is mainly because I expected to check that out myself and change the file accordingly. Especially because failures here do not make the deployment fail, Im not sure if this was needed or no. Hence the reminder, for myself :)

Jenkinsfile.integration
```diff
@@ -23,7 +23,7 @@ pipeline {
     parameters {
         /* first value in the list is the default */
         choice(choices: ['airship', 'osh'], description: 'Which deployment mechanism?', name: 'deployment')
-        choice(choices: ['caasp3', 'caasp4'], description: 'Which caasp version to use?', name: 'caasp_version')
+        choice(choices: ['caasp4', 'caasp3'], description: 'Which caasp version to use?', name: 'caasp_version')
```
jsuchome: Are you sure this is wise? We should still figure out some remaining problems, e.g. skuba networking

playbooks/roles/caasp4/templates/inventory-caasp4.yml.j2
Itxaka: it uses the inventory names, which in caasp3 are master-X and worker-X
Changing them in the caasp4 inventory was a mistake.

Im not sure why is this such a big deal tbh? We are just syncing the hostname with the inventory name so we can use the same playbooks as in caasp3 with caasp4. Who care what the names of the caasp workers/masters are in our inventory?

Jenkinsfile.integration
Itxaka: I dont expect this to be merged anytime soon, and Im sick of every push going into jenkins, canceling the caasp3 work and launching the caasp4 job instead.

This will be faster while we iterate over it and can be dropped in a second :+1:

playbooks/generic-deploy_airship.yml
Itxaka: fixed

files/setup-ssh.py
Itxaka: fixed?

files/id_shared
Itxaka: Fixed maybe with changing the keys to use the defaul user ones + jenkins keys

playbooks/roles/caasp4/templates/inventory-caasp4.yml.j2
jsuchome: I'm not saying it's a big deal, I only think your change does not make it same as in caasp3 (which probably does not matter)

Jenkinsfile.integration
jsuchome: OK

PR #658, opened 2019-08-02
Report custom context to github

Itxaka:
Use github-pr to report a combination of delivery mechanism +
caasp version to github in order to allow different jobs to run and
report back.

This allows us to run a combination of params for the same PR and
have easy forced checks for different paths, like airship+caasp3
and airship+caasp4, plus a variety of more combinations in the
future.

We could even have 2 jobs per PR in the future automatically,
but this is the first step to support that.

As a downside the normal context that reports automatically cannot
be disabled from jenkins as it would affect all other jobs that
use the github plugin, so instead we manage it ourselves and set it
to success at the start of the job as we create the other contexts,
so if this gets merged we should disable that mandatory check and enable
the airship+caasp3 instead

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -4,6 +4,11 @@
 TestDocs = true
 TestFunctional = true
 
+void setBuildStatus(String state, String context) {
+    full_context = "continuous-integration/jenkins/$context"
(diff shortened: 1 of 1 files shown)


Itxaka:
> We could look at changing how the job is triggered in the future to stop any pr-head being reported. However, my actual hope is that this is temporary and we remove the caaspv3 stuff in the near future.

Agreed, thats why I wiped this up, so its temporary and gets removed afterwards when we only have caasp4.

Itxaka:
We gonna move to caasp4 only so this is not needed. May come in handy for ses-in-rook so feel free to restore if needed.

PR #659, opened 2019-08-05
Creating needcleaup file inside workspace in the Jenkinsfile

ilausuch:
The needcleanup file removal is done after run.sh teardown.
This is a problem if teardown exists with an error, as the
needcleanup file won't be removed leaving gargage in /tmp,
This fixes it by moving this file into workspace

co-autors: @jhesketh @evrardjp

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -110,7 +110,7 @@ pipeline {
                 timeout(time: 10, unit: 'MINUTES', activity: true)
             }
             steps {
-                sh 'touch /tmp/${SOCOK8S_ENVNAME}.needcleanup'
+                sh 'touch ./.needcleanup'
                 sh "./run.sh deploy_network"
             }
         }
@@ -222,8 +222,8 @@ pipeline {
         }
         cleanup {
             script {
-                    sh '[ -f /tmp/${SOCOK8S_ENVNAME}.needcleanup ] && ./run.sh teardown && rm /tmp/${SOCOK8S_ENVNAME}.needcleanup'
-                }
+                sh 'if test -f ./.needcleanup; then ./run.sh teardown; fi'
+            }
         }
     }
 }


Jenkinsfile.integration
```diff
@@ -222,7 +222,7 @@ pipeline {
         }
         cleanup {
             script {
-                    sh '[ -f /tmp/${SOCOK8S_ENVNAME}.needcleanup ] && ./run.sh teardown && rm /tmp/${SOCOK8S_ENVNAME}.needcleanup'
```
toabctl: Isn't it correct to remove the file **only** if the cleanup (teardown) succeeded? I don't understand what you try to fix here.

Itxaka: If you have a docs only run, no network is created, so the teardown will fail and the file wont be cleaned up

Itxaka: Im guessing we could fix this based on the vars somehow

scottwulf: Good point. Don't modify the order. The "[ -f /tmp/${SOCOK8S_ENVNAME}.needcleanup ]" should be sufficient to skip teardown (i.e. remove the flag creation for docs). And if not, maybe introduce a new flag.

toabctl: > If you have a docs only run, no network is created, so the teardown will fail and the file wont be cleaned up

But if no network is created, the file is also not created.

ilausuch:
I've changed the PR message and also the title to clarify better the reason and the solution of this PR. (Thanks @evrardjp). But I didn't changed yet the commit message because the discussion is open and I am checking other ideas.

evrardjp:
Waiting for the "other ideas" before voting :)

toabctl:
I still don't get why it is a problem when teardown fails and the file is still there. Isn't that exactly the behavior we want to have? So teardown can run again and the file gets only deleted when teardown sucessfully ran?

Jenkinsfile.integration
jhesketh: Actually I think the check is wrong as it'll exit non-zero which Jenkins considers a failure.

The correct fix, I think, is here: https://github.com/SUSE-Cloud/socok8s/pull/663

However, if the teardown still fails then the file won't be removed so this reorder may still be useful. But in those cases actually having a needcleanup file hang around could be useful for us to see what didn't get fully cleaned.

ilausuch: In some conversations with @evrardjp a simple idea emerged, why don't keep the this file inside the workspace? Therefore if the CI eventually removes this workspace will remove this file too.

jhesketh: Because removing the Jenkins workspace doesn't mean that the engcloud resources have been properly torn down.

ilausuch: Yes, of course not. I agree with that. However I think the problem were are solving is how the needcleanup will be eventually removed (and where is stored) to prevent garbage in the CI. 
I understand if the task is tagged as functional then the CI will complete all the process, and finally, the teardown will be executed cleaning the resources. The condition prevents to execute the teardown (because fails) only in case of non-functional execution

jhesketh: Right, so either the teardown will be successful, after which the `.needcleanup` will be removed, or something fails during the teardown at which point having a `.needcleanup` hang around can be a useful indication to us.

In terms of garbage on the CI, it's a file in /tmp, so I'm not too concerned about it. Just my 2cents :-).

evrardjp: @jhesketh it will also be hard to track. Here everything is in the namespace.

For the test, I agree it could be made more userfriendly, as things will report a failure on docs run, which is improvable.

@ilausuch Can you make it a
```
if test -f ./.needcleanup; then
  ./run.sh teardown
fi
```

or alternatively, for the Jenkins wizards:

```
if (fileExists("/tmp/${env.SOCOK8S_ENVNAME}.needcleanup")) {
  sh './run.sh teardown'
}
```

jhesketh: > @jhesketh it will also be hard to track. Here everything is in the namespace.
> 
> For the test, I agree it could be made more userfriendly, as things will report a failure on docs run, which is improvable.
> 

I disagree as twe can put the SOCOK8S_ENVNAME in the file.

evrardjp: Oh yeah I missed that part. That's also fine for me. I just remember convos where we decided to put everything in workspace, and I thought it was backwards to put things outside namespace.

ilausuch:
This PR is closed because #663 is now merged and solves the problem

PR #662, opened 2019-08-06
Caasp4 Cleanup: fix a typo

jesusaurus:
Add jinja syntax to dereference the variable instead of treating it
as a string literal.


evrardjp:
I think this is fixed by another PR, but it doesn't hurt of doing so in this patch.

PR #663, opened 2019-08-07
Fix Jenkins cleanup

jhesketh:
Jenkins would fail if the needcleanup file didn't exist because the
shell test would return 1 which Jenkins considers a failure. Instead
properly test the file condition.


ilausuch:
An other proposal in #659

evrardjp:
Good catch @jhesketh ! Can we make that part of #659 ?

toabctl:
Nice. Thanks!

PR #666, opened 2019-08-08
Add openstack and heat clients to requirements.txt

jesusaurus:
When using a virtualenv, we need to install the openstack and heat CLIs.

### DIFF
--- a/script_library/requirements.txt
+++ b/script_library/requirements.txt
@@ -4,3 +4,5 @@ jmespath==0.9.3
 requests==2.21.0
 ara==0.16.3
 openstacksdk==0.22.0
+python-openstackclient
+python-heatclient


script_library/requirements.txt
```diff
@@ -4,3 +4,5 @@ jmespath==0.9.3
 requests==2.21.0
 ara==0.16.3
 openstacksdk==0.22.0
+python-openstackclient
+python-heatclient
```
kantarun: LGTM. Just wondering why we didnot see this issue earlier in virtualenv case ? Is it something changed recently as I am seeing issue with caasp v3 install case

jesusaurus: I haven't been able to figure out what changed either, I have no idea why this is just now showing up

script_library/requirements.txt
```diff
@@ -4,3 +4,5 @@ jmespath==0.9.3
 requests==2.21.0
 ara==0.16.3
 openstacksdk==0.22.0
+python-openstackclient
```
dirkmueller: So the commit message doesn't explain *why* we need those. normally the ansible scripts are using openstacksdk directly, why do we suddenly require the clients?

jesusaurus: I don't think it's sudden, but we have ansible command modules calling the openstack cli in `playbooks/openstack-create_network.yml` and `playbooks/roles/airship-deploy-tempest/tasks/main.yml`

Itxaka:
The problem with this is how venv and sistem wide installed packages work.

In my case, I had the openstack package installed system-wide so any call to the `openstack` binary would work as it was available, but as soon as I cleaned up that package from my computer, this arised, becuase the binary its not available in the venv path nor outside.

On the heatclient issue, no idea, but it may be similar. System-wide VS virtualenv. If we want to support virtualenv installs with no system pollution of packages, this should get in.

toabctl:
> The problem with this is how venv and sistem wide installed packages work.
> 
> In my case, I had the openstack package installed system-wide so any call to the `openstack` binary would work as it was available, but as soon as I cleaned up that package from my computer, this arised, becuase the binary its not available in the venv path nor outside.

Are you sure? The executable should be available from the virtualenv:
```
$ virtualenv x
[...]
$ source x/bin/activate
$ pip install python-openstackclient
[...]
$ which openstack
/home/tom/tmp/x/bin/openstack
```

PR #668, opened 2019-08-09
Check for skuba executable instead of installed RPM

toabctl:
We don't need the RPM installed, we need the skuba executable
available. It's totally fine to use a different skuba
version (eg. build from git).


toabctl:
Ah. Ok. Missed that. Thanks for the pointers!

PR #669, opened 2019-08-09
Add ses-rook (SOC-10065)

Itxaka:
Add deploy_ses_rook

Adds a role for deploying ses via rook
Adds the actions deploy_ses_rook and delete_ses_rook to run.sh
Adds disks to the caasp4 workers as they are needed to provision
ceph via rook
Adds testing of the ceph cluster by provisioning a pvc on a ceph
block storageclass
Adds the ses-rook path to jenkins
Adds a new choice to select the normal ses or ses-rook in jenkins

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -26,6 +26,7 @@ pipeline {
         choice(choices: ['caasp3', 'caasp4'], description: 'Which caasp version to use?', name: 'caasp_version')
         booleanParam(name: 'run_tempest', defaultValue: false, description: 'Run tempest tests?')
         choice(choices: ['smoke', 'all'], description: 'Run smoke or all tempest tests?', name: 'tempest_test_type')
+        choice(choices: ['ses_aio', 'ses_rook'], description: 'Which ses version to use?', name: 'ses_version')
     }
 
     environment {
@@ -35,6 +36,7 @@ pipeline {
         SOCOK8S_ENVNAME = "cloud-socok8s-${env.BRANCH_NAME.replaceAll("[^a-zA-Z0-9-]+", "-").toLowerCase()}-${env.BUILD_NUMBER}"
         SOCOK8S_OPENSUSE_MIRROR="https://provo-mirror.opensuse.org"
         OS_CLOUD = "engcloud-socok8s-ci"
+        SOCOK8S_TEST_CEPH_ROOK="true"
         KEYNAME = "engcloud-cloud-ci"
         DELETE_ANYWAY = "YES"
         DEPLOYMENT_MECHANISM = "openstack"
@@ -153,6 +155,7 @@ pipeline {
                     }
                 }
                 stage('Deploy SES') {
+                    when { expression  { params.ses_version == "ses_aio" } }
                     steps {
                         sh "./run.sh deploy_ses"
                     }
@@ -188,6 +191,13 @@ pipeline {
             }
             steps {
                 sh "source ${WORKSPACE}/sock_${env.SOCOK8S_ENVNAME} && ./run.sh setup_caasp_workers_for_openstack"
+            }
+        }
+
+        stage('Deploy SES on Kubernetes with rook') {
+            when { expression  { params.ses_version == "ses_rook" } }
+            steps {
+                sh "./run.sh deploy_ses_rook"
             }
         }
 
--- a/playbooks/generic-deploy_ses_rook.yml
+++ b/playbooks/generic-deploy_ses_rook.yml
@@ -1,0 +1,47 @@
+---
+- hosts: localhost
+  gather_facts: no
+  tasks:
+    - name: Caasp is deployed and we have workers
+      assert:
+        that: "{{ groups['caasp-workers'] | length > 0 }}"
+        fail_msg: You need Caasp deployed in order to deploy ses with rook
+        success_msg: Caasp workers found
+    - name: We have a deployer
+      assert:
+        that: "{{ groups['soc-deployer'] | length > 0 }}"
+        fail_msg: You need a soc-deployer node in order to deploy ses with rook
+        success_msg: Deployer found
+# TODO(itxaka): Remove this once the images are pushed to registry.suse.com
+- hosts: caasp-masters:caasp-workers
+  gather_facts: no
+  tasks:
+    - name: Configure insecure registry for registry.suse.de
+      become: true
+      replace:
+        path: "/etc/containers/registries.conf"
+        regexp: '^\[registries\.insecure]\nregistries = \[]\n'
+        replace: '[registries.insecure]\nregistries = ["registry.suse.de"]\n'
+      register: _registry_insecure
+      tags:
+        - bug
+    - name: Reload crio
+      become: true
+      systemd:
+        name: crio.service
+        state: reloaded
+      when: _registry_insecure.changed
+      tags:
+        - bug
+        - skip_ansible_lint
+- hosts: soc-deployer
+  gather_facts: yes
+  any_errors_fatal: true
+  vars_files:
+    - "{{ playbook_dir }}/../vars/common-vars.yml"
+  tasks:
+    - import_role:
+        name: common-deployer
+        tasks_from: kubectl-setup
+    - import_role:
+        name: ses-rook
--- a/playbooks/roles/ses-rook/defaults/main.yml
+++ b/playbooks/roles/ses-rook/defaults/main.yml
@@ -1,0 +1,5 @@
+ses_rook_repos:
+  ses_rook_sle15_sp1: "http://download.suse.de/ibs/Devel:/Storage:/6.0/SLE_15_SP1/"
+ses_rook_yaml_manifests_package: "rook-k8s-yaml"
+ses_rook_k8s_manifests_dir: "/usr/share/k8s-yaml/rook/ceph"
+ses_rook_config_path_in_caasp_workers: "/home/ses/var/lib/rook/"
--- a/playbooks/roles/ses-rook/files/test_pvc.yml
+++ b/playbooks/roles/ses-rook/files/test_pvc.yml
@@ -1,0 +1,14 @@
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: test-pvc
+  labels:
+    app: test
+spec:
+  storageClassName: rook-ceph-block
+  accessModes:
+  - ReadWriteOnce
+  resources:
+    requests:
+      storage: 1Gi
--- a/playbooks/roles/ses-rook/files/test_storageclass.yml
+++ b/playbooks/roles/ses-rook/files/test_storageclass.yml
@@ -1,0 +1,21 @@
+---
+apiVersion: ceph.rook.io/v1
+kind: CephBlockPool
+metadata:
+  name: replicapool
+  namespace: rook-ceph
+spec:
+  failureDomain: host
+  replicated:
+    size: 3
+---
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+   name: rook-ceph-block
+provisioner: ceph.rook.io/block
+parameters:
+  blockPool: replicapool
+  clusterNamespace: rook-ceph
+  fstype: xfs
+reclaimPolicy: Delete
--- a/playbooks/roles/ses-rook/tasks/main.yml
+++ b/playbooks/roles/ses-rook/tasks/main.yml
@@ -1,0 +1,99 @@
+---
+- name: Add SES repositories
+  become: yes
+  zypper_repository:
+    name: "{{ item['key'] }}"
+    repo: "{{ item['value'] }}"
+    autorefresh: True
+    auto_import_keys: yes
+    state: present
+  loop: "{{ ses_rook_repos | dict2items }}"
+  tags:
+    - ses-rook-repository
+
+- name: Install packages
+  become: yes
+  package:
+    name: "{{ ses_rook_yaml_manifests_package }}"
+    state: present
+  register: install_packages
+  until: install_packages is success
+  retries: 5
+  delay: 2
+  tags:
+    - ses-rook-install
+
+- name: Point to registry.suse.de until images are published at registry.suse.com
+  replace:
+    path: "{{ item }}"
+    regexp: 'registry.suse.com/ses/6/'
+    replace: 'registry.suse.de/devel/storage/6.0/containers/ses/6/'
+  loop:
+    - "{{ ses_rook_k8s_manifests_dir }}/operator.yaml"
+    - "{{ ses_rook_k8s_manifests_dir }}/cluster.yaml"
+    - "{{ ses_rook_k8s_manifests_dir }}/toolbox.yaml"
+
+- name: Create rook operator
+  command: "kubectl apply -f common.yaml -f operator.yaml"
+  register: _rook_operator_creation
+  args:
+    chdir: "{{ ses_rook_k8s_manifests_dir }}"
+  changed_when: "'created' in _rook_operator_creation.stdout"
+
+- name: Wait for rook operator to be running
+  command: "kubectl get pods -n rook-ceph -l app=rook-ceph-operator -o jsonpath='{.items[].status.containerStatuses[].ready}'"
+  register: _rook_ceph_operator_status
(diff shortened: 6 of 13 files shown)


Jenkinsfile.integration
```diff
@@ -24,6 +24,7 @@ pipeline {
         /* first value in the list is the default */
         choice(choices: ['airship', 'osh'], description: 'Which deployment mechanism?', name: 'deployment')
         choice(choices: ['caasp3', 'caasp4'], description: 'Which caasp version to use?', name: 'caasp_version')
+        choice(choices: ['ses', 'ses_rook'], description: 'Which ses version to use?', name: 'ses_version')
```
jsuchome: What about `ses_aio` (or `ses_all_in_one`) instead of just `ses`?

playbooks/roles/ses-rook/tasks/main.yml
```diff
@@ -0,0 +1,104 @@
+---
+- name: Ensure kubectl is properly configured
+  include_role:
+    name: common-deployer
+    tasks_from: kubectl-setup
```
jsuchome: Hm, won't this be called again from deploy-osh/airship-setup-deployer ?

playbooks/roles/ses-rook/tasks/main.yml
```diff
@@ -0,0 +1,104 @@
...
+    name: "{{ ses_rook_package }}"
+    state: present
+  register: install_packages
+  until: install_packages is success
+  retries: 5
+  delay: 2
+  tags:
+    - ses-rook-install
+
+- name: Point to registry.suse.de until images at published at registry.suse.com
```
jsuchome: typo: images _are_ published

script_library/deployment-actions-common.sh
```diff
@@ -22,6 +22,16 @@ function deploy_osh(){
     run_ansible ${socok8s_absolute_dir}/playbooks/generic-deploy_osh.yml
 }
 
+function deploy_ses_rook(){
+    echo "Deploying SES with rook"
```
jsuchome: I would extend it with "... into kubernetes cluster" because that's the difference to the other SES deployment

playbooks/roles/ses-rook/tasks/main.yml
ilausuch: Maybe airship-setup-deployer will be executed in a most advanced step?

Itxaka: yes it will, but unfortunately is needed in this step as well, so we need to execute it before otherwise there is no deploying ses-in-rook. Its idempotent so it should not matter much.

script_library/deployment-actions-common.sh
Itxaka: well, rook is a kubernetes orchestrator so there shouldnt be much confusion, but ok, will change it

Jenkinsfile.integration
Itxaka: good to me

playbooks/roles/ses-rook/tasks/main.yml
jsuchome: OK, we can reorganize the code if we decide we actually want to deploy with rook for real

Jenkinsfile.integration
Itxaka: fixed

playbooks/roles/ses-rook/tasks/main.yml
Itxaka: fixed

script_library/deployment-actions-common.sh
Itxaka: fixed

Itxaka:
currently ses is properly deployed via rook but airship fails because its looking for some vars only available from normal ses deployment.

As this is a POC Im not sure we want to continue pursuing this or we are ok with this as something to build upon

playbooks/generic-deploy_ses_rook.yml
```diff
@@ -0,0 +1,47 @@
...
+        state: reloaded
+      when: _registry_insecure.changed
+      tags:
+        - bug
+        - skip_ansible_lint
+- hosts: soc-deployer
+  gather_facts: yes
+  any_errors_fatal: true
+  vars_files:
+    - "{{ playbook_dir }}/../vars/common-vars.yml"
```
toabctl: Could we try to keep the `ses-rook` role self contained? And stop using all available variables just everywhere?

playbooks/roles/ses-rook/defaults/main.yml
```diff
@@ -0,0 +1,5 @@
+ses_rook_repos:
```
toabctl: Could you document the different variables so it's more clear what these variables are good for?

playbooks/roles/ses-rook/files/test_pvc.yml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
```
toabctl: nit: `---` missing

playbooks/roles/ses-rook/files/test_storageclass.yml
```diff
@@ -0,0 +1,20 @@
+apiVersion: ceph.rook.io/v1
```
toabctl: nit: `---` missing

playbooks/roles/ses-rook/tasks/main.yml
```diff
@@ -0,0 +1,104 @@
+---
+- name: Ensure kubectl is properly configured
```
toabctl: Can we call this outside of the `ses-rook` role? To not depend on the `common-deployer` role?

playbooks/roles/ses-rook/tasks/main.yml
```diff
@@ -0,0 +1,104 @@
...
+  include_role:
+    name: common-deployer
+    tasks_from: kubectl-setup
+
+- name: Add soc extra repositories
```
toabctl: nit: should that be `Add SES extra repositories` ?

playbooks/roles/ses-rook/tasks/main.yml
```diff
@@ -0,0 +1,104 @@
...
+  become: yes
+  zypper_repository:
+    name: "{{ item['key'] }}"
+    repo: "{{ item['value'] }}"
+    autorefresh: True
+    auto_import_keys: yes
+    state: present
+  loop: "{{ ses_rook_repos | dict2items }}"
+  tags:
+    - ses-rook-repository
```
toabctl: What is this tag good for?

playbooks/roles/ses-rook/tasks/remove_ses_rook.yml
```diff
@@ -0,0 +1,71 @@
...
+    name: "{{ ses_rook_package }}"
+    state: absent
+  register: remove_packages
+  until: remove_packages is success
+  retries: 5
+  delay: 2
+  tags:
+    - ses-rook-remove
+
+- name: Delete soc extra repositories
```
toabctl: I guess `... SES extra ...` ?

playbooks/generic-deploy_ses_rook.yml
Itxaka: why? common vars are...common vars, they are to be used by every other role/playbook. This role has no use outside of this repo, so its not like we plan on support it as standalone.

playbooks/roles/ses-rook/defaults/main.yml
Itxaka: really? I think its kind of clear with their names. Which one is not clear enough to what it refers?

playbooks/roles/ses-rook/tasks/main.yml
Itxaka: for retrying the same step without running the other steps, so you can re-run the same step again and again for whatever reason you wan (only that task its failing, you are tweaking it and want to test that task again, it timed out due to network issues, etc...)

playbooks/roles/ses-rook/tasks/main.yml
Itxaka: :+1:

playbooks/roles/ses-rook/tasks/main.yml
Itxaka: We could, but what do we gain here? Again, this role is integrated on socok8s, it only works there. Dow e want to make a generic ses-rook role that its not used anywhere but in this repo and maintain the extra changes that would require a more abstract role?

playbooks/roles/ses-rook/tasks/remove_ses_rook.yml
Itxaka: :+1:

playbooks/roles/ses-rook/files/test_pvc.yml
Itxaka: somehow `---` is not mandatory if you dont start your yaml with a directive

playbooks/generic-deploy_ses_rook.yml
toabctl: - I would say it's good style to keep things separate
- you can use common vars - just pass them to the role when using the role
- debugging is imo easier 
- testing is easier ( you could have a simple playbook that just deploys ses-rook)
- people might want to use it (eg. for crowbar/ardana SES 6 integration testing) outside of this repo

If people disagree I'm happy to remove my -1

playbooks/roles/ses-rook/defaults/main.yml
toabctl: - `ses_rook_path_in_workers` ? Which workers? CaaSP workers? SES OSD`s ?
- `ses_rook_manifests_dir` ? What manifests? For helm maybe?
- `ses_rook_package` ? I guess the RPM package name that needs to be installed on the deployer to have the manifests available?

playbooks/roles/ses-rook/tasks/main.yml
toabctl: I would say it's more clear (instead of more abstract). If you google for `ansible role` you get things like:
```
Ansible role is an independent component which allows reuse of common configuration steps.
```

If we don't keep roles independent, why are we using them? We could just have one role (eg `socok8s` and have 100 task files.

playbooks/generic-deploy_ses_rook.yml
Itxaka: while I agree with you in there, and its the way things should be. I dont think we have anything like that. Would be good if the role was standalone, which would bring all those improvements, but do we want to make it more generic?

I guess it brings it down to, are we gonna ship that part to customers? Or is this just a POC to test the viability of deployment and how airship behaves with ses-rook? If its the latter, I would say no more work that necessary is needed for something to be trashed later or improved upon when the requirements are clear.

In any case, this is still a POC/preview so that should be taken into account. Im glad to spend more time refining this, but I dont think that falls into the current task.

@jhesketh guidance here pretty please with a :cherries: on top?

playbooks/roles/ses-rook/tasks/main.yml
jsuchome: I do not understand, are you arguing against shared code in principle?

Itxaka: I do agree with that but no other role in our repo is independent. Its more like a separation of concerns and organization rather than a real role (flashbacks of crowbar all around :fearful: )

See above answer about this.

playbooks/generic-deploy_ses_rook.yml
toabctl: > while I agree with you in there, and its the way things should be. I dont think we have anything like that. 

We have - the `caasp4` role is standalone.

> Would be good if the role was standalone, which would bring all those improvements, but do > we want to make it more generic?
> 
> I guess it brings it down to, are we gonna ship that part to customers? Or is this just a POC to test the viability of deployment and how airship behaves with ses-rook? If its the latter, I would say no more work that necessary is needed for something to be trashed later or improved upon when the requirements are clear.

I don't see why it matters if we ship this to customers. It's just about how we organize code. Why do you think it is more work to keep the role separate? It's really just passing variables from the code where you use the role (`playbooks/generic-deploy_ses_rook.yml`) to the role. Nothing complicated. Nothing that needs a lot of thinking. Or am I missing something?

playbooks/roles/ses-rook/defaults/main.yml
Itxaka: I....I do not agree here but will try to make them clearer. Dont blame me if we end up with a 80 chars var thought, Im terrible at naming stuff!

playbooks/roles/ses-rook/tasks/main.yml
toabctl: @jsuchome no. I'm arguing for **reusable and shareable** code. I would like to have roles that are independent from each other so the roles could be reused (or even uploaded to ansible galaxy). And the connection between roles is done in the playbooks where the roles are used (instead of including our `common-vars` into every role).
I know that's not what we've done in the past but we could start with it for new roles.

playbooks/roles/ses-rook/tasks/main.yml
toabctl: Similar to `--start-at-task` when calling ansible-playbook?

Itxaka: I guess so (never heard of that) but this still allows you to trigger random tasks, instead of starting from task 4 and containing in a lineal fashion. This allows you to only trigger tasks with a given tag(s)

playbooks/generic-deploy_ses_rook.yml
Itxaka: To be fair, Im not even sure what the common vars are doing here, now that Im looking at it...

Itxaka:
@toabctl I updated it to move any dependant things on other roles/tasks into the playbook and remove any non-generic calls out, so the role should be pretty much standalone now!

Itxaka:
Seems to pass correctly with the latest updates: https://ci.suse.de/blue/organizations/jenkins/cloud-socok8s-jobs%2Fsocok8s-integration/detail/PR-669/21/pipeline

Removing the extra commit for CI + do-not-merge tag

playbooks/generic-deploy_ses_rook.yml
Itxaka: vars are now only used for the kubectl deployment, ses-rook should be standalone

playbooks/roles/ses-rook/defaults/main.yml
Itxaka: Hope they are clearer now?

Jenkinsfile.integration
```diff
@@ -26,6 +26,7 @@ pipeline {
         choice(choices: ['caasp3', 'caasp4'], description: 'Which caasp version to use?', name: 'caasp_version')
         booleanParam(name: 'run_tempest', defaultValue: false, description: 'Run tempest tests?')
         choice(choices: ['smoke', 'all'], description: 'Run smoke or all tempest tests?', name: 'tempest_test_type')
+        choice(choices: ['ses_aio', 'ses_rook'], description: 'Which ses version to use?', name: 'ses_version')
```
toabctl: Just for my understanding - does `ses_aio` mean SES5 and `ses_rook` SES6?

playbooks/roles/ses-rook/defaults/main.yml
toabctl: I was more thinking about commenting the variable instead of changing the name - anyway - looks better imo :)

playbooks/roles/ses-rook/tasks/main.yml
toabctl: Still not fixed but can be done later...

Jenkinsfile.integration
Itxaka: yep, ses_aio was the default name because of the all-in-one approach to having just one node. ses_rook could be 6 or 7 I think, not clean on that one yet.

jhesketh:
Do we somehow need to lock this to caaspv4 in order to pass ci?

jsuchome:
This code path is not executed by CI so the failures are unrelated...

Itxaka:
> Do we somehow need to lock this to caaspv4 in order to pass ci?

yep, ses-in-rook is only deployable under caasp4, so they go together

jsuchome:
Merging, does not affect CI

PR #673, opened 2019-08-12
Remove CaasPv3 as default

ilausuch:
Once CaaSPv4 is working we should make it the default and remove CaaSPv3 from socok8s.

Need to determine what needs removing (ie, what is v3 specific).

AC

    CaaSPv4 is deployed by default
    CaaSPv3 is removed
    Current CI continues to pass
    Update the documentation where required


Jenkinsfile.integration
```diff
@@ -23,7 +23,7 @@ pipeline {
     parameters {
         /* first value in the list is the default */
         choice(choices: ['airship', 'osh'], description: 'Which deployment mechanism?', name: 'deployment')
-        choice(choices: ['caasp3', 'caasp4'], description: 'Which caasp version to use?', name: 'caasp_version')
+        choice(choices: ['caasp4'], description: 'Which caasp version to use?', name: 'caasp_version')
```
Itxaka: should be removed if we are going for caasp4 only

playbooks/openstack-delete_caasp.yml
```diff
@@ -1,37 +1,39 @@
...
-  post_tasks:
-    - name: Load generic vars
-      include_vars: "{{ playbook_dir }}/../vars/common-vars.yml"
-    - name: Load openstack vars
-      include_vars: "{{ playbook_dir }}/../vars/deploy-on-openstack.yml"
-
-    - name: Define stackname to create
-      set_fact:
-        stackname: "{{ socok8s_envname }}-{{ stackname_suffix }}"
+#
```
Itxaka: Im guessing this is just a drop of openstack-delete_caasp + rename of the caasp4 file. I guess github is not that smart :P

Jenkinsfile.integration
```diff
@@ -128,16 +128,10 @@ pipeline {
...
             }
             parallel {
-                stage('Deploy CaaSP3') {
-                    when { expression  { params.caasp_version == "caasp3" } }
-                    steps {
-                        sh "./run.sh deploy_caasp"
-                    }
-                }
                 stage('Deploy CaaSP4') {
                     when { expression  { params.caasp_version == "caasp4" } }
```
Itxaka: remove if caasp4 only as it should alwasy run

run.sh
```diff
@@ -69,12 +69,8 @@ case "$deployment_action" in
         deploy_ses
         ;;
     "deploy_caasp")
-        # this is CaaSP 3
-        deploy_caasp
-        ;;
-    "deploy_caasp4")
         # this is CaaSP 4
```
Itxaka: would remove the comment

Jenkinsfile.integration
ilausuch: I understood that because the second point in the card acceptance criteria is remove CaaSPv3. Is it correct? Has to be removed from CI too? @jhesketh

Jenkinsfile.integration
ilausuch: Yes, good point

Jenkinsfile.integration
Itxaka: well, if we remove caasp3 playbooks, then its impossible for us to test caasp3 in CI so it should also be dropped

playbooks/openstack-delete_caasp.yml
ilausuch: Yes, this is the case

Jenkinsfile.integration
jsuchome: Yes, let's remove the choice completely

jsuchome:
Should we include documentation changes as part of this PR or do we leave it for later? (I'm fine with another PR)

jsuchome:
Please fix 2nd commit text: "CaaSPv3 cannot deloyed"

jsuchome:
Continued in https://github.com/SUSE-Cloud/socok8s/pull/691

PR #675, opened 2019-08-13
Check if ssh-agent is running for openstack deployment

jsuchome:
CaaSP4 requires ssh-agent running so let's check (and fail) early
before we get messages that terraform cannot connect to the nodes.


jsuchome:
@ilausuch what description? Do you think current one is missing something?

Itxaka:
Indeed seems to be missing the `why?`

Plus this may conflict with the caasp4 deployment, which need an ssh-agent but has to create it itself and store the sock. Would like to see one caasp4 run with this check

jsuchome:
Well, this was something I already did for kvm deployment and just forgot to include it in openstack part.

> Plus this may conflict with the caasp4 deployment, which need an ssh-agent but has to create it 

Does it? This CI step that starts ssh-agent is executed before deploy_caasp4 step, right?

Itxaka:
> Does it? This CI step that starts ssh-agent is executed before deploy_caasp4 step, right?

It should but tbh I dont know what the check does :)

ilausuch:
My comment was about the PR description that is empty.

jsuchome:
OK, I could add something to PR description , but more important is the commit message itself

PR #679, opened 2019-08-14
Modify the message of minimum version required for ansible

ilausuch:
Now is set to 2.8 but it doesn't work with 2.8.1 version so
it is confusing.

Testing with version 2.8.3 this works. So The message has been
modified

### DIFF
--- a/script_library/pre-flight-checks.sh
+++ b/script_library/pre-flight-checks.sh
@@ -106,7 +106,7 @@ check_ansible_requirements (){
     fi
     # We need ansible version 2.8 minimum
     if [[ $(ansible --version | awk 'NR==1 { gsub(/[.]/,""); print substr($2,0,2); }' ) -lt "28" ]]; then
-        echo "Insufficent version of ansible: 2.8 or greater is required."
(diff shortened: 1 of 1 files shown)


script_library/pre-flight-checks.sh
```diff
@@ -106,7 +106,7 @@ check_ansible_requirements (){
     fi
     # We need ansible version 2.8 minimum
     if [[ $(ansible --version | awk 'NR==1 { gsub(/[.]/,""); print substr($2,0,2); }' ) -lt "28" ]]; then
```
jhesketh: We should change the actual check if it's true that 2.8.3 is what is required. Do you know why 2.8.1 doesn't work?

script_library/pre-flight-checks.sh
```diff
@@ -106,7 +106,7 @@ check_ansible_requirements (){
...
-        echo "Insufficent version of ansible: 2.8 or greater is required."
+        echo "Insufficent version of ansible: 2.8.3 or greater is required."
```
jsuchome: Why is 2.8.3 or greater required? Don't you wan't to fix the check if it fails for your 2.8.1?

PR #680, opened 2019-08-14
Locally build site and transfer it to deployer

Itxaka:
Ansible template is a bit slower, especially when doing local to
remote as it needs to check if the files exists, then check if
its different from the computed one and then template+transfer it
to the remote node.

In my local tests building the site from the templates could take from
30 minutes on a good connection to almost an hour in a connection with
a high latency (~1s).

Instead, lets just build the whole site in localhost under the workspace
and then transfer it on one go to the remote machine.

Currently I clean up the local copy of the site to keep everything
clean, but potentially we could leave those files up for ease of
checking what the site looks like without having to log into the
deployer. But this can come later, currently it just cleans up
everything at the end so we dont polute the workspace more than its
needed.

This results in an astonishing speed increase, taking only about 40s
in my high latency scenario (~1s). No downsides that I can see in this
approach

### DIFF
--- a/playbooks/roles/airship-deploy-ucp/tasks/main.yml
+++ b/playbooks/roles/airship-deploy-ucp/tasks/main.yml
@@ -7,7 +7,8 @@
 
 - name: Set site path
   set_fact:
-    site_path: "{{ upstream_repos_clone_folder }}/airship/treasuremap/site/{{ socok8s_site_name }}"
+    site_path: "{{ upstream_repos_clone_folder }}/airship/treasuremap/site/{{ socok8s_site_name }}/"
+    workspace_site_path: "{{ socok8s_workspace }}/{{ socok8s_site_name }}/"
   tags:
     - always
 
@@ -18,15 +19,20 @@
     state: absent
 
 - name: Ensure site directory for {{ socok8s_site_name }} exists
-  become: yes
   file:
     path: "{{ site_path }}"
     state: directory
 
-- name: Create site manifest directory structure
-  become: yes
+- name: Ensure local site directory exists
+  delegate_to: localhost
   file:
-    path: '{{ site_path }}/{{ item.path }}'
+    path: "{{ workspace_site_path }}"
+    state: directory
+
+- name: Create site manifest directory structure
+  delegate_to: localhost
+  file:
+    path: '{{ workspace_site_path }}/{{ item.path }}'
     state: directory
   with_filetree: '{{ playbook_dir }}/../site/soc'
   when: item.state == 'directory'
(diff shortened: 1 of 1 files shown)


playbooks/roles/airship-deploy-ucp/tasks/main.yml
```diff
@@ -47,38 +53,48 @@
...
+    dest: "{{ site_path }}"
+    recursive: yes
+
+- name: Remove local site
+  file:
+    path: "{{ local_temp_path }}"
+    state: absent
+
 - name: Create pod scale profile
-  become: yes
```
jsuchome: Why can you remove `become` here? It does not seem relevant to the other checks

playbooks/roles/airship-deploy-ucp/tasks/main.yml
```diff
@@ -7,7 +7,8 @@
 
 - name: Set site path
   set_fact:
-    site_path: "{{ upstream_repos_clone_folder }}/airship/treasuremap/site/{{ socok8s_site_name }}"
+    site_path: "{{ upstream_repos_clone_folder }}/airship/treasuremap/site/{{ socok8s_site_name }}/"
+    local_temp_path: "{{ socok8s_workspace }}/{{ socok8s_site_name }}/"
```
jsuchome: I think _local_ and _temp_ are bit redundant. What about `tmp_site_path`, `local_site_path` or something like that?

playbooks/roles/airship-deploy-ucp/tasks/main.yml
Itxaka: good point, I did miss this one

playbooks/roles/airship-deploy-ucp/tasks/main.yml
Itxaka: local its to indicate that its local to the ansible run, so its on local-host

I did not want to have any confusion between loca as in locahost and local as in local to the remote node

jsuchome: So you can just exclude `local_` completely and leave `tmp_site_path`?

Anyway, this is just me nitpicking, feel free to ignore

Itxaka: changed it to workspace_site_path so its clearer...maybe?

playbooks/roles/airship-deploy-ucp/tasks/main.yml
Itxaka: fixed!

PR #681, opened 2019-08-14
Drop pattern for caasp4 nodes (libvirt)

rhafer:
There is no need to install any package or pattern manually as
skuba will do so when joining the no9des into the cluster, so
just drop the uneeded pattern

This is commit ea584a0fe ported for the libvirt terraform
implementation.

### DIFF
--- a/playbooks/roles/caasp4/templates/skuba-terraform-libvirt.tfvars.j2
+++ b/playbooks/roles/caasp4/templates/skuba-terraform-libvirt.tfvars.j2
@@ -29,7 +29,6 @@ packages = [
   "kernel-default",
   "-kernel-default-base",
   "ca-certificates-suse",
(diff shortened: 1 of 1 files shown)


jsuchome:
Could you fix the original typo in the commit message? (_no9des_)

rhafer:
@jsuchome Done.

Itxaka:
Thanks for that @rhafer 

Is the libvirt implementation being used somewhere? I kind of forget that it exists because I cannot use it but maybe we should focus a bit more into it?

toabctl:
> Is the libvirt implementation being used somewhere? I kind of forget that it exists because I cannot use it but maybe we should focus a bit more into it?

@Itxaka at least Ralf and me try to use it on our workstations.

PR #682, opened 2019-08-14
Add Jenkinsfile support for lockable resources

jesusaurus:
In order to move our CI testing from the cloud project to the cloud-ci
project, we need our jobs to support lockable resources. Add job
parameters for initial support.

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -43,9 +43,6 @@ pipeline {
         // reserve a resource if instructed to do so, otherwise use a dummy resource
         // and a zero quantity to fool Jenkins into thinking it reserved a resource when in fact it didn't
         // TODO(jharker): don't lock a resource for docs-only jobs
-        lock(label: reserve_env == 'true' ? cloud_env:'dummy-resource',
-             variable: 'reserved_env',
-             quantity: reserve_env == 'true' ? 1:0 )
         // Note(jhesketh): Unfortunately we can't set a global timeout for the
         //                 pipeline as it would also apply to the post stages
         //                 and hence interrupt our cleanup.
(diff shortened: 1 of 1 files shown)


Jenkinsfile.integration
```diff
@@ -41,6 +27,29 @@ pipeline {
         USER = "jenkins" /* Why isn't this set in the jenkins environment? */
     }
 
+    parameters {
+        /* first value in the list is the default */
+        choice(choices: ['airship', 'osh'], description: 'Which deployment mechanism?', name: 'deployment')
+        choice(choices: ['caasp3', 'caasp4'], description: 'Which caasp version to use?', name: 'caasp_version')
+        booleanParam(name: 'reserve_env', description: 'Should we reserve the cloud_env resource?', defaultValue: true)
+        string(name: 'cloud_env', description:'The environment identifier', defaultValue: env.SOCOK8S_ENVNAME)
```
jesusaurus: We will probably want to change the `cloud_env` default value to the label we apply to the new lockable resources for the slots, but I'm confused about how this interacts with the `agent` directive.

jesusaurus:
> @jesusaurus Could you describe a bit more why we need lockable resources (I don't even know what lockable resources are in this context) please?
> To switch to a different project, I thought changing the value in /home/jenkins/.config/openstack/clouds.yaml would be enough. Did I miss anything?

I thought the same thing, but Stephen told me that any jenkins jobs that run in the cloud-ci project must manage their resource usage using the jenkins lockable resources plugin. He also told me that the project is for the CI team, not for general CI, so we may be taking a different approach and creating a new project for our CI to manage its quota.

toabctl:
@stefannica out of interest - why do we need lockable resource? Could you give an example?
@jesusaurus then I guess we can create a new cloud-socok8s-ci project and skip the lockable resource requirement for now?

jesusaurus:
@toabctl yeah, I think a dedicated project is the way to go, but I'm waiting on someone who can create it for me.

Jenkinsfile.integration
```diff
@@ -199,6 +220,9 @@ pipeline {
         stage('Deploy Airship') {
             options {
                 timeout(time: 45, unit: 'MINUTES', activity: true)
+                lock(label: reserve_env == 'true' ? cloud_env:'dummy-resource',
```
kantarun: We need to rebase as tempest step is added so we need to add this lock logic there as well.

jesusaurus:
abandoned in favor of#694

PR #684, opened 2019-08-16
caasp4: Only touch port security on openstack deploy

rhafer:
Only disable Port security when DEPLOYMENT_MECHANISM=openstack. This
should fix libvirt based deployments.

### DIFF
--- a/playbooks/roles/caasp4/tasks/setup_nodes.yml
+++ b/playbooks/roles/caasp4/tasks/setup_nodes.yml
@@ -95,5 +95,7 @@
   args:
     executable: /bin/bash
   changed_when: False
+  when:
+    - socok8s_deployment_mechanism == 'openstack'
   tags:
     - bug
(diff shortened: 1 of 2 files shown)


jsuchome:
Why is it problem with libvirt?

rhafer:
> Why is it problem with libvirt? Shouldn't the search just return empty list of nodes?

It fails because `deploy_on_openstack_internal_network` is undefined in that case. But it would also fail e.g. if `openstack` is not installed.

jsuchome:
I'm surprised it does not fail in more places then :-)

playbooks/roles/caasp4/tasks/setup_nodes.yml
```diff
@@ -14,6 +14,11 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 #
+- name: Get deployment mechanism value
+  set_fact:
+    deployment_mechanism: "{{ lookup('env', 'DEPLOYMENT_MECHANISM') }}"
```
toabctl: Could we set the deployment mechanism variable in defaults/man.yml ? That way, defaults/main.yml documents which variables a user of the role can use to influence the bahaviour of the role.

rhafer: Hm, good point. I'll change that.

PR #685, opened 2019-08-16
caasp4: Allow to override repo mirros

rhafer:
This is currently limited to the libvirt templates. I still need to adapt it for the OpenStack case. (Makes sense to offer choice there as well. E.g. depending on which cloud is used.

### DIFF
--- a/playbooks/roles/caasp4/templates/skuba-terraform-libvirt.tfvars.j2
+++ b/playbooks/roles/caasp4/templates/skuba-terraform-libvirt.tfvars.j2
@@ -13,25 +13,25 @@ masters = {{ master_count }}
 workers = {{ worker_count }}
 
 repositories = {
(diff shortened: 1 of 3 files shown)


playbooks/roles/caasp4/templates/skuba-terraform-libvirt.tfvars.j2
```diff
@@ -25,6 +25,15 @@ repositories = {
   serverapps_updates = "http://download.suse.de/ibs/SUSE/Updates/SLE-Module-Server-Applications/15-SP1/x86_64/update/"
 }
 
+lb_repositories = {
```
toabctl: Do you plan to update the CI example from https://github.com/SUSE/skuba/blob/master/ci/infra/libvirt/terraform.tfvars.json.ci.example , too?

rhafer: It's there now: https://github.com/SUSE/skuba/pull/622

PR #686, opened 2019-08-19
[WIP][DNM] Test whatever more workers means speedier deployment

Itxaka:


Itxaka:
doesnt make any difference once you go over 2 which was the choking point. Seems like its more important to have better nodes than the number of those. Closing!

PR #690, opened 2019-08-19
Doc: Add repos for sle15sp1 deployer setup in kvm environment.

rajdharwadkar:


doc/source/deployment/setup-deployer.rst
```diff
@@ -63,17 +63,27 @@ We currently only allow to deploy master branch from sources.
...
 
      * ansible >= 2.8.3
      * git-core
      * jq
+     * python
      * python3-virtualenv
-     * python3-jmespath
-     * python3-netaddr
-     * python3-openstacksdk
+       SUSEConnect --product sle-module-desktop-applications/15.1/x86_64
```
jhesketh: So this is mixing in commands with a list of required software/packages. Instead should we move this to a section above (or even below) that gives a few preparation steps followed by a zypper one-liner to install all of these?

rajdharwadkar: I think this is more clear information of why we need to enable those repos.  Like  which package belong to which repo.

jhesketh: This section is simply listing the requirements, not how to install them. This is because in some cases the requirements may be installed with pip or from other sources/repositories.

So I think if we have a second section that shows how to install them with the actual zypper commands and SUSEConnect commands (maybe with a common on why SUSEConnect is needed) that would be helpful, but we shouldn't mix the two.

jhesketh:
(I haven't validated that the new requirements list is correct/complete for a KVM env)

rajdharwadkar:
> Actually, this guide is not specific to KVM afaik, so I'm not sure if it's correct to be removing openstacksdk etc. Could you please add the rationale for the dependency changes in the commit message so that we have the history on why they change?

Doc says its for developer :
We currently only allow to deploy master branch from sources.
1. (For developers only) Clone the repository.

rajdharwadkar:
> > Actually, this guide is not specific to KVM afaik, so I'm not sure if it's correct to be removing openstacksdk etc. Could you please add the rationale for the dependency changes in the commit message so that we have the history on why they change?
> 
> Doc says its for developer :
> We currently only allow to deploy master branch from sources.
> 
> 1. (For developers only) Clone the repository.

jhesketh:
> > Actually, this guide is not specific to KVM afaik, so I'm not sure if it's correct to be removing openstacksdk etc. Could you please add the rationale for the dependency changes in the commit message so that we have the history on why they change?
> 
> Doc says its for developer :
> We currently only allow to deploy master branch from sources.
> 
>     1. (For developers only) Clone the repository.

Yes, but developers may still be deploying onto the ECP. Wouldn't they therefore require the openstacksdk, or is that covered by openstackclient?

I guess my point was more to spell out why the requirements have changed.

jesusaurus:
I agree with @jhesketh, this section is for both deployment mechanisms, it's not specific to KVM. And for ECP deployments the deployer setup is 100% automated, so these manual commands are only needed for KVM deployments.

Maybe this should be moved to localhost setup, where these steps are also manually needed for ECP deployments.

rajdharwadkar:
> I agree with @jhesketh, this section is for both deployment mechanisms, it's not specific to KVM. And for ECP deployments the deployer setup is 100% automated, so these manual commands are only needed for KVM deployments.
> 
> Maybe this should be moved to localhost setup, where these steps are also manually needed for ECP deployments.

toabctl:
> And for ECP deployments the deployer setup is 100% automated, so these manual commands are only needed for KVM deployments.

And the plan is to fully automate it for KVM/libvirt setups, too. So I would prefer automation over documentation.

PR #691, opened 2019-08-20
Remove CaaSPv3

jsuchome:
Originally https://github.com/SUSE-Cloud/socok8s/pull/673


Jenkinsfile.integration
```diff
@@ -23,7 +23,6 @@ pipeline {
     parameters {
```
toabctl: - Why has the commit msg (for 42cdde8ecfd6f56cacfd65dd96261fdfe44a17c4) a cherry-pick line? Doesn't make sense imo

run.sh
```diff
@@ -69,13 +69,8 @@ case "$deployment_action" in
         deploy_ses
         ;;
     "deploy_caasp")
-        # this is CaaSP 3
         deploy_caasp
         ;;
-    "deploy_caasp4")
```
toabctl: I would prefer to keep `caasp4` everywhere and just call it in `deploy_caasp4` , too. That way the PR will be pretty simple.

run.sh
```diff
@@ -127,7 +127,7 @@ case "$deployment_action" in
         deploy_airship
         ;;
     "update_openstack")
-        deploy_airship update_airship_osh_site
+        deploy_airship update_airship_osh_site
```
toabctl: unrelated change

Jenkinsfile.integration
jsuchome: Because it was cherry-picked from the older one? I'm not sure I get your point

run.sh
jsuchome: Well it would be less simple now to revert previous changes :-)

Is there a better reason to keep caasp4 everywhere?

Jenkinsfile.integration
jsuchome: removed

PR #692, opened 2019-08-20
Drop venv + remove tox dir + set verbosity to a sane level

Itxaka:


Jenkinsfile.integration
```diff
@@ -112,6 +111,17 @@ pipeline {
                         }
                     }
                 }
+                stage('Clean tox dir') {
```
Itxaka: woops. this should be outside of the parallel!

PR #697, opened 2019-08-22
Notify Rocketchat channel on cleanup failure

Itxaka:
As we still have some issues when cleaning up (openstack failures,
resources blocked, disconnecting CI agents, etc...) sometimes
the cleanup will fail and there will be resources left over.

This introduces a very small role that will notify our channel on
failures to clean up so we can go and fix them manually until we
have a more stable CI/environment/tasks

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -267,8 +267,14 @@ pipeline {
                 // Do not leave the ssh-agent running
                 sh "pkill -f 'ssh-agent -a /tmp/${env.SOCOK8S_ENVNAME}'"
                 if (fileExists("/tmp/${env.SOCOK8S_ENVNAME}.needcleanup")) {
-                    sh './run.sh teardown'
-                    sh 'rm /tmp/${SOCOK8S_ENVNAME}.needcleanup'
+                    try {
+                        sh './run.sh teardown'
+                        sh 'rm /tmp/${SOCOK8S_ENVNAME}.needcleanup'
+                    } catch (e) {
+                        sh 'ansible-playbook playbooks/generic-notify_failure_rocket.yml -v'
+                        // manually set the build to failure if we could not cleanup resources
+                        currentBuild.result = 'FAILURE'
+                    }
                 }
             }
         }
--- a/playbooks/generic-notify_failure_rocket.yml
+++ b/playbooks/generic-notify_failure_rocket.yml
@@ -1,0 +1,21 @@
+---
+#
+# (c) Copyright 2019 SUSE LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+
+- hosts: localhost
+  gather_facts: false
+  roles:
+    - role: notify-failure-rocket
--- a/playbooks/roles/notify-failure-rocket/defaults/main.yml
+++ b/playbooks/roles/notify-failure-rocket/defaults/main.yml
@@ -1,0 +1,22 @@
+#
+# (c) Copyright 2019 SUSE LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+---
+
+rc_domain: "chat.suse.de"
+rc_protocol: "https"
+rc_url: "{{ rc_protocol }}://{{ rc_domain }}"
+rc_auth_info_creds: "https://gitlab.suse.de/cloud-qe/ardana-bm-info/raw/master/rc.yml"
+rc_channel: "cloud-airship"
--- a/playbooks/roles/notify-failure-rocket/tasks/main.yml
+++ b/playbooks/roles/notify-failure-rocket/tasks/main.yml
@@ -1,0 +1,44 @@
+#
+# (c) Copyright 2019 SUSE LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+---
+
+- name: Load rocketchat auth variables
+  uri:
+    url: "{{ rc_auth_info_creds }}"
+    return_content: yes
+  register: rocket_vars
+  no_log: True
+
+- name: Set rc_token var
+  set_fact:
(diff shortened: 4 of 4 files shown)


jsuchome:
As this is purely CI related, and very specific to our development (#cloud-airship channel in our rocketchat instance), I'm not sure it should be exposed by `run.sh` ...

evrardjp:
How can we try this before getting spammed?

playbooks/roles/notify-failure-rocket/tasks/main.yml
```diff
@@ -0,0 +1,42 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+---
+
+- name: Clone ardana-bm-info repository
```
evrardjp: What is this?

playbooks/roles/notify-failure-rocket/tasks/main.yml
```diff
@@ -0,0 +1,42 @@
...
+
+- name: Notify RocketChat of the failure to clean up
+  rocketchat:
+    domain: "{{ rc_domain }}"
+    token: "{{ rc_token }}"
+    username: "Jenkins"
+    channel: "{{ rc_channel }}"
+    validate_certs: no
+    protocol: "{{ rc_protocol }}"
+    attachments:
```
evrardjp: Why not simply using "msg?"
```msg: "Failure in Job teardown for {{ lookup('env','BRANCH_NAME') }}. See log: {{ lookup('env','RUN_DISPLAY_URL') }}" ```

File: playbooks/generic-notify_failure_rocket.yml [hunk 3]
```diff
@@ -0,0 +1,27 @@
...
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+
+- hosts: localhost
+  gather_facts: false
+  vars_files:
+    - "{{ playbook_dir }}/../vars/common-vars.yml"
+  pre_tasks:
```
evrardjp: Do we really need this, if it's only used in the default var? We could simply use the lookup there in the defaults.

playbooks/generic-notify_failure_rocket.yml
```diff
@@ -0,0 +1,27 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+
+- hosts: localhost
+  gather_facts: false
+  vars_files:
```
evrardjp: What do we need from common-vars?

playbooks/roles/notify-failure-rocket/defaults/main.yml
```diff
@@ -0,0 +1,24 @@
...
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+---
+
+rc_domain: "chat.suse.de"
+rc_protocol: "https"
+rc_url: "{{ rc_protocol }}://{{ rc_domain }}"
+rc_token: ""
```
evrardjp: can we leave this undefined, so that the role fails when this is not defined, instead of silently not working?

playbooks/roles/notify-failure-rocket/tasks/main.yml
```diff
@@ -0,0 +1,42 @@
...
+
+- name: Clone ardana-bm-info repository
+  git:
+    repo: "{{ rc_auth_info_git_url }}"
+    dest: "{{ rc_auth_info_dir }}"
+    version: "master"
+    force: yes
+
+- name: Load rocketchat auth variables
+  include_vars: "{{ rc_auth_info_dir }}/rc.yml"
```
evrardjp: can't we just ``set_fact`` ``rc_token`` by fetching it from a url?
``https://gitlab.suse.de/cloud-qe/ardana-bm-info/raw/master/rc.yml`` seems wide open to me behind the vpn.

playbooks/generic-notify_failure_rocket.yml
Itxaka: agh, nothing anymore. I sed it for the workspace name but we can use the jenkins envs for that.

playbooks/generic-notify_failure_rocket.yml
Itxaka: :+1:

playbooks/roles/notify-failure-rocket/defaults/main.yml
Itxaka: mmmh, that may hit us due to lazy_loading or instant loading. I dont remember which one does the lazy one (include_role vs import_role) but yeah we could do that.

playbooks/roles/notify-failure-rocket/tasks/main.yml
Itxaka: cloning repo to access the rc vars

playbooks/roles/notify-failure-rocket/tasks/main.yml
Itxaka: We could I guess.

playbooks/roles/notify-failure-rocket/tasks/main.yml
Itxaka: Looks nicer!
There is a lot of information there as to have everything in one line. Using attachment not only shows prominently in the chat but has the link embedded and the information is readable.

evrardjp: Fine for me then :)

playbooks/roles/notify-failure-rocket/tasks/main.yml
evrardjp: or simply use the lookup in the defaults.

Assuming you then make the arguments of the rocket chat call part of the default, then it becomes simple to reuse this across the whole org.

Jenkinsfile.integration
```diff
@@ -267,8 +267,13 @@ pipeline {
...
                 sh "pkill -f 'ssh-agent -a /tmp/${env.SOCOK8S_ENVNAME}'"
                 if (fileExists("/tmp/${env.SOCOK8S_ENVNAME}.needcleanup")) {
-                    sh './run.sh teardown'
-                    sh 'rm /tmp/${SOCOK8S_ENVNAME}.needcleanup'
+                    try {
+                        sh './run.sh teardown'
+                    } catch (e) {
+                        sh 'ansible-playbook playbooks/generic-notify_failure_rocket.yml -v'
+                    } finally {
+                        sh 'rm /tmp/${SOCOK8S_ENVNAME}.needcleanup'
```
jhesketh: I'd rather this stick around if the teardown fails. It gives us an alternate list of what might need cleaning up rather than rocketchat in case we miss messages.

(and worst case if we forget to remove the .needcleanup file after we have manually resolved something it's just an empty file in /tmp/).

evrardjp: :+1:

playbooks/roles/notify-failure-rocket/defaults/main.yml
evrardjp: It's not necessary.

Jenkinsfile.integration
Itxaka: good point. That also remind me that we need to set the build status to failure as we are caching an error!

playbooks/roles/notify-failure-rocket/defaults/main.yml
Itxaka: woops, its now removed

Jenkinsfile.integration
evrardjp: Agreed!

playbooks/roles/notify-failure-rocket/defaults/main.yml
```diff
@@ -0,0 +1,22 @@
...
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+---
+
+rc_domain: "chat.suse.de"
+rc_protocol: "https"
+rc_url: "{{ rc_protocol }}://{{ rc_domain }}"
+rc_auth_info_creds: "https://gitlab.suse.de/cloud-qe/ardana-bm-info/raw/master/rc.yml"
```
AbelNavarro: only protected by our VPN. I guess it's good for now.

PR #698, opened 2019-08-22
update osh-infra submodule to latest

Itxaka:
This fixes the pod logs gathering, which is failing due to
kubectl >=1.15 not having the "--show-all" flag anymore

Updated to I4ed3624737af4d5c90b1b5de451a0a0b75a5eda1


Itxaka:
I will cnacel the job as soon as there is running pods, so it can trigger the post action of gathering logs. Otherwise there is no other way of testing this correctly, so the CI for this will be red :=)

PR #700, opened 2019-08-22
Add log parser to CI

Itxaka:
Adds log parser step and rules for log parsing on CI for ease of
access to important lines in the CI jobs

### DIFF
--- a/log-parser-plugin-rules.txt
+++ b/log-parser-plugin-rules.txt
@@ -1,0 +1,66 @@
+start /\sPLAYBOOK:\s/
+start /^\./run.sh/
+
+info /\+\sansible-playbook\s/
+info /\sTASK\s\[/
+info /skipped/
+
+# set the retries as warning so we can check them faster
+warning /\sFAILED - RETRYING:/
+
+# not errors
+# this one is the task state at the end
+ok /failed=0/
+# this one comes from the task output
+ok /failed: false/
+# same as above
+ok /'failed': False/
+# this is part of the task options
+ok /failed_when_result:/
+# part of the systemd output
+ok /ignore_errors=no/
+# same systemd
+ok /FailureAction:/
+# same systemd
+ok /SystemCallErrorNumber:/
+# same systemd
+ok /StandardError: inherit/
+# same systemd
+ok /OnFailureJobMode:/
+# same systemd
+ok /Restart: on-failure/
+# description of a package on package install
+ok /<description>com_err/
+# same
+ok /name="libcom_err2"/
+# this is the output from the node upgrade
+ok /Failed to connect to bus: No such file or directory/
+# same as above
+ok /Failed to look up reference node - returned empty list/
+# same as above
+ok /dracut-install: ERROR:/
+# ignore package name downloads with error in the name
+ok /<download url="/
+# ignore package with error in the name
+ok /.*<solvable.*/
+# ignore package descriptions with error in the name
+ok /.*<description>.*/
+ok /.*</description>.*/
+# ignore %post failures in package installs
+ok /warning: %post/
+# from kernel install
+ok /Failed to import \/etc\/uefi\/certs\/91A3B0B5.crt/
+# output from ses salt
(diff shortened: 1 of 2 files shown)


evrardjp:
Interesting.

Ofc if we had clearer steps that would be less verbose that wouldn't be required in the first place. Ahem :)

evrardjp:
I like the idea though. :+1:

Itxaka:
Rules are now clean enough to only show real issues and point to the proper headers for playbooks and tasks

aspiers:
Very nice @Itxaka!!  Would it be possible to take this approach with other jobs and update https://github.com/SUSE/cloud/blob/master/docs/CI/keep-jenkins-green/jobs/openstack-mkcloud-log-parser-test.md accordingly?  /cc @stefannica @jdsn

Itxaka:
> I am not too fond of having no_log true on packages installs, it makes it harder to know what was done and will make the debugging harder. Except that I like this. Is it possible to improve the ansible callback so that we don't need to do this?

The problem is that the zypper module logs a *lot* of xml which causes a lot of false positives, and keeping a list of all the individual lines to ignore is a very time consuming task :(

I mean, do we really want to track the descriptions of the packages to avoid false positives?

Not sure if we could probably sent patches upstream to allow for more control over the stdout of the module so it doesnt log that much.

Itxaka:
> Very nice @Itxaka!! Would it be possible to take this approach with other jobs and update SUSE/cloud:docs/CI/keep-jenkins-green/jobs/openstack-mkcloud-log-parser-test.md@master accordingly? /cc @stefannica @jdsn

Sorry @aspiers I dont understand what do you mean by this? Doesn't other jobs also use the same approach? They dont have it on the jenkinsfile I guess, maybe that's what you mean?

aspiers:
[**@Itxaka**](https://github.com/Itxaka) commented on [August 28, 2019 11:51 AM](https://github.com/SUSE-Cloud/socok8s/pull/700#issuecomment-525690968 "2019-08-28T10:51:02Z - Replied by Github Reply Comments"):
> Sorry @aspiers I dont understand what do you mean by this? Doesn't other jobs also use the same approach?

I haven't seen any use it except the mkcloud job which I just linked to, however I haven't been following recent developments in CI closely so maybe I missed something.  If you know of others then please can you provide URLs?

> They dont have it on the jenkinsfile I guess, maybe that's what you mean?

Other than `mkcloud` I haven't seen any others configure it at *all*, let alone via `Jenkinsfile`.  The one I set up was certainly the first, because I had to ask @jdsn to install the Jenkins plugin at the time.

Itxaka:
> If we accept the CI callback plugin, we can remove the no_log: true

+1, I will rework this to remove those so when the callback is merged we dont need to change this

log-parser-plugin-rules.txt
```diff
@@ -0,0 +1,66 @@
...
+# shipyard output
+ok /Errors: 0,/
+# expected on ansible machine reboots
+ok /\[WARNING]: Failed to reset connection:Control socket/
+# ansible var set in playbooks/roles/airship-deploy-osh/tasks/main.yml:23
+ok /no_revision_error_msg/
+
+error /(?i)error/
+error /(?i)fail(ure|ed)/
+error /(?i)fatal/
```
kantarun: Wow, that's an extensive list. You must have zeroed-in logs for sometime to capture all these patterns. :)

log-parser-plugin-rules.txt
```diff
@@ -0,0 +1,66 @@
+start /\sPLAYBOOK:\s/
+start /^\./run.sh/
+
+info /\+\sansible-playbook\s/
+info /\sTASK\s\[/
+info /skipped/
+
+# set the retries as warning so we can check them faster
+warning /\sFAILED - RETRYING:/
```
evrardjp: This will be a lot of warnings, but it will bring stats/trends on how we deal with retries, so I like this.

PR #701, opened 2019-08-22
Script for setting up single-node k8s cluster and clusterctl

kantarun:
Script sets a single node minikube based k8s cluster on SLES-15-SP1.
Clusterctl with openstack provider is also added in this setup.

Work done as part of SOC-10087


jhesketh:
This is really just a place to chuck some quick notes that are linked to from various cards. There are future patches that will actually create a cluster etc, yes.

PR #705, opened 2019-08-26
Allow setting pod resource limits for osh

Itxaka:
All osh charts have some limits in place for all its pods but currently
we cannot enable those

This patch adds a way of setting the resource limits enabled or disabled
for all the pods in each osh chart via the common vars

COMPONENT_rabbitmq values are disabled for now as the current limits seem
too low and containers are killed due to OOMKilled

### DIFF
--- a/site/soc/software/charts/osh/openstack-ceph-config/ceph-config.yaml
+++ b/site/soc/software/charts/osh/openstack-ceph-config/ceph-config.yaml
@@ -53,4 +53,7 @@ data:
       job_namespace_client_key: false
       storageclass_cephfs: false
       storageclass_rbd: false
+    pod:
+      resources:
+        enabled: {{ openstack_helm_pod_resources_enabled['ceph'] }}
 ...
--- a/site/soc/software/charts/osh/openstack-cinder/cinder.yaml
+++ b/site/soc/software/charts/osh/openstack-cinder/cinder.yaml
@@ -50,6 +50,8 @@ data:
     timeout: {{ test_timeout }}
   values:
     pod:
+      resources:
+        enabled: {{ openstack_helm_pod_resources_enabled['cinder'] }}
       replicas:
         api: 1
         volume: 1
--- a/site/soc/software/charts/osh/openstack-cinder/rabbitmq.yaml
+++ b/site/soc/software/charts/osh/openstack-cinder/rabbitmq.yaml
@@ -23,6 +23,8 @@ data:
     timeout: {{ test_timeout }}
   values:
     pod:
+      resources:
+        enabled: {{ openstack_helm_pod_resources_enabled['cinder_rabbitmq'] }}
       replicas:
         server: 1
     monitoring:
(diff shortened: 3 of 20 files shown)


Itxaka:
> I understand where this comes from, but do we really need ansible to template a yaml file, which is templated by airship tooling for deployment? Isn't there a smarter way?

Something to bring up in the east+west meeting? tbh Im not sure how to, we need to be able to customize/override that site/soc so that ansible template step is unavoidable and then whatever is being doing under airship its another thing completely.

But I digress, That seems to be something to discuss with the whole team. Maybe airship 2.0 is magic and solves this? (no it wont, Im sure of that :P)

Itxaka:
Fake CI output, it jsut timed out.

Itxaka:
Latest CI is now REAL PASSING(TM) ;P

evrardjp:
> Something to bring up in the east+west meeting?

Definitely.

site/soc/software/charts/osh/openstack-ceph-config/ceph-config.yaml
```diff
@@ -53,4 +53,7 @@ data:
       job_namespace_client_key: false
       storageclass_cephfs: false
       storageclass_rbd: false
+    pod:
+      resources:
+        enabled: {{ openstack_helm_pod_resources_enabled['ceph'] }}
```
AbelNavarro: I fail to see how just enabling pod resources seem to be working. I was expecting something requesting specific [CPU/Memory/Disk](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#how-pods-with-resource-requests-are-scheduled). What am I missing here?

Itxaka: Yeah so...where to start :D


Airship uses the openstack-helm charts to deploy openstack rigth? So those charts already have limits/requests in place, but they need to be activated by setting the chart value `pod.resources.enabled` to `true`, otherwise it wont apply them on chart generation.

Airship uses this files to pass values to the osh charts, thus having a place for easy overriding. Adding this values here means that they will be applied on the osh charts when building and deploying them. See here[0] for where the limits/requests are in the upstream chart and here[1] for how they are applied depending on the value of `pod.resources.enabled` and here[2] the actual function from htk that expands those values into the k8s yaml

[0] https://github.com/openstack/openstack-helm-infra/blob/master/ceph-mon/values.yaml#L102
[1] https://github.com/openstack/openstack-helm-infra/blob/master/ceph-mon/templates/daemonset-mon.yaml#L123
[2] https://github.com/openstack/openstack-helm-infra/blob/master/helm-toolkit/templates/snippets/_kubernetes_resources.tpl

AbelNavarro: thanks, much clearer now.

Itxaka: > thanks, much clearer now.

are you sure? re-reading makes it sound less clearer to me :rofl:

PR #713, opened 2019-08-28
Enable tempest by default (SOC-10120)

Itxaka:
Enable tempest run for each CI job by default

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -23,7 +23,6 @@ pipeline {
     parameters {
         /* first value in the list is the default */
         choice(choices: ['airship', 'osh'], description: 'Which deployment mechanism?', name: 'deployment')
-        booleanParam(name: 'run_tempest', defaultValue: false, description: 'Run tempest tests?')
         choice(choices: ['smoke', 'all'], description: 'Run smoke or all tempest tests?', name: 'tempest_test_type')
         choice(choices: ['ses_aio', 'ses_rook'], description: 'Which ses version to use?', name: 'ses_version')
     }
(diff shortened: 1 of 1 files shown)


Jenkinsfile.integration
```diff
@@ -23,7 +23,6 @@ pipeline {
     parameters {
         /* first value in the list is the default */
         choice(choices: ['airship', 'osh'], description: 'Which deployment mechanism?', name: 'deployment')
-        booleanParam(name: 'run_tempest', defaultValue: false, description: 'Run tempest tests?')
```
AbelNavarro: I'd prefer this would remain configurable, but the default value to True. Maybe in the future we want to skip these tests for some scenarios.

Itxaka: Card AC says that it should be removed so....I also think it would be good to have it just in case, for manual tests that do not necessarily need a full tempest run.

Itxaka:
https://github.com/SUSE-Cloud/automation/pull/3660

Jenkinsfile.integration
jhesketh: If it makes sense to leave it, then lets do so and change the card. I'm happy either way.

Itxaka: I rather not have to bring an extra option around...plus I would need to change both PRs :P

Itxaka:
rebased so the tempest_test_type param is passed to the playbook

Jenkinsfile.integration
AbelNavarro: I guess it's ok, we can always bring that back if needed.

evrardjp: Let's solemnly vow we'll never bring that back, and always run tempest!

Jenkinsfile.integration
```diff
@@ -218,7 +217,6 @@ pipeline {
             }
             when {
                 allOf {
-                    expression { env.run_tempest == 'true' }
```
evrardjp: We can probably remove the conditional below: should we test osh, we would still need to use tempest.

Itxaka: good point, although it may not be ready yet for osh, which would require more changes to it to account for the differences? Let me try to test it if engcloud permits

evrardjp: Yes it's probably broken, but it gives us an incentive to fix :)

Itxaka: good point, done!

PR #720, opened 2019-08-29
Example clusterctl using ansible instead of bash

Itxaka:


Itxaka:
Not tested! Just an example. I think we should go for more ansible instead of more bash, even for research items as its really simple to do for smaller tasks and much easier to expand, i.e. expanding this to support tumbleweed is much easier than in bash

AbelNavarro:
that would have my +1 if tested.

PR #721, opened 2019-08-29
Allow keeping the CI machines after the run

Itxaka:
Allows to keep the instances up for 3 hours for troubleshooting them

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -25,6 +25,7 @@ pipeline {
         choice(choices: ['airship', 'osh'], description: 'Which deployment mechanism?', name: 'deployment')
         choice(choices: ['smoke', 'all'], description: 'Run smoke or all tempest tests?', name: 'tempest_test_type')
         choice(choices: ['ses_aio', 'ses_rook'], description: 'Which ses version to use?', name: 'ses_version')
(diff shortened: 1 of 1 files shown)


jsuchome:
If the option is set, it affects both failed and successful runs, right? That's what I think it should do.

Itxaka:
unfortunately this did not work, checking out the params to see if there is something wrong here....

Itxaka:
now its working, I was comparing a bool to a string....

AbelNavarro:
I think this was asked before, but you'll need to pass that parameter when triggering the build to be effective. So you either set the CI to always keeps the jobs 3 hours or manually re-trigger a build with a request to keep it. What would be the approach here?

Itxaka:
> I think this was asked before, but you'll need to pass that  parameter when triggering the build to be effective. So you either set  the CI to always keeps the jobs 3 hours or manually re-trigger a build  with a request to keep it. What would be the approach here?

So for now this is done on demand, by triggering a manual job with the option enabled.

In the future @jsuchome proposed having it check for a file so we can go into the jenkins instance and create that file for a running job, and it would keep it running.

In the future, if we get to a more stable green job, we could do this for failed instances only, but I think that approach is dangerous as any broken patch would block a lot of jobs.

evrardjp:
I like the future approach of a tempfile in the PR too. It's simple. And a bot can automatically mark the PR as do not merge if it includes said file :)

PR #724, opened 2019-08-30
Reduce volume size for rabbit

Itxaka:
While 5Gb migth be a good size for the default db volumes, for
rabbitmq is way too much.

It doesnt help that airship creates a rabbitmq container per component
so we end up with several 5Gb volumes around for a lot of containers.

Instead set it to a more reasonable size (1Gb) which should allow for
rabbitmq to store to persist to disk with no issues


jsuchome:
@rajdharwadkar did you test with different values when you actually raised it to 5Gb?

PR #727, opened 2019-08-30
Skip error conditions for Airship wait action (SOC-10375)

AbelNavarro:
When checking the status of 'airship describe' in the 'Wait for update software' action sometimes we receive temporary errors (503,504) that make the task to stop. Since we consider these to be spurious errors we are going to skip those.

Reference: SOC-10375

### DIFF
--- a/playbooks/roles/airship-deploy-osh/tasks/main.yml
+++ b/playbooks/roles/airship-deploy-osh/tasks/main.yml
@@ -184,7 +184,10 @@
         OS_PASSWORD: "{{ lookup('password', secrets_location + '/ucp_shipyard_keystone_password ' + password_opts) }}"
       register: shipyard_desc_action
       failed_when: "'failed' in shipyard_desc_action.stdout"
-      until: shipyard_desc_action.stdout.find('Processing') < 0 and shipyard_desc_action.stdout.find('running') < 0
+      until:
+        - shipyard_desc_action.stdout.find('Processing') == -1
+        - shipyard_desc_action.stdout.find('running') == -1
+        - shipyard_desc_action.stdout.find('Error') == -1
       retries: "{{ airship_deploy_openstack_timeout | int * 2 }}"
       delay: 30
       tags:


playbooks/roles/airship-deploy-osh/tasks/main.yml
```diff
@@ -186,7 +186,9 @@
         SHIPYARD_IMAGE: "{{ shipyard_image }}"
         OS_PASSWORD: "{{ lookup('password', secrets_location + '/ucp_shipyard_keystone_password ' + password_opts) }}"
       register: shipyard_desc_action
-      failed_when: "'failed' in shipyard_desc_action.stdout"
+      failed_when: >
+          "'failed' in shipyard_desc_action.stdout" or
+          "'Error' in·shipyard_desc_action.stdout"
```
jsuchome: _in·shipyard_desc_action.stdout_ ? What's after _in_ ?

AbelNavarro: oops, that's 'vi' showing blank spaces. Should have used emacs. Fixing.

Itxaka: do we want to fail out when this happens? because sometimes is just a retryable error like the temporary 503 that we get..

For that we should modify the until I guess

AbelNavarro: I think you are right, but it's gonna look ugly to check for error there. Fixing.

evrardjp: > For that we should modify the until I guess

That until raises my eyebrows, there is no test to find in the stdout that things were fine? Just the absence of "running" or "Processing"? That's weird.

evrardjp: > For that we should modify the until I guess

That until raises my eyebrows, there is no test to find in the stdout that things were fine? Just the absence of "running" or "Processing"? That's weird.

evrardjp: The until is just to tell you when to stop, failure or not.
That until raises my eyebrows, there is no test to find in the stdout that things were fine? Just the absence of "running" or "Processing"? That's weird.

playbooks/roles/airship-deploy-osh/tasks/main.yml
```diff
@@ -187,7 +187,10 @@
         OS_PASSWORD: "{{ lookup('password', secrets_location + '/ucp_shipyard_keystone_password ' + password_opts) }}"
       register: shipyard_desc_action
       failed_when: "'failed' in shipyard_desc_action.stdout"
-      until: shipyard_desc_action.stdout.find('Processing') < 0 and shipyard_desc_action.stdout.find('running') < 0
+      until: >
+          shipyard_desc_action.stdout.find('Processing') < 0 and
+          shipyard_desc_action.stdout.find('running') < 0 and
+          shipyard_desc_action.stdout.find('Error') < 0
```
evrardjp: for clarity you can replace the until with:

```
until:
  - shipyard_desc_action.stdout.find('Processing') == -1
  - shipyard_desc_action.stdout.find('running') == -1
  - shipyard_desc_action.stdout.find('Error') == -1
```

or add "not" on each condition and make it search for `> 0`.

So you mean it will continue until it doesn't find errors/running/processing? Is that really what you want?

evrardjp: For me we should be smarter when parsing stdout.
```
Lifecycle:             Complete
(...)
Dag Status:            success
```
Is the success.

So until condition is find Lifecycle: Complete and dag status is success OR (insert main error you want to stop the deploy for).

AbelNavarro: I think it's a bit more complex than checking for `Lifecycle: Complete`. That `Lifecycle` field could be `Pending, Complete, Failed, Paused, Unknown, null`. If we stop at `Complete` we miss the other possible estates. The upstream code seems to be waiting for `!Pending` (https://github.com/airshipit/shipyard/blob/master/tools/execute_shipyard_action.sh#L112).

On top of that we get those occasional errors (503, 504) in stdout, and we want to skip those. Since `Error` is not used in that command output seems like a good fit to catch those errors.

Probably doesn't look clean but it's quite precise and short to do what we want. I'm also adding the cleanup you propose.

evrardjp: I have the impression we are in agreement with the first paragraph.
Thanks for listing the states.

What's wrong with the following?

```
failed_when:
  - shipyard_desc_action.stdout.find("Lifecycle: Complete") == -1
until:
  - _run.stdout.find("Complete") > 0 or _run.stdout.find("Failed") > 0
```

If complete is not found, it will continue, except when Failed (a known error) is reached. 503 will get retried, so will Pending, Paused, Unknown, and null, which sounds all fair to me.

It's by far easier to grasp.

Itxaka:
Cant we push to have this asap? Currently any 503 makes it fail :(

playbooks/roles/airship-deploy-osh/tasks/main.yml
AbelNavarro: Your approach looks cleaner, but it's not complete because it doesn't cover all the other estates. In addition to that, `Complete` doesn't mean successful, because you can have failed actions (https://github.com/airshipit/shipyard/blob/master/tools/execute_shipyard_action.sh#L138). Even if it's dirty, I'd keep it like it is right now.

Itxaka:
Reading the code, Im not sure if this will fix the error when finding a 503. Doesnt look like. But it still improves the current state so Im good with that :)

playbooks/roles/airship-deploy-osh/tasks/main.yml
evrardjp: But that's the whole point of the change in the until. We don't need to care about all the states.
We only care about a finished state, which is only happening on Complete or Failed.
The rest is transient state, which we don't care. Or did I get that wrong?

To cover the case when a result is Complete: but not Failed while failing (this is terribly wrong by shipyard btw), we could add an extra line in the failed_when. This becomes.

```
failed_when: >
  shipyard_desc_action.stdout.find("Lifecycle: Complete") == -1 or
  (shipyard_desc_action.stdout.find("Lifecycle: Complete") != -1 and
  shipyard_desc_action.rc != 0)
until:
  - _run.stdout.find("Lifecycle: Complete") > 0 or _run.stdout.find("Lifecycle: Failed") > 0
```

evrardjp: I don't see a reason to keep doing dirty hacks.

AbelNavarro: I think there would be a problem with that approach by checking the `rc` variable (return code?) because I'm not sure if there are failed steps that return code will change. I think probably we would instead need to search for the `failed` string there.

About that `until` why that use of the `_run` variable (I haven't seen it in other parts of our code). Probably it's equivalent to `shipyard_desc_action`.

AbelNavarro:
This was already being addressed in https://github.com/SUSE-Cloud/socok8s/pull/476 . Closing this one.

PR #730, opened 2019-09-02
Run LVM2 on the caasp nodes

jsuchome:
So that rook-ceph can correctly activate the volumes.
See also bsc#1148954

### DIFF
--- a/playbooks/roles/caasp4/templates/skuba-terraform-libvirt.tfvars.j2
+++ b/playbooks/roles/caasp4/templates/skuba-terraform-libvirt.tfvars.j2
@@ -39,6 +39,7 @@ packages = [
   "-kernel-default-base",
   "ca-certificates-suse",
   "ceph-common",
+  "lvm2"
 ]
 
 authorized_keys = [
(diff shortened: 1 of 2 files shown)


playbooks/generic-deploy_ses_rook.yml
```diff
@@ -34,6 +34,14 @@
       tags:
         - bug
         - skip_ansible_lint
+- hosts: caasp-workers
```
Itxaka: comment pointing to why or maybe the issue number/link?

playbooks/generic-deploy_ses_rook.yml
```diff
@@ -34,6 +34,14 @@
...
+  gather_facts: no
+    - name: Start LVM service
```
evrardjp: tasks:

playbooks/generic-deploy_ses_rook.yml
evrardjp: Good idea.

jsuchome:
Removed the commit that manually started the services. It seems ceph does it by itself so no need for us to do the work

PR #731, opened 2019-09-02
CI: Switch SES deployment to rook

jsuchome:
We should wait at least for https://github.com/SUSE-Cloud/socok8s/pull/730

This is only the CI, no code cleanup yet

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -24,7 +24,6 @@ pipeline {
         /* first value in the list is the default */
         choice(choices: ['airship', 'osh'], description: 'Which deployment mechanism?', name: 'deployment')
         choice(choices: ['smoke', 'all'], description: 'Run smoke or all tempest tests?', name: 'tempest_test_type')
-        choice(choices: ['ses_aio', 'ses_rook'], description: 'Which ses version to use?', name: 'ses_version')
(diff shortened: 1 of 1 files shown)


Itxaka:
just rebased to include the better flavors, lets see if we don not hit a 503 this time

Itxaka:
Green to deploy but a lot of tempest failures with error 503, we need to check the pod logs to see if something went wrong or what.

jhesketh:
The west squad also made a good point that we want to make sure SES6 is working well for the KVM type deployment before completely removing SES5.

jsuchome:
@jhesketh Good point but it actually is related to https://github.com/SUSE-Cloud/socok8s/pull/733 which removes the code path to deploy SES5. 
This PR only changes what is done in CI which does not affect KVM users anyway.

Itxaka:
job retriggered to check if its a consistent failure

jsuchome:
(I still believe this one could be merged, as it affects only CI, and still keeps non-rook deployment path. Assuming it passes of course)

evrardjp:
Love to idea, sad that it's red in Jenkins.

jsuchome:
We need https://github.com/SUSE-Cloud/socok8s/pull/735 before, registry.suse.de does not seem to contain right images.

PR #732, opened 2019-09-02
Parse subunit output from tempest into junit

Itxaka:
Currently we have no way of tracking the output of tempest in
the CI.

This patch adds a new playbook to extract the logs from the tempest
container, transform them into junit files and parse them on CI
for easy tracking of the tempest tests.

Just setting SOCOK8S_TEMPEST_PARSE_RESULTS=True should be enough to get
the output file with the tempest results

What it does its that it creates a temporal container with the tempest
pvc attached in order to extract the data from the tempest run, then
brings it to ansible_host, creates a temporal virtualenv to install
the tools needed for transofrming the data, and then transforms it
into a junit file so the CI can understand the output.

This requires a small change in our tempest deployment, which creates
a workspace at /var/lib/tempest/data, which is where the pvc is mounted,
in order to save the results. This also means that you can run tempest
several times and all the results will be in the same place, accesible
by creating a container and mounting the pvc.

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -47,6 +47,7 @@ pipeline {
         USE_ARA = "True"
         ARA_DIR = "$WORKSPACE"
         SOCOK8S_DEPLOY_DSTAT = "YES"
+        SOCOK8S_TEMPEST_SUBUNIT_OUTPUT = "True"
     }
 
     stages {
@@ -226,6 +227,18 @@ pipeline {
                   export AIRSHIP_TEMPEST_LOG_STDOUT=True
                   ./run.sh test
                 """
+            }
+            post {
+                success {
+                    script {
+                        try {
+                            sh "./run.sh parse_tempest_ci"
+                            junit testResults: 'tempest.xml', allowEmptyResults: true, keepLongStdio: true
+                        } catch(e) {
+                            echo "Could not parse tempest results"
+                        }
+                    }
+                }
             }
         }
     }
--- a/playbooks/generic-tempest-output-to-junit.yml
+++ b/playbooks/generic-tempest-output-to-junit.yml
@@ -1,0 +1,83 @@
+---
+- hosts: soc-deployer
+  gather_facts: yes
+  tasks:
+    - name: Run container to attach tempest pvc
+      shell: |
+        set pipefail -o
+        cat <<EOF | kubectl -n openstack apply -f -
+        apiVersion: v1
+        kind: Pod
+        metadata:
+          name: tempest-data
+        spec:
+          containers:
+          - name: tempest-data
+            image: opensuse/leap
+            command: ["/bin/bash", "-ec", "while :; do echo '.'; sleep 5 ; done"]
+            volumeMounts:
+            - name: mypvc
+              mountPath: /data
+          volumes:
+          - name: mypvc
+            persistentVolumeClaim:
+              claimName: pvc-tempest
+        EOF
+      register: tempest_data_pod_create
+      ignore_errors: true
+      tags:
+        - skip_ansible_lint
+    - name: Wait until data pod api pod is deployed
+      command: "kubectl get pod -n openstack tempest-data -o jsonpath='{.status.containerStatuses[].ready}'"
+      register: tempest_data_pod_status
+      until: tempest_data_pod_status.stdout == "true"
+      retries: 60
+      delay: 2
+      changed_when: False
+      when: tempest_data_pod_create.rc == 0
+    - name: Cat the tempest output into a local file
+      shell: "kubectl -n openstack exec -i tempest-data -- cat /data/.stestr/0 > {{ ansible_env.HOME }}/tempest.subunit"
+      register: _tempest_output
+      when: tempest_data_pod_status.rc == 0
+    - name: Fetch the tempest output
+      fetch:
+        dest: "{{ playbook_dir }}/../tempest.subunit"
+        src: "{{ ansible_env.HOME }}/tempest.subunit"
+        flat: yes
+      when:
+        - _tempest_output.rc == 0
+    - name: Delete the data pod
+      command: kubectl -n openstack delete pod tempest-data --grace-period=5
+      when: tempest_data_pod_create.rc == 0
+
+- hosts: localhost
+  gather_facts: no
+  tasks:
+    - name: Check that previous task were successfull
+      stat:
+        path: "{{ playbook_dir }}/../tempest.subunit"
+      register: _tempest_file
+    - name: Install required packages in a virtualenv
+      pip:
+        name:
+          - junitxml
+          - python-subunit
+        virtualenv: "{{ playbook_dir }}/../junitvirtualenv"
+        virtualenv_python: python2  # junit stuff does not work under py3 for some reason, needs more investigation
+      when: _tempest_file.stat.exists
+    # This task can output rc 1 even if it succeeds!
+    # subunit2junitxml will forward the stream results to the exit code, so if tests fail, exit code will be 1
+    - name: Transform subunit into junit
+      shell: |
+        source {{ playbook_dir }}/../junitvirtualenv/bin/activate
+        cat {{ playbook_dir }}/../tempest.subunit | subunit-1to2 | subunit2junitxml > {{ playbook_dir }}/../tempest.xml
+      args:
+        executable: /bin/bash
+      failed_when: False
+      tags:
+        - skip_ansible_lint
+      when: _tempest_file.stat.exists
+    - name: Remove virtualenv
+      file:
+        path: "{{ playbook_dir }}/../junitvirtualenv"
+        state: absent
--- a/playbooks/roles/airship-deploy-tempest/defaults/main.yml
+++ b/playbooks/roles/airship-deploy-tempest/defaults/main.yml
@@ -22,3 +22,5 @@ cirros_test_image_url: "http://download.cirros-cloud.net/0.4.0/cirros-0.4.0-x86_
 use_blacklist: true
 # Flag to hide tempest pod logs during ansible tempest run. Default is True (show tempest logs)
 print_tempest_log: "{{ lookup('env','AIRSHIP_TEMPEST_LOG_STDOUT') | default('True', true) | bool }}"
+# wheter to output tempest in subunit format
+tempest_subunit: "{{ lookup('env', 'SOCOK8S_TEMPEST_SUBUNIT_OUTPUT') | default(False, True) | bool }}"
--- a/playbooks/roles/airship-deploy-tempest/tasks/main.yml
+++ b/playbooks/roles/airship-deploy-tempest/tasks/main.yml
@@ -137,6 +137,7 @@
       when:
         - _tempest_run_pod_name.rc == 0
         - print_tempest_log
+        - not tempest_subunit
 
     - name: Show command for gathering tempest logs
       debug:
@@ -155,3 +156,4 @@
       when:
         - _tempest_run_pod_name.rc == 0
         - print_tempest_log
+        - not tempest_subunit
--- a/run.sh
+++ b/run.sh
@@ -172,6 +172,9 @@ case "$deployment_action" in
     "delete_ses_rook")
         delete_ses_rook
         ;;
+    "parse_tempest_ci")
+        parse_tempest_output
+        ;;
     *)
         echo "Invalid option, Check --help for valid options"
         ;;
--- a/script_library/deployment-actions-common.sh
+++ b/script_library/deployment-actions-common.sh
@@ -83,3 +83,8 @@ function gather_dstat_output(){
     echo "Gathering dstat output"
     run_ansible ${socok8s_absolute_dir}/playbooks/generic-gather_dstat_output.yml
 }
+
+function parse_tempest_output(){
+    echo "Parsing tempest output for CI"
+    run_ansible ${socok8s_absolute_dir}/playbooks/generic-tempest-output-to-junit.yml
+}
--- a/script_library/deployment-actions-kvm.sh
+++ b/script_library/deployment-actions-kvm.sh
@@ -53,6 +53,10 @@ function clean_kvm(){
     echo "Not implemented"
 }
 function teardown(){
+    if [[ ${SOCOK8S_TEMPEST_PARSE_RESULTS} == "True" ]]
+    then
+      parse_tempest_ci
+    fi
     if [[ ${SOCOK8S_GATHER_LOGS:-"NO"} == "YES" ]]
     then
         gather_logs
--- a/script_library/deployment-actions-openstack.sh
+++ b/script_library/deployment-actions-openstack.sh
@@ -61,6 +61,10 @@ function clean_openstack(){
     run_ansible ${socok8s_absolute_dir}/playbooks/openstack-delete_network.yml
 }
 
+function parse_tempest_ci() {
+    parse_tempest_ci
+}
+
 function teardown(){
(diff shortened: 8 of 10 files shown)


playbooks/roles/airship-deploy-tempest/defaults/main.yml
```diff
@@ -22,3 +22,5 @@ cirros_test_image_url: "http://download.cirros-cloud.net/0.4.0/cirros-0.4.0-x86_
 use_blacklist: true
 # Flag to hide tempest pod logs during ansible tempest run. Default is True (show tempest logs)
 print_tempest_log: "{{ lookup('env','AIRSHIP_TEMPEST_LOG_STDOUT') | default('True', true) | bool }}"
+# wheter to output tempest in subunit format
+tempest_subunit: "{{ lookup('env', 'SOCOK8S_TEMPEST_PARSE_CI') | default(False, True) | bool }}"
```
jsuchome: Why do we want it false by default? That will mean you'll need to explicitly set it before the CI run to get the tempest results...

Itxaka: So we can set it to true on the jenkinsfile but on developers machines is not run by default :D

jsuchome: Ah, true, I'm complaining at the wrong place...

Itxaka:
does not work, pod is completed by the time we try to do this. Would be cool if results were stored in the pvc so we could launch a container and extract those, maybe even do everything (transformation) inside the container

Itxaka:
This is now working, yay!

Itxaka:
OMG IS WORKING: https://ci.suse.de/blue/organizations/jenkins/cloud-socok8s-integration/detail/PR-732/2/tests

And its turning the job into red due to the test output!

Itxaka:
> Looks cool. I will propose a few simplifications with the code later today.

@evrardjp when is later? This should be getting in asap so we can start working on those failing tests asap. Being able to track those is of utmost importance.

Itxaka:
Last change (after the approvals) is to make the step isolated from other so it can fail gracefully and it doesnt intervene with the teardown.

evrardjp:
> > Looks cool. I will propose a few simplifications with the code later today.
> 
> @evrardjp when is later? This should be getting in asap so we can start working on those failing tests asap. Being able to track those is of utmost importance.

Agreed

playbooks/generic-tempest-output-to-junit.yml
```diff
@@ -0,0 +1,83 @@
...
+  tasks:
+    - name: Run container to attach tempest pvc
+      shell: |
+        set pipefail -o
+        cat <<EOF | kubectl -n openstack apply -f -
+        apiVersion: v1
+        kind: Pod
+        metadata:
+          name: tempest-data
+        spec:
```
evrardjp: This could use an init container that mounts the volume and runs cat /data/.stestr/0 | subunit-1to2 | subunit2junitxml > /data/tempest-junit.xml (for example, evrardjp/subunit2junit).

playbooks/generic-tempest-output-to-junit.yml
```diff
@@ -0,0 +1,83 @@
...
+            command: ["/bin/bash", "-ec", "while :; do echo '.'; sleep 5 ; done"]
+            volumeMounts:
+            - name: mypvc
+              mountPath: /data
+          volumes:
+          - name: mypvc
+            persistentVolumeClaim:
+              claimName: pvc-tempest
+        EOF
+      register: tempest_data_pod_create
```
evrardjp: In this case, L26 to L49 can go away.

playbooks/generic-tempest-output-to-junit.yml
```diff
@@ -0,0 +1,83 @@
...
+      when: tempest_data_pod_status.rc == 0
+    - name: Fetch the tempest output
+      fetch:
+        dest: "{{ playbook_dir }}/../tempest.subunit"
+        src: "{{ ansible_env.HOME }}/tempest.subunit"
+        flat: yes
+      when:
+        - _tempest_output.rc == 0
+    - name: Delete the data pod
+      command: kubectl -n openstack delete pod tempest-data --grace-period=5
```
evrardjp: I would make it force and grace-period of 0. Why waiting, we're gonna crash the whole thing anyway?

playbooks/generic-tempest-output-to-junit.yml
```diff
@@ -0,0 +1,83 @@
...
+      shell: |
+        set pipefail -o
+        cat <<EOF | kubectl -n openstack apply -f -
+        apiVersion: v1
+        kind: Pod
+        metadata:
+          name: tempest-data
+        spec:
+          containers:
+          - name: tempest-data
```
evrardjp: If you move to nginx image, you could expose the /data. You can then curl the data, though I wouldn't expect this to be necessary. You could keep the current loop.

playbooks/generic-tempest-output-to-junit.yml
```diff
@@ -0,0 +1,83 @@
...
+    - name: Fetch the tempest output
+      fetch:
+        dest: "{{ playbook_dir }}/../tempest.subunit"
+        src: "{{ ansible_env.HOME }}/tempest.subunit"
+        flat: yes
+      when:
+        - _tempest_output.rc == 0
+    - name: Delete the data pod
+      command: kubectl -n openstack delete pod tempest-data --grace-period=5
+      when: tempest_data_pod_create.rc == 0
```
evrardjp: L51 to end can go away. However you would have to add a task, before the delete, to fetch the files from pod (either by providing a service in the manifest, and curling) or kubectl cp.

playbooks/generic-tempest-output-to-junit.yml
evrardjp: We don't even have to delete, as it will automatically get destroyed when tearing things down.

playbooks/generic-tempest-output-to-junit.yml
```diff
@@ -0,0 +1,83 @@
+---
+- hosts: soc-deployer
+  gather_facts: yes
+  tasks:
+    - name: Run container to attach tempest pvc
+      shell: |
+        set pipefail -o
+        cat <<EOF | kubectl -n openstack apply -f -
```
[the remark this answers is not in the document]
Itxaka: I also thought about this, but its so small and so useless to have that file in there that I dont think it warrants the hassle of having to copy it to the remote node. This way everything is contained in the single playbook, making it easier to manage/fix/track/change

playbooks/generic-tempest-output-to-junit.yml
Itxaka: using suse containers all the way babyyy.

playbooks/generic-tempest-output-to-junit.yml
Itxaka: I dont want to make this more complicated, and that sounds more complicated. Why have init containers to do that when I can do it locally?

Also, this would depend on a extra image, which means more data to download. Separating what the container does to make it minimal means we can move this image to be a really small one (alpine?) so the mount+extraction is really fast. This is already a pain in the ass (mounting a pod to extract data) so I want to minimize the impact of this as much as possible.

Also, I dont want to use containers that much for data processing as they can go away at any moment, this is just a help container to get the data from the pvc and extract it, processing can be done offline.

playbooks/generic-tempest-output-to-junit.yml
Itxaka: I dunno to be honest, Im not sure how killing it could affect the pvc attached, so I didnt want to go all hardcore here. If there is a guarantee of not affecting the pvc then by all means lets kill it fast!

playbooks/generic-tempest-output-to-junit.yml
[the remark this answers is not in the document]
evrardjp: I would rather go with a place where manifests are stored in a main folder, as it might be the first of a series.

playbooks/generic-tempest-output-to-junit.yml
evrardjp: You have to deal with dependency management, and installing stuff in a venv.
venv vs container, it's not really a big difference.

Here the host stays clean, it only needs kubectl.

playbooks/generic-tempest-output-to-junit.yml
evrardjp: Yeah, it was only in the case of the refactor above. In that case, after it's fetched, there is nothing else to do.

playbooks/generic-tempest-output-to-junit.yml
evrardjp: It's fine for me.

playbooks/generic-tempest-output-to-junit.yml
Itxaka: I dont want to use a personal image for this, it should be using a generic image available to all. But I also dont want to manage installing stuff inside a container. So unless there is an image supported by an org (say the official subunit container) that can guarantee that the image is well supported and its not going to break and so on, this is a big no for me.

This also means that we are hiding our processing. Part of it done on ansible, part of it doing on k8s. That just makes it more difficult to maintain and find any issues that arise.

Im totally 100% against this idea, sorry.

>You have to deal with dependency management, and installing stuff in a venv.
venv vs container, it's not really a big difference.

There is actually. Ansible has a module for this that guarantees several things (backwards compatibility, idempotency, etc) meanwhile the container+init does not and we have to manage that ourselves. Notonly that but for a simple task (create venv, install 2 packages, delete venv) we are introducing 3 more layers (k8s, containers, remote)

> Here the host stays clean, it only needs kubectl. In the future, that can even go away, and only have podman and shell.

Host is also clean here, plus it already has python running (otherwise ansible...) so create a venv and deleting it its clean enough. Also this is only for CI.

playbooks/generic-tempest-output-to-junit.yml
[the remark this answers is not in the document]
Itxaka: do not agree here, self contained playbook for a CI-only task sounds much better than a manifest dir which we currently dont need.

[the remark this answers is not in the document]
evrardjp: We'll agree to disagree.

playbooks/generic-tempest-output-to-junit.yml
evrardjp: > I dont want to use a personal image for this, it should be using a generic image available to all. But I also dont want to manage installing stuff inside a container.

This is for CI. We can build our own in IBS/OBS that wouldn't be a big deal.

> This also means that we are hiding our processing. Part of it done on ansible, part of it doing on k8s.

You are currently doing an ansible that runs a shell script that generates the k8s manifest to then apply it using kubectl. I am not sure this is worse, as this effectively teaches us the good practices of k8s, whether they suck or not.

> Im totally 100% against this idea, sorry.
> 
> > You have to deal with dependency management, and installing stuff in a venv.
> > venv vs container, it's not really a big difference.
> 
> There is actually. Ansible has a module for this that guarantees several things (backwards compatibility, idempotency, etc) meanwhile the container+init does not and we have to manage that ourselves. Notonly that but for a simple task (create venv, install 2 packages, delete venv) we are introducing 3 more layers (k8s, containers, remote)

We'll agree to disagree. There is a non zero count of persons that stumbled on the venv thing in this team.

> Host is also clean here, plus it already has python running (otherwise ansible...) so create a venv and deleting it its clean enough. Also this is only for CI.

If this is only for CI, why would it be a problem to make it a container?

evrardjp: We have to remember of the end goal here. I am fine with ansible running for CI. I think at the end of the day SOC10 should not be exposing ansible plays, as the org is terrible with it (too old ansible, nobody would care except us, the state of the architecture of our plays and roles is just awful). If we can replace the whole series of tasks and keep the same feature for CI using native k8s objects, why wouldn't we do that?

We should merge this PR in any case, as it's good enough, and doesn't hurt the project, quite the contrary.

playbooks/generic-tempest-output-to-junit.yml
```diff
@@ -0,0 +1,83 @@
...
+      register: _tempest_file
+    - name: Install required packages in a virtualenv
+      pip:
+        name:
+          - junitxml
+          - python-subunit
+        virtualenv: "{{ playbook_dir }}/../junitvirtualenv"
+        virtualenv_python: python2  # junit stuff does not work under py3 for some reason, needs more investigation
+      when: _tempest_file.stat.exists
+    # This task always outputs rc 1.
```
kantarun: Is above statement applies when there are no failures ?
Following statement provides explanation but above statement was confusing for me with 'always' word

playbooks/generic-tempest-output-to-junit.yml
```diff
@@ -0,0 +1,83 @@
...
+          - name: tempest-data
+            image: opensuse/leap
+            command: ["/bin/bash", "-ec", "while :; do echo '.'; sleep 5 ; done"]
+            volumeMounts:
+            - name: mypvc
+              mountPath: /data
+          volumes:
+          - name: mypvc
+            persistentVolumeClaim:
+              claimName: pvc-tempest
```
kantarun: What's this name refers to. 
Is it pvc name used in tempest run chart ? What if somebody overrides this with different name.

playbooks/generic-tempest-output-to-junit.yml
```diff
@@ -0,0 +1,83 @@
...
+      args:
+        executable: /bin/bash
+      failed_when: False
+      tags:
+        - skip_ansible_lint
+      when: _tempest_file.stat.exists
+    - name: Remove virtualenv
+      file:
+        path: "{{ playbook_dir }}/../junitvirtualenv"
+        state: absent
```
kantarun: Is there need to remove copied tempest file on runner host i.e. {{ playbook_dir }}/../tempest.subunit

Jenkinsfile.integration
```diff
@@ -244,6 +245,14 @@ pipeline {
                         reportName: 'ARA Report'
             ]
             logParser projectRulePath: 'log-parser-plugin-rules.txt', showGraphs: true, useProjectRule: true, parsingRulesPath: null
+            script {
+                try {
+                    sh "./run.sh parse_tempest_ci"
```
kantarun: What's the behavior when tempest pod is not deployed (because of some site update error before this pod is deployed and tempest executed) as its 'always' executed.

playbooks/generic-tempest-output-to-junit.yml
Itxaka: If we do that we need to change it ourselves manually but this is for CI only so it will be pretty clean when we do so :)

playbooks/generic-tempest-output-to-junit.yml
Itxaka: I dont think so, plus it helps if there was any issue when transforming, or if we want to use the subunit file for something

Jenkinsfile.integration
Itxaka: I changed this to only trigger on tempest step success :+1:

playbooks/generic-tempest-output-to-junit.yml
Itxaka: You are rigth, this is confusing, let me reword this

Itxaka:
This is actually passing but it has 4 failing tempest test, so the job goes to fail: https://ci.suse.de/blue/organizations/jenkins/cloud-socok8s-integration/detail/PR-732/25/tests

Lets merge this! I will follow up on this with the tempest fixes/skip/expected failures

PR #734, opened 2019-09-03
Bump the worker flavors

Itxaka:
Bump the workers flavors so they have more ram as we are currently
short on ram when everything is deployed

### DIFF
--- a/playbooks/roles/caasp4/defaults/main.yml
+++ b/playbooks/roles/caasp4/defaults/main.yml
@@ -42,7 +42,7 @@ skuba_ci_terraform_worker_count: 2
 # the flavor name used for master nodes
 skuba_ci_terraform_master_flavor: "m1.large"
 # the flavor name used for worker nodes
-skuba_ci_terraform_worker_flavor: "m1.medlarge"
+skuba_ci_terraform_worker_flavor: "m1.xxlarge"
 
 # A public key file that is used by terraform when setting up nodes
 # Login to the CaaSP 4 nodes is possible with this priv/pub key combination


playbooks/roles/caasp4/defaults/main.yml
```diff
@@ -42,7 +42,7 @@ skuba_ci_terraform_worker_count: 2
 # the flavor name used for master nodes
 skuba_ci_terraform_master_flavor: "m1.large"
 # the flavor name used for worker nodes
-skuba_ci_terraform_worker_flavor: "m1.medlarge"
+skuba_ci_terraform_worker_flavor: "m2.xlarge"
```
jsuchome: m1.xlarge is slightly smaller while still larger than m1.medlarge, what about that one?

Itxaka: 16Gb vs 24Gb, we have the resources :)

dirkmueller: What is the goal? get more ram? how about using m1.xxlarge (same cpu/disk but more ram)?

I'm okay with more ram, but not with more cpu/disk atm.

evrardjp: I am fine with xxlarge, it is by far more ram than I thought. I think it's interesting enough to see how well we'll fare. Then we can upgrade to more vcpus if necessary.  Let's not move too many factors in one go :)

Itxaka: we currently have a load average of 25 in workers, sometimes that even ends in 50 for 1 worker, depending on the pods that fall unto them. So I think a cpu bump would be appreciated

Itxaka: > I'm okay with more ram, but not with more cpu/disk atm.

Why not?

Itxaka:
hey look at that, less than 3 hours of jobs, airship deployed correctly, tempest did not fail with any 503! Seems that increasing resources is helping smooth the deploy!

playbooks/roles/caasp4/defaults/main.yml
dirkmueller: because we're running low on capacity with disk and cpu

Itxaka: but... How does make sense?

We reduced the number of jobs to be able to bump the resources...

So you rather have 7 jobs in parallel consuming X cpus rather than 5 jobs in parallel consuming X cpus? That makes no sense to me.

In that case lets reduce the max cpus reserved in the project to what we can really use instead, otherwise that number is not real as we cannot use it.


But this is a discussion for another time. @dirkmueller can you remove your request-for-change from this PR? Its only bumping the ram so it should be good to go?

Itxaka: oh I see, m2.xlarge is using more disk that we dont need, let me change this to m1.xxlarge

Itxaka:
we should not wait for the CI for this to be merged

evrardjp:
@Itxaka I would like to see the CI for this, to see the timing. I think it's valuable input.

Itxaka:
oh, you will see it @evrardjp its just that we should not wait to merge it (job will keep running) in order to rebase and test the SES rook stuff today :)

PR #733, opened 2019-09-02
Remove the code for standalone SES node

jsuchome:
~WIP: there's still some code using common-ses role where it should not be needed~

PR #736, opened 2019-09-04
run.sh to default to ses_rook and airship

AbelNavarro:


jsuchome:
This is already (partially) addressed as part of https://github.com/SUSE-Cloud/socok8s/pull/733

AbelNavarro:
> This is already (partially) addressed as part of #733

Right, better drop this PR then.

PR #737, opened 2019-09-04
Less retries, longer intervals

jsuchome:
For actions that we know are taking long times, there's no need
to check every 10s for a result.

### DIFF
--- a/playbooks/roles/airship-deploy-osh/tasks/main.yml
+++ b/playbooks/roles/airship-deploy-osh/tasks/main.yml
@@ -190,8 +190,8 @@
         - shipyard_desc_action.stdout.find('Processing') < 0
         - shipyard_desc_action.stdout.find('running') < 0
         - shipyard_desc_action.stdout.find('Error') != 0
-      retries: "{{ airship_deploy_openstack_timeout | int * 2 }}"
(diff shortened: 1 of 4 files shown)


playbooks/roles/airship-deploy-osh/tasks/main.yml
```diff
@@ -185,8 +185,8 @@
       register: shipyard_desc_action
       failed_when: "'failed' in shipyard_desc_action.stdout"
       until: shipyard_desc_action.stdout.find('Processing') < 0 and shipyard_desc_action.stdout.find('running') < 0
-      retries: "{{ airship_deploy_openstack_timeout | int * 2 }}"
-      delay: 30
+      retries: "{{ airship_deploy_openstack_timeout | int }}"
```
AbelNavarro: this effectively sets the same timeout (half retries, double delays). This proves insufficient at least on our internal deployments (times out). I would let the number of retries as they were but with the increased delays.

jsuchome: Yes, my point here was not changing he overall timeouts, only the cleanup so that there's not that many retries for actions that take too much time

PR #738, opened 2019-09-04
Use password for ssh auth in tempest

Itxaka:
Set tempest to use auth authentication on the ssh tests


Itxaka:
not valid. why the hell there is an option and code that deals with this but it doesnt work? wtf openstack??

PR #740, opened 2019-09-05
Delete ara dir on job start

Itxaka:
We want to have a clean slate when we trigger ara. As we are
storing everything in the WORKSPACE, we are safe to delete the dir

This fixes having a really big ara db on long running branches

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -127,6 +127,13 @@ pipeline {
                 sh "rm -Rf .tox"
             }
         }
+        stage('Clean up ara dir') {
+            steps {
+                dir("${env.WORKSPACE}/.ara/") {
+                    deleteDir()
+                }
+            }
+        }
         stage('Create network') {
             when { expression { return  TestFunctional } }
             options {


Itxaka:
> Wait. Workspace is create per-job, right? So how could ARA dir exist when you (re)start a job?

Master and long running branches do not remove the workspace.

Itxaka:
In regards to that, that comes from https://github.com/SUSE-Cloud/automation/blob/master/jenkins/ci.suse.de/templates/cloud-socok8s-integration.yaml#L72

but Im not sure why it was added and Im wary of touching the jobs again as they broke last time...maybe I should just try it.

Itxaka:
I created https://github.com/SUSE-Cloud/automation/pull/3680 to see if it can be fixed like that

Itxaka:
> I thought periodics and all jobs would be in a different workspace, due to build id differing. Is build ID wrong? Maybe I misunderstood this.

It is, seems like build *id* is not used, only build name, so workspace for those are for example:
`WORKSPACE=/home/jenkins/workspace/cloud-socok8s-daily-airship-integration`

evrardjp:
> > I thought periodics and all jobs would be in a different workspace, due to build id differing. Is build ID wrong? Maybe I misunderstood this.
> 
> It is, seems like build _id_ is not used, only build name, so workspace for those are for example:
> `WORKSPACE=/home/jenkins/workspace/cloud-socok8s-daily-airship-integration`

wouldn't that be the problem to solve?

Itxaka:
> wouldn't that be the problem to solve?

This problem should be already solved by https://github.com/SUSE-Cloud/automation/pull/3680 so this should be closed, Im just waiting for confirmation.

But indeed it doesnt seem sane to reuse the same environment. If we go with build ids, then we need to clean the workspace at the end of the job always as to not leave stuff around. And Im very weary of touching anything jenkins related anymore as to not break it.

Itxaka:
not needed anymore as https://github.com/SUSE-Cloud/automation/pull/3680 fixed it

PR #741, opened 2019-09-05
SOC-10462 Workaround bad LB issue

evrardjp:
Without this patch, helm test would fail, because of kubectl
port-forward tunnel getting closed on more than 50 seconds connection
streams.

This is a problem, as helm test is part of our delivery path.

This should fix it by making sure the LB is not used, and we
directly target the master node, until a longer term solution
is found.

### DIFF
--- a/playbooks/roles/caasp4/tasks/setup_k8s.yml
+++ b/playbooks/roles/caasp4/tasks/setup_k8s.yml
@@ -42,7 +42,10 @@
   register: _skuba_cluster_dir
 
 - name: Init skuba cluster
-  command: "skuba cluster init --control-plane {{ _terraform_json_output.ip_load_balancer.value }} {{ skuba_cluster_name }}"
+  #Go back to use LB IP when long term fix is merged.
+  #See also SOC-10463 and https://bugzilla.suse.com/show_bug.cgi?id=1149211
+  #command: "skuba cluster init --control-plane {{ _terraform_json_output.ip_load_balancer.value }} {{ skuba_cluster_name }}"
+  command: "skuba cluster init --control-plane {{ _terraform_json_output.ip_masters.value[0] }} {{ skuba_cluster_name }}"
   args:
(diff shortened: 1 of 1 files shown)


jsuchome:
Please mention a bug number in the commit message (and possibly in the comment)

evrardjp:
> Please mention a bug number in the commit message (and possibly in the comment)

If that's a policy we strictly want to follow, should we add testing for it?

jsuchome:
We do not follow it strictly for _any_ issue, but this is a workaround for a bug and I believe bugs should be reported and tracked. 
So you can't really test it if you mean some automatic check for commit messages.

evrardjp:
> We do not follow it strictly for _any_ issue, but this is a workaround for a bug and I believe bugs should be reported and tracked.
> So you can't really test it if you mean some automatic check for commit messages.

Got it. Fair.

jsuchome:
Is there a bug?

evrardjp:
> We do not follow it strictly for _any_ issue, but this is a workaround for a bug and I believe bugs should be reported and tracked.
> So you can't really test it if you mean some automatic check for commit messages.

Got it. Fair.

> Is there a bug report?

You can find it by looking at the card this code and PR refers to. However, here is the link: https://bugzilla.suse.com/show_bug.cgi?id=1149211

jsuchome:
So, could you add bsc number to the commit message instead of SOC number to the PR title? Commit messages are stored for history, and we want to expose bug numbers instead of numbers of cards in internal task tracking system.

(BTW I may be blind, but I just do not see that bug reference in https://jira.suse.com/browse/SOC-10462)

evrardjp:
> So, could you add bsc number to the commit message instead of SOC number to the PR title? Commit messages are stored for history, and we want to expose bug numbers instead of numbers of cards in internal task tracking system.
> 
> (BTW I may be blind, but I just do not see that bug reference in https://jira.suse.com/browse/SOC-10462)

Will do.

PR #743, opened 2019-09-06
Update the cinder configuration to match changes in upstream charts

jsuchome:
This is because https://review.opendev.org/#/c/666599/. Specific pool
values now have to be provided for each cinder pool we are using.

Specifically, look at the change in job-storage-init for the changed locations of the values:

https://review.opendev.org/#/c/666599/6/cinder/templates/job-storage-init.yaml


jsuchome:
You could just see that `ses_cluster_configuration['cinder']['rbd_store_pool']` is already used in that template, it's not a new addition :-)

PR #744, opened 2019-09-06
Move from ibs in provo to download.suse.de

Itxaka:
There is an ongoing issue with the provo mirror that blocks CI.
Lets move to download.suse.de for now to unblock CI

### DIFF
--- a/vars/common-vars.yml
+++ b/vars/common-vars.yml
@@ -20,7 +20,7 @@ suse_infra_image_version: "latest-opensuse_15"
 suse_osh_openstack_release: "rocky"
 
 obs_mirror: "{{ lookup('env','SOCOK8S_OPENSUSE_MIRROR') | default('https://download.opensuse.org', true) }}"
-ibs_mirror: http://ibs-mirror.prv.suse.net/ibs
+ibs_mirror: "{{ lookup('env','SOCOK8S_IBS_MIRROR') | default('http://ibs-mirror.prv.suse.net/ibs', true) }}"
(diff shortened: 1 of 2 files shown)


vars/common-vars.yml
```diff
@@ -20,7 +20,9 @@ suse_infra_image_version: "latest-opensuse_15"
 suse_osh_openstack_release: "rocky"
 
 obs_mirror: "{{ lookup('env','SOCOK8S_OPENSUSE_MIRROR') | default('https://download.opensuse.org', true) }}"
-ibs_mirror: http://ibs-mirror.prv.suse.net/ibs
+# temporaly disable ibs-mirror from provo as its down
+#ibs_mirror: http://ibs-mirror.prv.suse.net/ibs
+ibs_mirror: http://download.suse.de/ibs
 dist_mirror: http://ibs-mirror.prv.suse.net/dist
```
jsuchome: Isn't this used anywhere?

Itxaka: not anymore, caasp related

evrardjp:
Can we fix what's wrong in provo instead?

Itxaka:
> Could you elaborate on what kind of issues the mirror has? Is it outdated? Not available? Something else?

@nicolasbock not available, the folks at #buildops are already aware of that :)

Itxaka:
provo is back so this is not needed (should have sent it yesterday afternoon when it started to fail).

Will remove the extra for CI but leave the possibility of overriding just in case?

Itxaka:
actually no, lets just close it

PR #745, opened 2019-09-06
Disable helm test containers on CI

Itxaka:
We are interested in the tempest tests to check for issues.
Running the container tests as part of the CI is useless as we
cannot extract any data from those and thye make the deployment
slower.

Not only that, but those tests are run as part of upstream so they
have already been confirmed to work. Running them again makes no
sense from a CI point of view.

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -48,6 +48,7 @@ pipeline {
         ARA_DIR = "$WORKSPACE"
         SOCOK8S_DEPLOY_DSTAT = "YES"
         SOCOK8S_TEMPEST_SUBUNIT_OUTPUT = "True"
+        SOCOK8S_RUN_CONTAINER_TESTS= "False"
     }
 
     stages {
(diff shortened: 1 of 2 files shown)


Itxaka:
> Sadly this would discard ALL tests, including the non-openstack ones. Whether they should be another way to test, that's a different story.

postgres and mariadb are ridiculous tests. Not to mention that the airship tests are not even tests.

Meanwhile this will probably shave up to 30 minutes from deploying airship i.e. minimum of 5 minutes per test container * 6 components (keystone, glance, cinder, neutron, nova, heat).

Im not sure its OUR responsibility to test for charts and containers while that is done upstream. Plus we offer the option to run the full test, which could be moved into the daily job.

Not only that but in cases where those tests fail... we would have catch it anyway as well, because those containers working are prerequisites of the next ones.

Also, I have not seen 1(one) failure of those test containers in all my deployments so Im not sure that is even a real issue not running those tests.

When we have a CI that not only fails all the time, but takes 3+ hours to test a single patch, we need to be pragmatic and try to reduce that asap, in the end that affects our daily work in a significant matter.

Itxaka:
FYI deploy_airship step took 1h 17m 27s which is around 30 minutes less than usual.

Itxaka:
> I will be okay if we make sure we have extra test coverage cards in our backlog for the armada api.

Does that make sense with the upcoming change to airship ?

Or do you mean like having our own testing of the armada api?

PR #746, opened 2019-09-06
Retry teardown on CI

Itxaka:
Currently if one of the steps inside the teardown script fails,
the whole thing will stop. Be it a temporal problem or a real
issue.

This patch wraps the teardown into a retry statement so we try to
cleanup a max of 3 times and then resort to notifying via RocketChat


Itxaka:
I forgot to properly catch errors!

Itxaka:
Hey it actually worked! Tried 3 times and in the third it cleaned it up properly!

evrardjp:
I am not too fond of this approach tbh. I would prefer we deal with the things properly.There is almost no reason to keep the parallelism when things will run on top of caasp.

PR #748, opened 2019-09-06
Correctly add missing labels to first worker node

jsuchome:
Previous check was not able to find a master node, probably because
of the switch to CaaSP4.
This led to the situation where no worker node got
'openstack-helm-node-class=primary' label which is required as a node
selector for (at least) nova-api-metadata pod.

### DIFF
--- a/playbooks/roles/deploy-osh/tasks/enroll-caasp-nodes-for-osh.yml
+++ b/playbooks/roles/deploy-osh/tasks/enroll-caasp-nodes-for-osh.yml
@@ -4,14 +4,14 @@
     set -o pipefail
     kubectl get nodes -o json | jq -M -r '.items[].metadata.name'
   changed_when: false
-  register: _all_casp_nodenames
+  register: _all_caasp_nodenames
 
 - name: Ensure enough nodes are running for HA
   fail:
     msg: "You do not have enough nodes to run OSH in HA"
   when:
     - suse_osh_deploy_require_ha | bool
-    - _all_casp_nodenames.stdout_lines | length < 3
+    - _all_caasp_nodenames.stdout_lines | length < 3
 
 # TODO(evrardjp): Make this conditional if a previous enrollment was already done
 - name: Enroll node for OSH control plane
@@ -23,7 +23,7 @@
     - _enrollforopenstack.rc != 0
     - _enrollforopenstack.stderr.find("already has a value (enabled)") == -1
   with_items:
-    - "{{ _all_casp_nodenames.stdout_lines }}"
+    - "{{ _all_caasp_nodenames.stdout_lines }}"
 
 - name: Enroll node for OSH compute node
   command: "kubectl label nodes {{ item }} openstack-compute-node=enabled"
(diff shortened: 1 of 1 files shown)


jsuchome:
I've not tested it yet (just manual commands, not the way they're wrapped in ansible)

playbooks/roles/deploy-osh/tasks/enroll-caasp-nodes-for-osh.yml
```diff
@@ -45,24 +45,29 @@
...
+  shell: |
+    set -o pipefail
+    kubectl get node --selector='node-role.kubernetes.io/master' -o json | jq '.items[0].metadata.name'
+  changed_when: false
+  register: _master_node
+
+- name: Find first non-master node
+  shell: |
+    set -o pipefail
+    kubectl get node --selector='!node-role.kubernetes.io/master' -o json | jq '.items[0].metadata.name'
```
evrardjp: nit: In those cases you can directly use jsonpath instead of json | jq :)

playbooks/roles/deploy-osh/tasks/enroll-caasp-nodes-for-osh.yml
```diff
@@ -45,24 +45,29 @@
...
   register: _nodes
 
 - name: Find master node
-  set_fact:
-    master_node: "{{ _nodes.stdout_lines | join('') | from_json | json_query(query) }}"
-  vars:
-    query: 'items[].spec.{node: externalID, taints: taints} | [?taints].node'
+  shell: |
+    set -o pipefail
+    kubectl get node --selector='node-role.kubernetes.io/master' -o json | jq '.items[0].metadata.name'
```
evrardjp: nit: you could probably -o name

playbooks/roles/deploy-osh/tasks/enroll-caasp-nodes-for-osh.yml
```diff
@@ -71,14 +76,12 @@
...
+  when: _primary_node is defined
 
 # TODO(evrardjp): Move this to check the return code and task content
 - name: Remove taints when only 3 nodes and ha is required
-  command: "kubectl taint nodes {{ item }} node-role.kubernetes.io/master:NoSchedule-"
-  with_items: "{{ master_node }}"
+  command: "kubectl taint nodes {{ _master_node.stdout }} node-role.kubernetes.io/master:NoSchedule-"
   when:
-    - "(master_node | default('')) | length > 0"
+    - _master_node is defined
```
evrardjp: nit: It will always be defined, else it won't run this task because of a failure :) 
You can remove this.

playbooks/roles/deploy-osh/tasks/enroll-caasp-nodes-for-osh.yml
jsuchome: Oh, that's a good point!

playbooks/roles/deploy-osh/tasks/enroll-caasp-nodes-for-osh.yml
```diff
@@ -45,24 +45,29 @@
...
   changed_when: false
   register: _nodes
 
 - name: Find master node
-  set_fact:
-    master_node: "{{ _nodes.stdout_lines | join('') | from_json | json_query(query) }}"
-  vars:
-    query: 'items[].spec.{node: externalID, taints: taints} | [?taints].node'
+  shell: |
+    set -o pipefail
```
Itxaka: is pipefail needed here? There is no pipes in the command

jsuchome: You are right, but I'll just go ahead and merge it anyway...

PR #753, opened 2019-09-10
Enabled not Eanbled

evrardjp:
ERMAHGERD [1] spelling of enabled is hard.

[1]: https://user.fm/files/v2-d82d6601bd01bdafea94f9bd7cd55e99/ermahgerd.jpg


jsuchome:
Force merging despite the tempest failure that is unrelated ... what could possibly go wrong?

PR #754, opened 2019-09-10
Drop ingress override for k8s

Itxaka:
We are no longer using the dev-patcher VIPs patch so we should use
the ingress used by default upstream

This may or may not help with the 503?


Itxaka:
ci is passing, failure reported due to tempest tests: https://ci.suse.de/blue/organizations/jenkins/cloud-socok8s-integration/detail/PR-754/3/pipeline

PR #757, opened 2019-09-11
Enable apparmor profile for dnsmasq

Itxaka:
Looks like apparmor is now enabled by default again so we need
to run the profile task to make sure neutron dhcp containers
have access to the host routes to read and write dhcp releases

This enables back the dnsmasq profile which was temporarily disabled
due to apparmor being off is previous caasp4 versions


evrardjp:
I have not tested this, but I see it as a positive iteration.

PR #760, opened 2019-09-11
Ignore failures detected by junit

jesusaurus:
The junit step is detecting known failures and marking the jobs as
unstable, which github interprets as a failure. By setting the
health-scale factor to zero, we disable the test result contribution
to the build score, and the jobs should no longer be marked as unstable.

This should be reverted once the tempest tests are passing reliably.

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -233,7 +233,7 @@ pipeline {
                     script {
                         try {
                             sh "./run.sh parse_tempest_ci"
-                            junit testResults: 'tempest.xml', allowEmptyResults: true, keepLongStdio: true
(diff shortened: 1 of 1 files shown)


Itxaka:
Is this needed know that we have the blacklist? The point is to make the tests always mandatory so we dont break stuff with our changes, so tests should contribute to the job status, otherwise we have no assurance that we are not breaking stuff :)

BTW, those blacklisted tests should be removed pretty soon as we are working on fixing those

jesusaurus:
No, this was meant as an alternative approach, it's no longer useful now that the blacklist change has merged.

Itxaka:
> No, this was meant as an alternative approach, it's no longer useful now that the blacklist change has merged.

Sorry I did miss your comment in 756 before merging!

PR #761, opened 2019-09-12
[DNM]TEST: change docker0 iface for eth0

Itxaka:
We are using crio, not docker

### DIFF
--- a/site/soc/software/charts/osh/openstack-compute-kit/neutron.yaml
+++ b/site/soc/software/charts/osh/openstack-compute-kit/neutron.yaml
@@ -42,6 +42,9 @@ data:
       interface:
         tunnel: {{ neutron_tunnel_device }}
     conf:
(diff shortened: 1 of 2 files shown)


site/soc/software/charts/osh/openstack-compute-kit/neutron.yaml
```diff
@@ -42,6 +42,10 @@ data:
       interface:
         tunnel: {{ neutron_tunnel_device }}
```
aojea: this defaults to `docker0`, right?

Itxaka: nope, it defaults to null which means that osh will automagically get the proper interface:

```
 # If tunnel is null there is a fallback mechanism to search
 # for interface with routing using tunnel network cidr.
```

PR #764, opened 2019-09-12
Run tempest test in serial mode

AbelNavarro:


jsuchome:
Did you try more than one run to ensure this is not just a chance?

AbelNavarro:
> Did you try more than one run to ensure this is not just a chance?

nope, will run it locally to test that.

Itxaka:
lets retrigger the CI job as well

AbelNavarro:
There have been 3 failed rebuilds on this PR. The problems seem to be unrelated to the change:
- Invalid repository
- Failure on terraform apply
- time out while deploying kube-system-ingress

I tried locally as well. Several tests were failing, only one with 503 error, but also several failing due to timeouts trying to reach `compute.openstack.svc.cluster.local` which I think it's related to those 503s.

So, this is an improvement over the previous situation but it doesn't solve all problems.

PR #765, opened 2019-09-13
Intelligently restart either docker or crio

jesusaurus:
In caasp4 we no longer have docker running on the caasp workers,
remove the docker restart that will now fail.

### DIFF
--- a/playbooks/roles/setup-caasp-workers/tasks/copy-certificates.yml
+++ b/playbooks/roles/setup-caasp-workers/tasks/copy-certificates.yml
@@ -9,14 +9,25 @@
     dest: "/etc/pki/trust/anchors"
   register: _copiedcert
 
-- name: Refresh ca certs
-  become: True
-  command: update-ca-certificates
-  when: _copiedcert is changed
+- block:
+    - name: Refresh ca certs
+      become: True
+      command: update-ca-certificates
 
-- name: Restart docker
-  become: True
+    - name: Populate service info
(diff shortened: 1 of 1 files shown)


jsuchome:
Do we need that copy certificate part? Do we need whole 'build image' path now that the images should be built in OBS? 
If yes, and this code is still needed, maybe we need to restart crio service instead of docker

jesusaurus:
There are two code paths that run these tasks. The `build_images` path may not be needed anymore (if it's not we should remove that playbook), but this also runs during `setup_caasp_workers_for_openstack` which looks like it may run on either the caasp workers or the deployer/registry.

I've expanded the change to check for either docker or cri-o and restart as needed.

jhesketh:
The build images will hopefully go away once we start consuming images from I/OBS. However lets leave it until we have that workflow more well implemented.

playbooks/roles/setup-caasp-workers/tasks/copy-certificates.yml
```diff
@@ -9,14 +9,25 @@
     dest: "/etc/pki/trust/anchors"
   register: _copiedcert
 
-- name: Refresh ca certs
-  become: True
-  command: update-ca-certificates
-  when: _copiedcert is changed
+- block:
```
Itxaka: pretty cool, I didnt knew you could use a block to trigger using a when instead of having each task with the when!

PR #766, opened 2019-09-17
Manually mount BPFFS on caasp workers

Itxaka:
When cilium pods are restarted, there is network disruption.
Enabling the BPF fs in the workers should avoid any network
disruption when pods are restarted.

There is a PR upstream to fix this, but for the moment lets apply it
ourselves


Itxaka:
rebased

PR #767, opened 2019-09-17
Skip the tempest step for OSH (SOC-10225)

jsuchome:
That might be useful for scenarios that include their own set of tests
(like deploy_osh deployment path).

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -219,7 +219,12 @@ pipeline {
             options {
                 timeout(time: 240, unit: 'MINUTES', activity: true)
             }
-            when { expression { return  TestFunctional } }
+            when {
+                allOf {
+                    expression { params.deployment == "airship" }
+                    expression { return  TestFunctional }
+                }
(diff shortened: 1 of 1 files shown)


jsuchome:
I'm not sure what you mean - automatically skip tempest run based on deployment value?

With this I make it possible to still run "airship" tempest, it just is not a default

Itxaka:
> I'm not sure what you mean - automatically skip tempest run based on deployment value?

when deployment is `airship` you run this step, when its `osh` we can run a different step that only deploys tempest and gathers the output.

We have the `deployment` value int here already, its what makes the osh or airhsip path so we should be able to use it for the tempest deploy which is different in both parts.

Plus we will still want to run tempest on osh, just as a different step

evrardjp:
I think @Itxaka said it better than I did, but it's the same idea.

jsuchome:
Well, the tempest run that is executed by deploy_osh is (currently) not an extra step, but whatever.

PR #768, opened 2019-09-17
Run tempest at the end of OSH deployment by default

jsuchome:


playbooks/roles/deploy-osh/defaults/main.yml
```diff
@@ -99,4 +99,4 @@ suse_osh_deploy_neutron_yaml_overrides: {}
 suse_osh_deploy_tempest_yaml_overrides: {}
 
 # override in your extravars to deploy and run tempest
-deploy_tempest: False
```
Itxaka: would prefer if this was set to false+ ENV var so we can override it from the CI instead

jsuchome: We have it true by default for airship so why not for osh?

Itxaka: you are right here!

PR #769, opened 2019-09-17
[DNM][WIP] Add CI scale profile

Itxaka:
This bumps ingress and keystone api to 2 pods so they can deal with
all the traffic properly


Itxaka:
doesnt really bump anything. I wonder if half of the site/soc stuff is ignored...

PR #770, opened 2019-09-18
Bump ingress to 0.23.0

Itxaka:
According to versions we should have a newer ingress version as
we are using commit 5e1ecd9840397bf9e8829ce0d98fcb721db1b74e for
ingress-htk which includes this bump but somehow airship is ignoring
this and using an older version (0.20.0)

It just happens that this version has a know problem of returning
a 503 on backend reload.

In other words, this may fix the dreaded 503 once and for all.

This approach forces the image to be the 0.23.0 as a test, not final


Itxaka:
This is wrong but it served it purpose of finding the actual issue (helm chart missing for ingress)

Continued in #771

PR #771, opened 2019-09-18
Add missing ingress chart

Itxaka:
We were missing the ingress chart version. This meant that we were
picking the ingress images from ... somewhere which gave us really
really old images (0.20.0)

Bumping this to a commmit from 5 months ago like the companion
ingress-htk has, which should give us a 0.23.0 image

### DIFF
--- a/site/soc/software/config/versions.yaml
+++ b/site/soc/software/config/versions.yaml
@@ -47,6 +47,11 @@ data:
         reference: 5e1ecd9840397bf9e8829ce0d98fcb721db1b74e
         subpath: helm-toolkit
         type: git
+      ingress:
+        location: https://opendev.org/openstack/openstack-helm-infra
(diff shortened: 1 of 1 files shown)


Itxaka:
> This seems correct, however why wasn't there a problem before?

Who says it wasnt?

Before we tracked the tempests test we were basically ignoring the output of the deploy_airship task. No matter if it worked or failed, we discarded and turned the job green (still happening now, see SOC-10375) so no wonder we didnt noticed this.

Before airship got into our peaceful lives, we were using osh and that deploys the latest version of ingress thus skipping this issue, if it existed.

Itxaka:
This doesnt solve the 503 but at least we have a newer version which fixes some stuff and we control the sha of the ingress chart

Itxaka:

PR #772, opened 2019-09-18
Wait up to 120s for connections to finish before reloading

Itxaka:
Looks like there is a limit of 10 seconds for a connection to
finish when a reload is done on ingress before it gets killed.

This means that long requests like those of openstack can get killed
if ther is a backend reload in the start of middle of them.

This bumps it to 120s so we can let those requests finish. This should
not affect anything else as this is only applied to ongoing requests,
so any new requests will get a reloaded backend as this will only
affect the worker dealing with that request

### DIFF
--- a/site/soc/software/charts/kubernetes/ingress/ingress.yaml
+++ b/site/soc/software/charts/kubernetes/ingress/ingress.yaml
@@ -16,5 +16,9 @@ metadata:
       - method: merge
         path: .
   storagePolicy: cleartext
-data: {}
+data:
+  values:
+    conf:
+      ingress:
+        worker-shutdown-timeout: "120s"
 ...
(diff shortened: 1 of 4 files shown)


Itxaka:
@jgu17 I hope the subtitutions and merge paths are ok? Could not find much documentation regarding that upstream so I had to wing it by checking the other yamls

jsuchome:
I do not see this option in ingress helm chart, what am I missing?

Itxaka:
> I do not see this option in ingress helm chart, what am I missing?

This is interesting, so nginx ingress has an option to load config FROM a configmap :o

https://github.com/openstack/openstack-helm-infra/blob/master/ingress/templates/bin/_ingress-controller.sh.tpl#L37

So this is very cool because you can add options to the confimap live and the backend will reload and use them!

And the ingress-conf configmap options are here: https://github.com/kubernetes/ingress-nginx/blob/master/docs/user-guide/nginx-configuration/configmap.md with its counterparts in plan nginx.

jsuchome:
Oh. thanks for explanation.

evrardjp:
Imagine if that was a CRD instead of a configmap, to have an application level understanding of what's going on? :shocked: !

Itxaka:

PR #773, opened 2019-09-18
Bump kesytone chart

Itxaka:
Includes changes to probes to avoid probewarnings


jsuchome:
Could you add a link the changes you mention  in commit message?

Itxaka:
> Could you add a link the changes you mention in commit message?

Sorry I sent this yesterday nigth to check whether it would be working on the latest version or not, its just a test for the moment until I find out why its not working, as it may require more changes.

jsuchome:
OK, thanks for explanation (but if you indeed find this is needed, include the link in the commit message :-) )

Itxaka:

PR #774, opened 2019-09-19
Change MariaDB's backend to local path

jsuchome:


Itxaka:
no airship version? :p

jsuchome:
I'm just exploring for now ...

Itxaka:
Im gonnna guess that due to this, we __need__ to have a HA deployment for both components? Otherwise if we have a single pod and it breaks there is the possibility of being relaunched in a different node, and localpath is only available on the worker where the node was spawned?

Also HA queues I guess so messages dont get lost?

Also whats the difference between this and directly using Local Persistent Volumes? Is just an extra layer on top? Any benefits?

jsuchome:
> Im gonnna guess that due to this, we need to have a HA deployment for both components? 

AFAIU, if mariadb has 3 pods it will deploy as galera cluster. And you've already noted that this currently modifies only `osh` path which is using the default (3 nodes) setup.

For HA rabbitmq, I honestly have no idea if there's anything (besides pod number) to configure

evrardjp:
I think we shouldn't have to worry, in this PR, how rabbitmq is deployed. I will trust @Gsantomaggio to come up with a generic approach for a Rabbitmq cluster on top of k8s that we could leverage in OpenStack. He is also core on oslo.messaging, so I trust him a lot for rabbitmq and its openstack integration (as much as we can trust openstack integration of rabbitmq) :)

jsuchome:
I'm not sure how to understand your rabbitmq comment: we should not worry and keep it using ceph, or we should not worry and switch to local?

evrardjp:
> I'm not sure how to understand your rabbitmq comment: we should not worry and keep it using ceph, or we should not worry and switch to local?

Please continue to a PoC with local PVs :)
If any change is required on the helm chart (which I doubt), just don't focus on them, and skip rabbitmq for now.
I would love to see us using the upstream helm charts from rabbitmq instead of the OSH forked ones.

Itxaka:
I would worry because if we are using localpath with a single pod, any restart of those pods will cause data to be lost if they get rescheduled in a different worker. This is not about changes to helm charts but more of introducing more instability in our CI runs, where one restart can trash the full DB :D

jsuchome:
I can remove the rabbitmq part for now and leave it for some followup. Anyway it seems the biggest issue is database

jsuchome:
@Itxaka adapted for airship (though not yet tested with it)

AbelNavarro:
I'm leaning towards @Itxaka reasoning that Local Persistent Volumes may be the way to go. For several reasons:

- Local Path Provisioner can be dinamically created, but if I was a CaaSP admin I won't like to see users eating up local storage without my knowledge. I can see the case were CaaSP and SOC admins are on different teams with different priorities.
- Local Path Provisioner is under heavy development while Local Persistent Volumes are already GA
- With Local Persistent Volumes you ensure the pod using it is scheduled to be in the same node (https://kubernetes.io/blog/2019/04/04/kubernetes-1.14-local-persistent-volumes-ga/)

jsuchome:
But:
- this is mostly for development and CI. For production usage, there needs to be done more of course. For example when deploying you may need to specify directory for where the storage is created. And this also is part of the policy file. So this all can be configured in real world.

- it is much easier to set up Local Path Provisioner, it automatically gives you storage class and the only thing to change in helm charts is to point to the new SC

jsuchome:
> With Local Persistent Volumes you ensure the pod using it is scheduled to be in the same node
> (https://kubernetes.io/blog/2019/04/04/kubernetes-1.14-local-persistent-volumes-ga/)

I'm actually wondering what is better for Galera. If new mariadb pod is created that should be part of existing galera cluster, does it matter that it misses some data? It should be able to sync them from the rest of cluster. It might actually be an advantage for the case the node went completely down and we do _not_ want to schedule anything there...

Am I wrong?

Itxaka:
for galera it should be ok I guess as it will catch up with the other nodes, wont it?

For standalone nodes it would be terrible as it basically means losing everything for a restart. And losing everything -> broken deployment, yay! As like we have not enough of those already :P

This seems good enough for me for our CI. More robust and fast is always good!

playbooks/roles/common-deployer/tasks/local-provisioner.yml
```diff
@@ -0,0 +1,22 @@
+---
+- name: Clone local-path provisioner
+  git:
+    repo: "https://github.com/rancher/local-path-provisioner"
+    dest: "{{ upstream_repos_clone_folder }}/local-path-provisioner"
+    update: yes
+    force: yes
+    version: "v0.0.10"
```
Itxaka: can this be sent in a var so we can test new versions by overriding it?

jsuchome: I guess. Actually I only added the version because lint told me to

AbelNavarro:
> But:
> 
> * this is mostly for development and CI. For production usage, there needs to be done more of course. For example when deploying you may need to specify directory for where the storage is created. And this also is part of the policy file. So this all can be configured in real world.
> * it is much easier to set up Local Path Provisioner, it automatically gives you storage class and the only thing to change in helm charts is to point to the new SC

Let me take back the argument that CaaSP and SOC may have different admins. At this moment, at networking level, the implementations are quite crossed. This means that to debug SOC probably you'll need to SSH into the CaaSP worker and fiddle in there. The logical network interfaces are almost at the same level, so it is going to be difficult to separate CaaSP and SOC administration/operation.

evrardjp:
> I'm leaning towards @Itxaka reasoning that Local Persistent Volumes may be the way to go. For several reasons:
> 
>     * Local Path Provisioner can be dinamically created, but if I was a CaaSP admin I won't like to see users eating up local storage without my knowledge. I can see the case were CaaSP and SOC admins are on different teams with different priorities.

Rancher dynamic provisioner is based on host path and local PVs, afaik. It just leverages hostPath to then provision the PVs you can then use with a PVC.
It's using both techs. It's not one vs the other.

If there are CaaSP and SOC admins, they could get different (Cluster)Role bindings, if really necessary. But I seriously doubt that should happen. If they are so different, they can simply get two k8s clusters.

> 
>     * Local Path Provisioner is under heavy development while Local Persistent Volumes are already GA

It's a simple shim, and in k8s, everything is under dev. :p 

> 
>     * With Local Persistent Volumes you ensure the pod using it is scheduled to be in the same node (https://kubernetes.io/blog/2019/04/04/kubernetes-1.14-local-persistent-volumes-ga/)

See comment above. Is the dynamic provisioner not doing that?

evrardjp:
> I'm actually wondering what is better for Galera. 

We should indeed ensure it's always scheduled to the same node, and reschedule on the same node. It's simpler for operations.

evrardjp:
> for galera it should be ok I guess as it will catch up with the other nodes, wont it?

That's exactly what we want to avoid as much as we can, as catching up might take time. Assuming a reboot of the node, if not using a local PV, you would reschedule on a new node, and get to a full sync, while in the case of a local PV you would only do a partial sync.

> For standalone nodes it would be terrible as it basically means losing everything for a restart. And losing everything -> broken deployment, yay! As like we have not enough of those already :P

Exactly.

evrardjp:
> Let me take back the argument that CaaSP and SOC may have different admins. At this moment, at networking level, the implementations are quite crossed. This means that to debug SOC probably you'll need to SSH into the CaaSP worker and fiddle in there. The logical network interfaces are almost at the same level, so it is going to be difficult to separate CaaSP and SOC administration/operation.

We probably don't want to have this SSH fiddling. I think we should document when we did SSH fiddling, and propose solutions.

jsuchome:
So, what do you think, can we proceed with this one or do we want some alternative solution?

aojea:
+1 for merging, the immediate benefit seems bigger than the trade offs ... and we can always iterate. Also, you can still have multiple storage backends and pick the one that you want, this is just an alternative

PR #775, opened 2019-09-19
Create CI scale profile

Itxaka:
Add a scale pod profile for ci which has 2 pods for keystone api
to be more resilient to pod failure or timeout. Also adds new
options to the ingress controller for openstack, namely:
 - proxy_next_upstream: Now also catch 503 and pass to next upstream
 - proxy_next_upstream_timeout: timeout after 10s
 - proxy_next_upstream_tries: number of tries to pass to next upstream

All this options are explained in the nginx docs[0]

Also changes the scale profile env var to something more exclusive
of our socok8s repo so its clear where its coming from, changed
from scale_profile to SOCOK8S_SCALE_PROFILE

[0] http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_next_upstream

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -49,6 +49,7 @@ pipeline {
         SOCOK8S_DEPLOY_DSTAT = "YES"
         SOCOK8S_TEMPEST_SUBUNIT_OUTPUT = "True"
         SOCOK8S_RUN_CONTAINER_TESTS= "False"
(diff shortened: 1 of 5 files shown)


Itxaka:
Looks like this worked...good to know we have this + proxy_next_upstream in case that the other patches dont work.

Itxaka:
one of the test nodes had issues and several pods were in error state..

Itxaka:
Not needed but can be used as example I guess

PR #777, opened 2019-09-19
Try to fix 503

Itxaka:
Several commits with efforts to improve our ingress backend reloading so much that causes connections to drop.

I still have an extra commit to add if this doesnt work.


jsuchome:
glance-registry is also mentioned in `playbooks/roles/deploy-osh/files/common-images-overrides.yml`, could it be dropped as well? (it does not have any effect, but just for the cleanup)

Itxaka:
This works, splitting everything as 2 of this patches are not needed anymore and the bump keystone to latest will change when the needed patches are merged upstream

PR #778, opened 2019-09-20
Use default number of nodes (3) to deploy MariaDB galera cluster

jsuchome:
See SOC-10344

This is mostly a test PR to see how the CI will react. It's possible more is needed than just this simple change...

### DIFF
--- a/site/soc/software/charts/osh/openstack-mariadb/mariadb.yaml
+++ b/site/soc/software/charts/osh/openstack-mariadb/mariadb.yaml
@@ -15,13 +15,6 @@ metadata:
       - method: delete
         path: .values.labels.prometheus_mysql_exporter
   storagePolicy: cleartext
-  substitutions:
-    - src:
-        schema: pegleg/PodScaleProfile/v1
-        name: pod-scale-profile
-        path: .pods.osh.mariadb.server.min
-      dest:
-        path: .values.pod.replicas.server
 data:
   wait:
     timeout: {{ openstack_helm_deploy_timeout }}
@@ -32,8 +25,6 @@ data:
     pod:
       resources:
         enabled: {{ openstack_helm_pod_resources_enabled['mariadb'] }}
-      replicas:
-        server: 1
     volume:
       size: {{ db_volume_size }}
       backup:


site/soc/software/charts/osh/openstack-mariadb/mariadb.yaml
```diff
@@ -32,8 +32,6 @@ data:
     pod:
       resources:
         enabled: {{ openstack_helm_pod_resources_enabled['mariadb'] }}
-      replicas:
```
Itxaka: only removing this is not enough because on line 22 we are replacing the `.values.pod.replicas.server` from the chart with our values from `.pods.osh.mariadb.server.min` which comes from the pod scale profile (in our case, the default pod-scale-mininal.yaml)

In order to increase this to 3, we need to change that value in the pod scale profile

jsuchome: Thanks, good point!

However I think we do not want to change the definition of `pod-scale-minimal.yaml` profile, to keep it minimal, right? So here what should be enough is to remove this value replace ... I hope

Itxaka: yep, that should work but....then you are removing the user being able to pass a custom profile for the number of pods, so they cant configure if they want just 1 pod or 30, which is bad IMO, especially becuase we still have that `.pods.osh.mariadb.server` option in both scale profiles (minimal and ha)

jsuchome: Ok, then I should point to different profile. That has to be the right solution.

jsuchome: Update: it is not. I cannot point to another profile while another one is being used as a default

jsuchome:
Surprisingly it works like a charm. And no problems when it is still using external SES...

jsuchome:
@Itxaka So what do you suggest? I'd think creating the new profile, right?

(But speaking about end-user: won't this user setup look differently at the time of Airship 2.0 anyway?)

PR #781, opened 2019-09-21
Use dev patcher to patch keystone chart, fixes 503  (SOC-10514)

Itxaka:
In order to use a non-merged patch[0] to keystone to fix the probe
timeouts we need to move to local patching and passing the local
patched repo to ariship

This also bumps the keystone-api pod timeout for the probes to 30s
as by default this is 5s and it causes the ingress pod to remove
keystone-api from the pool as soon as it receives a timeout, which
happens when there is load in the servers (i.e. tempest tests)

As a side effect this also includes a minor change to our site/soc
config to keep using the same name as we used before for keystone
rabbit as the way of generating the name was changed here[1]

[0] https://review.opendev.org/#/c/683220/
[1] https://github.com/openstack/openstack-helm/commit/9bcf0df94c1a8455e17aa79f20efaf2736eed0f4

### DIFF
--- a/site/soc/software/charts/osh/openstack-keystone/keystone.yaml
+++ b/site/soc/software/charts/osh/openstack-keystone/keystone.yaml
@@ -27,6 +27,14 @@ data:
     enabled: {{ run_tests }}
     timeout: {{ test_timeout }}
   values:
+    probes:
+      keystone_api:
+        # bump probe timeout to 30s to avoid ingress removing the pods from the pool when they take long to respond
+        # default upstream is 5s
+        readinessProbe:
+          timeoutSeconds: 30
+        livenessProbe:
+          timeoutSeconds: 30
     pod:
       resources:
         enabled: {{ openstack_helm_pod_resources_enabled['keystone'] }}
--- a/playbooks/roles/airship-deploy-ucp/tasks/main.yml
+++ b/playbooks/roles/airship-deploy-ucp/tasks/main.yml
@@ -330,6 +330,18 @@
     - name: Wait until all pods to be ready
       command: "{{ upstream_repos_clone_folder }}/openstack/openstack-helm/tools/deployment/common/wait-for-pods.sh {{ ucp_namespace_name }} 3600"
 
+    # TODO(aagate): Add a changed_when: to help idempotency
+    # local changes need to go into the armada-api pod as thats the one executing the code
+    - name: Copy patched repos into Armada-api pod
+      command: "{{ item }}"
+      loop:
+        - kubectl exec {{ armada_pod_name }} -n ucp -- mkdir -p /armada/airship-components
+        - kubectl cp {{ upstream_repos_clone_folder }}/openstack/openstack-helm-infra {{ ucp_namespace_name }}/{{ armada_api_pod_name }}:/armada/airship-components/
+        - kubectl cp {{ upstream_repos_clone_folder }}/openstack/openstack-helm {{ ucp_namespace_name }}/{{ armada_api_pod_name }}:/armada/airship-components/
+      tags:
+        - install
+        - skip_ansible_lint
+
   rescue:
     - name: List all ucp pods
       command: "kubectl get pods -n {{ ucp_namespace_name }}"
--- a/playbooks/roles/dev-patcher/defaults/main.yml
+++ b/playbooks/roles/dev-patcher/defaults/main.yml
@@ -7,4 +7,7 @@ dev_patcher_packages:
 #merged patches will be skipped
 #unrelated patches will fail (because not cloned)
 #conflicting patches will fail (because need resolving)
-dev_patcher_patches: []
+dev_patcher_patches:
+  # https://review.opendev.org/#/c/683220/ -> Keystone: allow probes configuration
+  - - 683220
+    - current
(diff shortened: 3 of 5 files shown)


Itxaka:
Failure was due to deckhand

```
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/falcon/api.py", line 244, in __call__
    responder(req, resp, **params)
  File "/usr/lib/python3.6/site-packages/shipyard_airflow/policy.py", line 339, in secure_handler
    return f(slf, req, resp, *args, **kwargs)
  File "/usr/lib/python3.6/site-packages/shipyard_airflow/control/api_lock.py", line 49, in func_wrapper
    return func(self, req, resp, *args, **kwargs)
  File "/usr/lib/python3.6/site-packages/shipyard_airflow/control/configdocs/configdocs_api.py", line 87, in on_post
    empty_collection=empty_coll)
  File "/usr/lib/python3.6/site-packages/shipyard_airflow/control/configdocs/configdocs_api.py", line 213, in post_collection
    document_data)
  File "/usr/lib/python3.6/site-packages/shipyard_airflow/control/helpers/configdocs_helper.py", line 624, in add_collection
    self.deckhand.put_bucket(collection_id, document_string)
  File "/usr/lib/python3.6/site-packages/shipyard_airflow/control/helpers/deckhand_client.py", line 124, in put_bucket
    response = self._put_request(url, document_data=documents)
  File "/usr/lib/python3.6/site-packages/shipyard_airflow/control/helpers/deckhand_client.py", line 345, in _put_request
    'Unable to Invoke deckhand: {}'.format(str(rex)),
shipyard_airflow.control.helpers.deckhand_client.DeckhandAccessError

2019-09-21 13:53:21,125 INFO     bb10fe0d-caa1-4da7-9d2c-d50db23430b8 bb10fe0d-caa1-4da7-9d2c-d50db23430b8 shipyard 2cfe082d56eb452eaee505d535e934f8 logging_mw(57) process_response - POST http://shipyard.ucp.svc.cluster.local/api/v1.0/configdocs/soc-design?buffermode=replace&verbosity=1 - 500 Internal Server Error
2019-09-21 13:53:21,126 DEBUG    bb10fe0d-caa1-4da7-9d2c-d50db23430b8 bb10fe0d-caa1-4da7-9d2c-d50db23430b8 shipyard 2cfe082d56eb452eaee505d535e934f8 logging_mw(63) process_response - Errored Response body: {"kind": "status", "apiVersion": "v1.0", "metadata": {}, "status": "Failure", "message": "Unhandled Exception raised: ", "reason": "", "details": {"errorType": "DeckhandAccessError", "errorCount": 1, "messageList": [{"message": "An error occurred, but was not specified", "error": true}]}, "code": "500 Internal Server Error", "retry": true}```

**Comment by jsuchome** [comment]
_This also bumps the keystone-api pod timeout for the probes to 30s
as by default this is 5s and it causes the ingress pod to remove
keystone-api from the pool as soon as it receives a timeout, which
happens when there is load in the servers_ 

So... isn't this an upstream issue?

**Comment by Itxaka** [comment]
> So... isn't this an upstream issue?

Kind of and kind not. In the case of having more than one keystone-api pod (which is the default upstream, other than gating) and/or running this on baremetal instances where the performance is increased VS our engcloud nodes, this timeout is not reached or only one of the 2 pods is not responding momentarily, so the requests are still reaching keystone. In our case, because we default to just one pod, and our performance is not that good, and we also run the airship pods along, which are taxing the nodes, its much much easier for our keystone to take more than 5 seconds to respond, hence the ingress pod takes it out of the pool as soon as those 5 seconds have been reached.

So in that instance the issue is downstream, be it our lack of HA or our lacking performance.

In the case of the probe configuration, that IS an upstream issue, as one-size-fits-all does not work here, we may want to configure that to either be much more letinent or be more agressive with the timeouts. Hence that part was modified upstream to allow finer configuration for the probe options.


Lets discuss on the daily of this is not clear :)

**Comment by jsuchome** [comment]
So increasing our default number of pods for keystone would be alternative solution

**Comment by Itxaka** [comment]
> So increasing our default number of pods for keystone would be alternative solution

Indeed! and that was done to test it under https://github.com/SUSE-Cloud/socok8s/pull/775 and it workerd. It still gave some 503 from time to time, probably due to resource starvation, but it was more stable.

I think that in the short future we may need that (so HA ci deployment to be closer to what we are gonna give customers) + still some timeout configuration to not be that agressive (maybe 15 seconds instead of 30?)

**Comment by Itxaka** [comment]
yay!

File: site/soc/software/charts/osh/openstack-keystone/keystone.yaml [hunk 1]
```diff
@@ -27,6 +27,14 @@ data:
     enabled: {{ run_tests }}
     timeout: {{ test_timeout }}
   values:
+    probes:
+      keystone_api:
+        # bump probe timeout to 30s to avoid ingress removing the pods from the pool when they take long to respond
+        # default upstream is 5s
+        readinessProbe:
+          timeoutSeconds: 30
```
AbelNavarro: I think these 30s values are ok for our CI environment due to the performance issues we found and the low amount of replicas we use. For production I don't think these values are going to be acceptable. Waiting 30 seconds to receive an error sounds like too much time.

Since we want to improve our current situation this is a +1 for me, but we do have to remember to tune down these values in the future or make them configurable.

Itxaka: totally agree, see incoming PR to make this configurable from the scale profile

PR #782, opened 2019-09-21
Add missing ingress chart

Itxaka:
We were missing the ingress chart version. This meant that we were
picking the ingress images from ... somewhere which gave us really
really old images (0.20.0)

Bumping this to a commmit from 5 months ago like the companion
ingress-htk has, which should give us a 0.23.0 image


Itxaka:
rebased!

PR #784, opened 2019-09-23
Move probe config to scale profile

Itxaka:
Move the config for the probes into the scale profile so its
configurable instead of hardcoded and its present in a single
place

### DIFF
--- a/playbooks/roles/airship-deploy-ucp/files/profiles/pod-scale-ha.yaml
+++ b/playbooks/roles/airship-deploy-ucp/files/profiles/pod-scale-ha.yaml
@@ -10,6 +10,21 @@ metadata:
     abstract: false
     layer: site
 data:
+  probes:
+    osh:
+      keystone_api:
+        readinessProbe:
+          initialDelaySeconds: 15
+          periodSeconds: 10
+          timeoutSeconds: 1
+          successThreshold: 1
+          failureThreshold: 3
+        livenessProbe:
+          initialDelaySeconds: 50
+          periodSeconds: 20
+          timeoutSeconds: 5
+          successThreshold: 1
+          failureThreshold: 3
   pods:
     ucp:
       armada:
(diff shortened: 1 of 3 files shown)


Itxaka:
dont merge until upstream its merged as the format migth change

jsuchome:
I like this ... although we might want different profile as well

Itxaka:
would be really cool if this provided a way of setting a default value so it would be possible to have a missing value in the scale profile. It kind of does with the gobal stuff, but thats not good enough. Having something in the site/soc that says "pick this value from the scale profile and if doesnt exists just use this default" would be really nice to avoid filling the profiles with so much needed stuff....

jgu17:
> 
> 
> would be really cool if this provided a way of setting a default value so it would be possible to have a missing value in the scale profile. It kind of does with the gobal stuff, but thats not good enough. Having something in the site/soc that says "pick this value from the scale profile and if doesnt exists just use this default" would be really nice to avoid filling the profiles with so much needed stuff....

should the default value in the chart yamls already? Profile is just another way of overrides.

Itxaka:
> should the default value in the chart yamls already? Profile is just another way of overrides.

mmmh, but using substitution means that it will substitute the whole thing from the profile into the chart rigth? So unless the profile contains everything, it will be missing things. 

so profile:

```
data:
  probes:
    osh:
      keystone_api:
        readinessProbe:
          initialDelaySeconds: 15
```

Chart:
```
probes:
  osh:
    keystone_api:
      readinessProbe:
        initialDelaySeconds: 1
		anotherValue: 1
        anotherValue2: 1
```

So if we use substitution like this:
```

    - src:
        schema: pegleg/PodScaleProfile/v1
        name: pod-scale-profile
        path: .probes.osh.keystone_api
      dest:
        path: .values.probes.keystone_api
```

We end up with the following in the final treasuremap:
```
probes:
  osh:
    keystone_api:
      readinessProbe:
        initialDelaySeconds: 15
```

So we will be missing the other default values.

We could just use the substitution for each value, but that is just too verbose and difficult to track.

Is it possible to do a `merge` stanza and pick the values *from* the scale profile?

PR #785, opened 2019-09-23
Remove bridge-utils package installation from OSH

ilausuch:
The repository that contains bridge-utils was removed in #750
so the installation of OSH cannot continue.

Because this package is not going to be used any more, is removed
from the installation

### DIFF
--- a/playbooks/roles/deploy-osh/defaults/main.yml
+++ b/playbooks/roles/deploy-osh/defaults/main.yml
@@ -18,7 +18,6 @@ suse_osh_deploy_packages:
   - python-selinux
   - ceph-common
   - nfs-utils
-  - bridge-utils
 
 suse_osh_openstackclient_packages:
   - python3-openstackclient


jsuchome:
Let's restart the CI with `deploy_osh` once it fails

evrardjp:
I thought this was still required for neutron on the nodes. I assume that with ovs  it might not be required.

evrardjp:
Why did we need bridge-utils? AFAIK it was for OSH it might be worth checking if this breaks osh.

Itxaka:
> Why did we need bridge-utils? AFAIK it was for OSH it might be worth checking if this breaks osh.

This is being installed in the deployer, which makes no sense. It was introduced on commit number 0, so it may be a leftover from testing or something.

jsuchome:
Can we remove more from `suse_osh_deploy_packages`?

PR #786, opened 2019-09-23
Increase the timeout for neutron test which we are hitting lately

jsuchome:
In my setups 'helm test neutron'  often takes about 20 minutes.


AbelNavarro:
If your tests take 20 minutes and you set the timeout to exactly 20 minutes, won't that be a bit tight?

jsuchome:
Yeah, it might be tight, but as I suspect my 20 minutes is caused by slow HW or whatever resource I'm lacking so I'm hoping for better times for other people.
However I can increase it if you think it would be better

AbelNavarro:
The timeout increase looks good to me if it fits you. If 20 minutes is what you need it's ok for me. But if you need 22 minutes don't be shy and add it here. I'm approving it in case 20 fits you.

PR #791, opened 2019-09-24
Remove SUSE-cloud/automation module

rajdharwadkar:
Removing SUSE-cloud/automation module because it is not used.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
-[submodule "submodules/SUSE-Cloud/automation"]
-	path = submodules/SUSE-Cloud/automation
-	url = https://github.com/SUSE-Cloud/automation.git
 [submodule "submodules/openstack-helm-infra"]
 	path = submodules/openstack-helm-infra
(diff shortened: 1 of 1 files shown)


jsuchome:
So, why we needed it in the first place? Because of CaaSP3? (It would be great to mention the reason for removal in the commit message)

jhesketh:
It's no longer needed/used afaik, but yes, mentioning that in the commit message would be helpful.

rajdharwadkar:
> So, why we needed it in the first place? Because of CaaSP3? (It would be great to mention the reason for removal in the commit message)

I am not sure why and when this was added.

rajdharwadkar:
> You need to remove the automation from the .gitmodules, otherwise it would fail: https://github.com/SUSE-Cloud/socok8s/blob/master/.gitmodules

yes thanks . Fixed.

rajdharwadkar:
We need these changes to go in first : 
https://github.com/SUSE-Cloud/socok8s/pull/731/files
https://github.com/SUSE-Cloud/socok8s/pull/733/files

PR #792, opened 2019-09-25
Fix keystone patch on dev-patcher

Itxaka:
There was some changes required on the keystone patch upstream.
Now that an agreement has been reached, the patch has been changed
so we need to adapt our downstream values for keystone to make it
compatible and fix the 503s again.


Itxaka:
patch was merged upstream we we can just bump the keystone chart to latest and use this patch, yay! no more copying around stuff!

PR #793, opened 2019-09-25
Fix deleting rook task

jsuchome:
Do not reference files that were not used during the deployment

### DIFF
--- a/playbooks/generic-delete_ses_rook.yml
+++ b/playbooks/generic-delete_ses_rook.yml
@@ -15,6 +15,8 @@
 - hosts: soc-deployer
   gather_facts: no
   any_errors_fatal: true
+  vars_files:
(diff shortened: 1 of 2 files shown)


playbooks/generic-delete_ses_rook.yml
```diff
@@ -15,6 +15,8 @@
 - hosts: soc-deployer
   gather_facts: no
   any_errors_fatal: true
+  vars_files:
+    - "{{ playbook_dir }}/../vars/common-vars.yml"
```
Itxaka: is this needed? commit doesnt say why and Im not seeing any vars that are not self contained?

jsuchome: It's for the repo path, I will adapt commit message

Itxaka: ouch, so this is not standalone anymore then :(

PR #794, opened 2019-09-25
Some fixes for the libvirt deployment

rhafer:
With these changes I am able to deploy CaaSPv4 and Rook inside that CaaSP as well as completing the "setup_caasp_workers_for_openstack" step. (I am still waiting for the reset of the steps to complete)

Note: There is some overlap to https://github.com/SUSE-Cloud/socok8s/pull/780, but different to that I've tried to avoid spawning yet another VM for the deployer node by just (mis-)using the lb-node as the deployer.


script_library/deployment-actions-common.sh
```diff
@@ -12,6 +12,11 @@ function build_images(){
     run_ansible ${socok8s_absolute_dir}/playbooks/generic-build_images.yml -e "build_osh_images=yes"
 }
 
+function configure_ccp_deployer() {
+    echo "Configure CCP deployer node"
+    run_ansible ${socok8s_absolute_dir}/playbooks/openstack-configure_ccp_deployer.yml
```
Itxaka: can you rename this file to "generic" instead?

playbooks/openstack-osh_hostconfig.yml
```diff
@@ -1,5 +1,6 @@
 ---
 - hosts: soc-deployer
+  become: yes
```
Itxaka: ?

playbooks/roles/caasp4/tasks/setup_nodes.yml
```diff
@@ -56,6 +56,19 @@
     src: inventory-caasp4.yml.j2
     dest: "{{ socok8s_workspace }}/inventory/caasp4.yml"
 
+- name: Extend inventory to co-use the LB node as OSH deployer node
```
Itxaka: how would this work if the ip is not really a node? Any further tasks that want to use the deployer node for a play will fail.

playbooks/roles/caasp4/tasks/setup_nodes.yml
```diff
@@ -56,6 +56,19 @@
...
+  copy:
```
Itxaka: also, this works in parallel in the CI with the real deployer, so there may be file conflicts writing to the soc-deployer.yml file

playbooks/roles/caasp4/tasks/setup_nodes.yml
```diff
@@ -56,6 +56,19 @@
...
+    content: "{{ soc_aio_overrides }}"
```
Itxaka: Im also not sure what this really does?? Its doing something completely out of scope for the caasp role plus its only for a specific deployment but it will affect both deployments...why not have this somewhere else if needed?

Itxaka:
if you dont want to use an extra node for this, just point your soc-deployer inventory to the caasp master instead? Seems like a more proper approach rather than faking a node and modifying common parts for something kvm specific

playbooks/roles/caasp4/tasks/setup_nodes.yml
rhafer: Hm, good point.  I think I need to skip this when `deployment_mechanism==openstack`

playbooks/roles/caasp4/tasks/setup_nodes.yml
Itxaka: for this, kvm-deploy_caasp.yml seems the perfect play to move this to. Its out of the role, its specific to kvm, and its clear to what is doing.

playbooks/openstack-osh_hostconfig.yml
rhafer: Those tasks need to run as root. But the terraform deployed nodes don't allow ssh logins for root. That why we set  `ansible_user' to `sles` in the inventory `become: yes` here to run the task as root.
I was hoping the commit message would explain it good enough. Seems it didn't :/

playbooks/roles/caasp4/tasks/setup_nodes.yml
rhafer: In case of a kvm/libvirt deployment it is always real node. I just forgot to skip for openstack deployments.

playbooks/openstack-osh_hostconfig.yml
Itxaka: ah, I missed the commit message completely so thats my fault!

playbooks/roles/caasp4/tasks/setup_nodes.yml
rhafer: Yeah, I first tried to implement it inside a new kvm-deploy_ccp_deployer.yml task. But that added so much boilerplate that I didn't really like it. I guess I need to revisit that decision :)

rhafer:
> if you dont want to use an extra node for this, just point your soc-deployer inventory to the caasp master instead? Seems like a more proper approach rather than faking a node and modifying common parts for something kvm specific

It's not a fake node. That LB node is something real in the kvm deployment. I think using that node as the deployer (instead of master) creates a setup that better matches the openstack deployment than if we'd use the master as the deployer. (Unless there is the plan to do that for the openstack deployments as well)

I agree that putting the code into the caasp role was probably the wrong decision.

script_library/deployment-actions-common.sh
rhafer: done

playbooks/roles/caasp4/tasks/setup_nodes.yml
rhafer: > for this, kvm-deploy_caasp.yml seems the perfect play to move this to. Its out of the role, its specific to kvm, and its clear to what is doing.

I've now move it to it's now playbook (kvm-deploy_ccp_deployer.yml). Mainly to closely match openstack-* step. But I can also add it to the kvm-deploy_caasp if you prefer it that way.

Itxaka:
> It's not a fake node. That LB node is something real in the kvm deployment.

Interesting. I expected kvm to be able to provide a lb IP like openstack :o

rhafer:
> 12288 for every 3 workers? 36 GB? If that's really needed I'm no longer interested in having a local deployment (my workstation has only 32 GB RAM)...

Of course it depends a lot on what exactly you deploy. But IIRC in the ECP deployments we ran out of memory with 8GB nodes. And when looking at crowbar/mkcloud (https://github.com/SUSE-Cloud/automation/blob/master/scripts/mkcloud#L98) we're also at 12G for the controllers.

I must admin that I haven't really tried to find a minimum for the worker_memory yet. I just know that the default of 2048 is much too low. (And of course it's always possible to override the values)

Itxaka:
> Of course it depends a lot on what exactly you deploy. But IIRC in the ECP deployments we ran out of memory with 8GB nodes. And when looking at crowbar/mkcloud (SUSE-Cloud/automation:scripts/mkcloud@master#L98 ) we're also at 12G for the controllers.

We are actually using 20Gb nodes, but that could be lowered to 16Gb (Usually free memory is around 2-4gb). 8Gb was too low and resulted in OOMKills all around by the end of the deployment. *if* you only deploy osh, you could get away with much less, but with airship in the mix the resources get really high (More components running, each osh component gets its own rabbitmq, etc..)

playbooks/roles/caasp4/templates/skuba-terraform-libvirt.tfvars.j2
```diff
@@ -13,31 +13,34 @@ masters = {{ master_count }}
...
+  serverapps_pool = "{{ ibs_mirror }}/SUSE/Products/SLE-Module-Server-Applications/15-SP1/x86_64/product/"
+  caasp_updates = "{{ ibs_mirror }}/SUSE/Updates/SUSE-CAASP/4.0/x86_64/update/"
+  sle_server_updates = "{{ ibs_mirror }}/SUSE/Updates/SLE-Product-SLES/15-SP1/x86_64/update/"
+  basesystem_updates = "{{ ibs_mirror }}/SUSE/Updates/SLE-Module-Basesystem/15-SP1/x86_64/update/"
+  containers_updates = "{{ ibs_mirror }}/SUSE/Updates/SLE-Module-Containers/15-SP1/x86_64/update/"
+  serverapps_updates = "{{ ibs_mirror }}/SUSE/Updates/SLE-Module-Server-Applications/15-SP1/x86_64/update/"
 }
 
+master_memory = 4096
+worker_memory = 12288
```
Itxaka: should move this into a var so it can be overriding from the extravars?

playbooks/roles/caasp4/templates/skuba-terraform-libvirt.tfvars.j2
```diff
@@ -13,31 +13,34 @@ masters = {{ master_count }}
...
+  caasp_pool = "{{ ibs_mirror }}/SUSE/Products/SUSE-CAASP/4.0/x86_64/product/"
+  sle_server_pool = "{{ ibs_mirror }}/SUSE/Products/SLE-Product-SLES/15-SP1/x86_64/product/"
+  basesystem_pool = "{{ ibs_mirror }}/SUSE/Products/SLE-Module-Basesystem/15-SP1/x86_64/product/"
+  containers_pool = "{{ ibs_mirror }}/SUSE/Products/SLE-Module-Containers/15-SP1/x86_64/product/"
+  serverapps_pool = "{{ ibs_mirror }}/SUSE/Products/SLE-Module-Server-Applications/15-SP1/x86_64/product/"
+  caasp_updates = "{{ ibs_mirror }}/SUSE/Updates/SUSE-CAASP/4.0/x86_64/update/"
+  sle_server_updates = "{{ ibs_mirror }}/SUSE/Updates/SLE-Product-SLES/15-SP1/x86_64/update/"
+  basesystem_updates = "{{ ibs_mirror }}/SUSE/Updates/SLE-Module-Basesystem/15-SP1/x86_64/update/"
+  containers_updates = "{{ ibs_mirror }}/SUSE/Updates/SLE-Module-Containers/15-SP1/x86_64/update/"
+  serverapps_updates = "{{ ibs_mirror }}/SUSE/Updates/SLE-Module-Server-Applications/15-SP1/x86_64/update/"
```
kantarun: In general, I like this change as it makes sense. 
In my local setup, I encountered frequent timeout during cloud-init of nodes and primarily issue was related to kernal-default package download and install which is around 250 MB.
So to get around this issue, I did end up setting up RMT server locally and then set up basesystem_pool and baseystem_updates referencing from that local RMT repos. Having this name as RMT server uri primarily gave the idea otherwise I even may not have looked into this approach.
Point is that in situations like this, having RMT server uri different from ibs mirror can be useful.

playbooks/roles/caasp4/templates/skuba-terraform-libvirt.tfvars.j2
kantarun: +1

vars/deploy-on-kvm.yml
```diff
@@ -1,2 +1,2 @@
 # number of caasp worker nodes
-deploy_on_kvm_caasp_workers: 1
+deploy_on_kvm_caasp_workers: 3
```
kantarun: Apparently we need 3 nodes for rook monitor quorums. Earlier I have 2 worker node and ceph monitor and manager pods never came up with that number

playbooks/roles/caasp4/templates/skuba-terraform-libvirt.tfvars.j2
rhafer: > Point is that in situations like this, having RMT server uri different from ibs mirror can be useful.

I am non entirely sure what you mean here. As far as I can see all repos that use `ibs_mirror` in socok8s currently can be mirrored using an RMT. That's why I remove the the `rmt_mirror` thing.

rhafer:
> Most of changes look fine to me and it will need some more changes in setup_airship flow as deployer is ssh with 'sles' user and not root user as done in openstack mechanism.

Yes. Same is true for the deploy_osh step. I'll work through those in a followup PR

> We may want to have similar node setup for openstack and KVM mechanism. So it makes sense to reduce the footprint if deployer is not needed. And we may want to make related changes for both mechanism in that case.

Yes. But this PR is not really getting rid of the deployer. It's just about co-using the Loadbalancer node (which does not exist in the openstack deployment) as a deployer node. Getting rid of the deployer completely should be done elsewhere IMO.

> In my workstation, I have bridge network to LAN. So I was not able to make it work with default libvirt templates found in skuba repo

Is there a reason why you can't use the default NAT network that the skuba templates create. It works just fine. Why do you need a bridge network? I'd rather not fork the skube templates too much. It will just mean another thing we need to maintain.

playbooks/roles/caasp4/templates/skuba-terraform-libvirt.tfvars.j2
rhafer: Done

playbooks/roles/caasp4/templates/skuba-terraform-libvirt.tfvars.j2
```diff
@@ -13,31 +13,30 @@ masters = {{ master_count }}
...
-  sle_server_pool = "{{ rmt_mirror_uri }}/SUSE/Products/SLE-Product-SLES/15-SP1/x86_64/product/"
-  basesystem_pool = "{{ rmt_mirror_uri }}/SUSE/Products/SLE-Module-Basesystem/15-SP1/x86_64/product/"
-  containers_pool = "{{ rmt_mirror_uri }}/SUSE/Products/SLE-Module-Containers/15-SP1/x86_64/product/"
-  serverapps_pool = "{{ rmt_mirror_uri }}/SUSE/Products/SLE-Module-Server-Applications/15-SP1/x86_64/product/"
-  sle_server_updates = "{{ rmt_mirror_uri }}/SUSE/Updates/SLE-Product-SLES/15-SP1/x86_64/update/"
-  basesystem_updates = "{{ rmt_mirror_uri }}/SUSE/Updates/SLE-Module-Basesystem/15-SP1/x86_64/update/"
-  containers_updates = "{{ rmt_mirror_uri }}/SUSE/Updates/SLE-Module-Containers/15-SP1/x86_64/update/"
-  serverapps_updates = "{{ rmt_mirror_uri }}/SUSE/Updates/SLE-Module-Server-Applications/15-SP1/x86_64/update/"
+  caasp_pool = "{{ ibs_mirror }}/SUSE/Products/SUSE-CAASP/4.0/x86_64/product/"
+  sle_server_pool = "{{ ibs_mirror }}/SUSE/Products/SLE-Product-SLES/15-SP1/x86_64/product/"
```
jsuchome: Could you update the locations in `skuba-terraform-openstack.tfvars.j2` as well? (could be in separate PR if you want).

playbooks/roles/caasp4/templates/skuba-terraform-libvirt.tfvars.j2
kantarun: ibs_mirror is used throughout in different plays and can have many more repos which are not needed in cloud-init steps. So if rmt and ibs mirror URI are same that means RMT server needs to have all those additional repos mirrored as well. Issue may not appear important unless the setup is done on slow VPN.

Anyway, that's more of workaround and specific to env, so we may not have to handle it here. So fine with ibs_mirror if that's the preference.

kantarun:
> > Most of changes look fine to me and it will need some more changes in setup_airship flow as deployer is ssh with 'sles' user and not root user as done in openstack mechanism.
> 
> Yes. Same is true for the deploy_osh step. I'll work through those in a followup PR
> 
> > We may want to have similar node setup for openstack and KVM mechanism. So it makes sense to reduce the footprint if deployer is not needed. And we may want to make related changes for both mechanism in that case.
> 
> Yes. But this PR is not really getting rid of the deployer. It's just about co-using the Loadbalancer node (which does not exist in the openstack deployment) as a deployer node. Getting rid of the deployer completely should be done elsewhere IMO.

I meant we have deployer node in 'openstack' mechanism deployment but not in KVM node. Its fine if that can be done in separate PR for openstack mechanism


> > In my workstation, I have bridge network to LAN. So I was not able to make it work with default libvirt templates found in skuba repo
> 
> Is there a reason why you can't use the default NAT network that the skuba templates create. It works just fine. Why do you need a bridge network? I'd rather not fork the skube templates too much. It will just mean another thing we need to maintain.

I have bridge network as VMs can be hosted on different machines in LAN (to meet memory requirement) and LAN manages DHCP for those VMs. This way, those VMs can communicate to each other without any additional routing magic.
Also plan is to have some of those changes added in skuba side once changes are tested and finalized. 
I know there are other folks who have similar bridge network setup to get around hardware/memory requirement.

rhafer:
> I meant we have deployer node in 'openstack' mechanism deployment but not in KVM node.

There is a deployer node in the KVM mode. It just happens to be co-located with the Loadbalancer node. The bigger difference is that there is no real Loadbalancer node in the OpenStack mode.

> I have bridge network as VMs can be hosted on different machines in LAN (to meet memory requirement) and LAN manages DHCP for those VMs. This way, those VMs can communicate to each other without any additional routing magic.

Ah, ok.

> Also plan is to have some of those changes added in skuba side once changes are tested and finalized.

Wouldn't it be better to get those changes in to skuba directly, without first forking half of the skuba terraform files here? But I guess this is a discussion that we should have on the other PR.

> Can you clarify, where is your socok8s repo hosted (ansible runner node). Is it directly on libvirt (workstation) host or in a VM ? If in a VM (which will have it own NAT virtual network), will it work with NAT network created by terraform for cluster nodes?

Currently I am running it directly on the host. But for it to work from a VM I think you just need to put that VM on the bridged network not the entire cluster.

PR #795, opened 2019-09-25
Merge suse_airship_deploy_storage and suse_osh_deploy_storage into one

jsuchome:
Use suse_deploy_storage in common-deployer part


jsuchome:
Well, let's see if I broke it (there's actually one change in namespace which I do not understand ... we do not have `ceph` namespace anywhere, right?)

AbelNavarro:
does the `suse_` prefix add any value here? Feels like it could actually be removed.

jsuchome:
@AbelNavarro I think there were the times that people could expect socok8s does not need to be SUSE only tool

PR #796, opened 2019-09-26
Remove kubectl setup parts that were relevant to CaaSP3 only

jsuchome:


jsuchome:
> bye velum!

You will be missed ... by NO ONE!

jsuchome:
rebased

playbooks/roles/common-deployer/tasks/kubectl-setup.yml
```diff
@@ -15,65 +15,25 @@
     path: "{{ ansible_user_dir }}/.kube"
     mode: "0700"
 
-- name: Check if automation tooling files are present
+- name: Find if runner has a kubeconfig file
   stat:
-    path: /opt/ccp/environment.json
-  register: automation_environment_json
+    path: "{{ kubeconfig_file_path }}"
```
Itxaka: I would watch out because this var is wrongly used and it may need to change. Its pointing to socok8s_deploy_config_location but socok8s_deploy_config_location should be pointed to a different place, not to the workspace as the var its used in a remote machine to store files and has nothing to do with the local socok8s workspace!

TL;DR: I would change `kubeconfig_file_path` to be `"{{ socok8s_workspace }}/kubeconfig` and *then* use it here.

jsuchome: Fair enough ... but I have no idea why `socok8s_deploy_config_location` points to `socok8s_workspace ` ...

PR #798, opened 2019-09-26
[WIP] Rework tempest deployment

Itxaka:
This reworks tempest deployment to be more close to what we do
upstream. Also renames some of the vars to make them more clear.
Tempest network/subnet/pools are now deployed using heat so
its more consistent

Together with this, we change the default values for some of the
vars to be able to have a sane environment with proper values
for vars (i.e. no more docker0)

Also adds a section to add masquerading in our caasp workers so
floating ips work


heat-templates/tempest-network.yml
```diff
@@ -0,0 +1,56 @@
...
+        get_param: subnet_name
+      network:
+        get_resource: public_net
+      cidr:
+        get_param: subnet_cidr
+      gateway_ip:
+        get_param: subnet_gateway
+      enable_dhcp: false
+      dns_nameservers:
+        - 10.96.0.10
```
AbelNavarro: where this IP came from?

Itxaka: upstream osh templates, will probably be changed to something

PR #799, opened 2019-09-26
Remove keystone local patch

Itxaka:
Patch was merged upstream so we need to remove and adapt our charts
to the latest keystone chart revision


Itxaka:
deckhand keeps crashing because why not?

```2019-09-28 10:55:35.276 7 ERROR armada.api keystoneauth1.exceptions.connection.ConnectFailure: Unable to establish connection to http://deckhand-api.ucp.svc.cluster.local:9000/api/v1.0/revisions/1/rendered-documents: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response',))
```

PR #800, opened 2019-09-30
Check if OSD's were deployed with the rook deployment

jsuchome:
It could happen that no OSD exist and deploy_ses_rook playbook
still finishes without any error. Let's make an explicit check for OSD's.

### DIFF
--- a/playbooks/roles/ses-rook/tasks/main.yml
+++ b/playbooks/roles/ses-rook/tasks/main.yml
@@ -103,6 +103,13 @@
   retries: 10
   delay: 45
 
+- name: Wait for OSD pods to appear
+  command: "kubectl get pods -n rook-ceph -l app=rook-ceph-osd"
+  register: _rook_ceph_osd_pods
+  until: "'Running' in _rook_ceph_osd_pods.stdout"
+  retries: 4
+  delay: 30
+
 - name: Set up SES for OpenStack
   import_tasks: setup_for_openstack.yml
   vars:


playbooks/roles/ses-rook/tasks/main.yml
```diff
@@ -103,6 +103,17 @@
   retries: 10
   delay: 45
 
+- name: Check for presense of OSD pods
```
AbelNavarro: "presence"

jsuchome: thanks

playbooks/roles/ses-rook/tasks/main.yml
```diff
@@ -103,6 +103,17 @@
...
   delay: 45
 
+- name: Check for presence of OSD pods
+  shell: |
+    set -o pipefail
+    kubectl get pods -n rook-ceph -l app=rook-ceph-osd --no-headers=true --field-selector=status.phase=Running | wc -l
+  register: _rook_ceph_osd_pods
+
+- name: Make sure there is some OSD pod running
+  fail:
```
Itxaka: do we want to directly fail or could it be that they take a bit to appear?

Also, cna we sync between the checks for existence with the other ones above? I dont mind either way but I would really like to see us using the same approach on all of them :)

jsuchome: In all the tests I've done at this point osd pod either already exists (esp. after all those tests passing) or it will not appear.
Anyway: you're suggesting to use similar wait loop like the other tests, right? OK, sure, I can do it.

Itxaka: >  Anyway: you're suggesting to use similar wait loop like the other tests, right? OK, sure, I can do it.

Mainly yeah, so its just the same thing all over

jsuchome: Updated

PR #801, opened 2019-10-01
[DNM][WIP] Bump airship

Itxaka:
Collection of commits due to bumping airship versions


site/soc/secrets/passphrases/osh_rabbitmq_erlang_cookie.yaml
```diff
@@ -0,0 +1,11 @@
...
+schema: deckhand/Passphrase/v1
+metadata:
+  schema: metadata/Document/v1
+  name: osh_rabbitmq_erlang_cookie
+  layeringDefinition:
+    abstract: false
+    layer: site
+  storagePolicy: cleartext
+data: {{ lookup('password', secrets_location + '/osh_rabbitmq_erlang_cookie ' + password_opts) }}
+...
```
jsuchome: What's this for? I do not see it used in the code (not mentioning the commit message)

site/soc/secrets/passphrases/ucp_postgres_exporter_password.yaml
```diff
@@ -0,0 +1,11 @@
...
+schema: deckhand/Passphrase/v1
+metadata:
+  schema: metadata/Document/v1
+  name: ucp_postgres_exporter_password
+  layeringDefinition:
+    abstract: false
+    layer: site
+  storagePolicy: cleartext
+data: {{ lookup('password', secrets_location + '/ucp_postgres_exporter_password ' + password_opts) }}
+...
```
jsuchome: What's this for? I do not see it used in the code (not mentioning the commit message)

Itxaka: This is due to changes upstream in the global chart for postgres

All this documents gets stored in the database and then can be called to substitute those values in other files. 

In this case you can see that this is used on the global chart for postgres here: https://github.com/airshipit/treasuremap/blob/16cdf842d2f0d2410b5d156bfbc9f123f233070a/global/software/charts/ucp/core/postgresql.yaml#L72

This will look for a `deckhand/Passphrase/v1` object which has the name `ucp_postgres_exporter_password` and use its value to substitute the value `.values.endpoints.postgresql.auth.exporter.password` in the postgress chart

jsuchome: Oh. Thanks. 
So the value is stored in treasuremap and then used as a replacement value in postgres chart, right? 

But who saves the original value into secrets_location?

Itxaka: its actually deckhand the one that parses and saves all of this documents I think.

The original value comes from ansible. This took me a while to understand as it looks like a normal lookup but its not. Lookup + password makes it so ansible tries to read the file provided and if it doesnt exists it will create that file + write a random password into it. So on the first pass this lookup tries to read the file at `secrets_location + '/ucp_postgres_exporter_password'` and then becuase it doesnt find it, it will create a password and store it in that file and substitute it in thjis template.

Quite weird if you dont know how password lookup works!

jsuchome: Ouch. I did not expect to learn something like this today. Thanks again!

PR #802, opened 2019-10-03
Set local-path as a default storage class for rabbitmq

jsuchome:
Adapt all current profiles so that the option is configurable by user.

To be discussed with @Gsantomaggio


Gsantomaggio:
The code looks good to me, but I'd like to test it.
Is there an easy way?

jsuchome:
Well ... it depends if you've tried socok8s yourself already.
(You can take a look e.g. at https://confluence.suse.com/display/SOC/Deploying+socok8s+in+ECP+Cheat+Sheet)

Or I can try to deploy it and let you in so you can look around ...

evrardjp:
This is a stepping stone towards using upstream rabbitmq helm charts from the helm hub like @Gsantomaggio is using and/or a rabbitmq operator that we could build on top of it :)

PR #804, opened 2019-10-04
Another cleanup of variables

jsuchome:


playbooks/roles/deploy-osh/files/common-images-overrides.yml
```diff
@@ -1,79 +1,20 @@
...
-    barbican_api: "{{ suse_osh_registry_location }}/openstackhelm/barbican:{{ suse_osh_image_version }}"
-    barbican_db_sync: "{{ suse_osh_registry_location }}/openstackhelm/barbican:{{ suse_osh_image_version }}"
-    bootstrap: "{{ suse_osh_registry_location }}/openstackhelm/heat:{{ suse_osh_image_version }}"
-    cinder_api: "{{ suse_osh_registry_location }}/openstackhelm/cinder:{{ suse_osh_image_version }}"
-    cinder_backup: "{{ suse_osh_registry_location }}/openstackhelm/cinder:{{ suse_osh_image_version }}"
-    cinder_db_sync: "{{ suse_osh_registry_location }}/openstackhelm/cinder:{{ suse_osh_image_version }}"
-    cinder_scheduler: "{{ suse_osh_registry_location }}/openstackhelm/cinder:{{ suse_osh_image_version }}"
-    cinder_volume: "{{ suse_osh_registry_location }}/openstackhelm/cinder:{{ suse_osh_image_version }}"
-    cinder_volume_usage_audit: "{{ suse_osh_registry_location }}/openstackhelm/cinder:{{ suse_osh_image_version }}"
     congress_api: "{{ suse_osh_registry_location }}/openstackhelm/congress:{{ suse_osh_image_version }}"
```
Itxaka: This image pointing is also used in airship `site/soc/software/config/versions.yaml` can we unify those in common-vars maybe?

jsuchome: I'm not sure. The images I left in common-images-overrides.yml are only those we are not explicitly changing (from upstream helm chart overrides or directly from our socok8s values overrides that did not make it upstream). So I think currently there's no overlap of `common-images-overrides.yml` and ` site/soc/software/config/versions.yaml`

But it's possible I did not get your point.

vars/common-vars.yml
```diff
@@ -16,7 +16,11 @@ socok8s_libvirtuuid: "{{ socok8s_workspace }}/libvirt.uuid"
 suse_osh_registry_location: "docker.io"
 suse_openstack_image_version: "rocky-opensuse_15"
 suse_infra_image_version: "latest-opensuse_15"
-suse_osh_openstack_release: "rocky"
+suse_openstack_release: "rocky"
+
+# ClusterRole to allow privileged containers
+suse_privileged_cluster_role: suse:caasp:psp:privileged
```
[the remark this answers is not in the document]
jsuchome: Doesn't seem necessary. It worked before, check the diff.

